PR #17, opened 2017-10-13
0.3.1

arcondello:


README.rst
```diff
@@ -54,15 +51,15 @@ Install
 
 Compatible with Python 2 and 3.
 
-`pip install dimod`
+>>> pip install dimod
```
orenshk: but this isn't a python interpreter prompt

dimod/responses/numpy_response.py
```diff
@@ -290,11 +290,23 @@ def add_samples_from_array(self, samples, energies, sample_data=None, sorted_by_
 
     def samples_array(self):
         """Returns the :obj:`numpy.ndarray` containing the samples."""
-        return self._samples
+        import numpy as np
```
orenshk: A few general questions:
1. should samples/energy_array be a `@property`? 
2. why store self._samples in a samples name and not just return self._samples?

A nitpick:
`return np.empty((0, 0), dtype=int) if self._samples is None else self.samples`?

PR #48, opened 2018-01-25
Response refactor

arcondello:
The unittests will largely fail because I still need to update all of the reference samplers for the new Response. I'll do that as part of the sampler API update.


dimod/response/response.py
```diff
@@ -0,0 +1,439 @@
...
+        if self._futures:
+            self._add_samples_future()
+
+        return self._data
+
+    @df_data.setter
+    def df_data(self, df_data):
+        self._data = df_data
+
+    def merged(self, sorted_by_energy=True):
```
bellert: Nice!

bellert: Though I did find a case where this breaks... if there is a variable named `'energy'` as well as one named `'energy_data'` this will result in two columns named `'energy_data'`; worse if there is only one named `'energy_data'` it will result in the energies being stored under `'energy'` while another column is called `'energy_data'` (no name conflict, but opposite behaviour of the case where the energies are stored under `'energy_data'` because another column was named `'energy'`)

PR #63, opened 2018-02-09
Structured sampler

arcondello:
* Adds a new Structured base class that can be inherited by Samplers and Composed Samplers
* Separate the Composite and Sampler base classes.

There are many todos still. These will be filled in with subsequent testing

### DIFF
--- a/dimod/classes/structured.py
+++ b/dimod/classes/structured.py
@@ -1,0 +1,101 @@
+"""
+Structured
+==========
+
+A structured sampler can only sample from binary quadratic models with a specific graph.
+
+Creating a Structured Sampler
+-----------------------------
+
+A simple example of a structured sampler is one that only has two variables and one interaction.
+
+>>> class TwoVariables(dimod.Sampler, dimod.Structured):
+...     def __init__(self):
+...         dimod.Sampler()
+...         dimod.Structured(self, [0, 1], [(0, 1)])
+
+This sampler as constructed won't work (because we have not overwritten any of the sample
+methods), but it is sufficient to demonstrate accessing the properties.
+
+First the structure is made a property of the sampler and it can be accessed appropriately.
+
+>>> sampler = TwoVariables()
+>>> sampler.structure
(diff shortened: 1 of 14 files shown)


dimod/classes/structured.py
```diff
@@ -0,0 +1,101 @@
...
+
+class Structured():
+    """todo
+
+    """
+    def __init__(self, nodelist, edgelist, sort_nodes=True, sort_edges=True):
+
+        if sort_nodes:
+            nodelist = sorted(nodelist)
+        elif not isinstance(nodelist, list):
```
bellert: Probably don't need to check if it's a list (just a plain `else`) because I imagine redundantly converting a list to a list is very fast.

dimod/classes/structured.py
```diff
@@ -0,0 +1,101 @@
...
+            nodelist = list(nodelist)
+
+        if sort_edges:
+            edgelist = sorted(tuple(sorted(edge)) for edge in edgelist)
+        elif not isinstance(edgelist, list):
```
bellert: Probably don't need to check if it's a list (just a plain `else`) because I imagine redundantly converting a list to a list is very fast.

dimod/classes/structured.py
arcondello: `list(alist)` makes a shallow copy, so it can be slow.

bellert: Good to know. I'm surprised Python doesn't check for that case.

dimod/classes/structured.py
```diff
@@ -0,0 +1,101 @@
...
+
+Creating a Structured Composite
+-------------------------------
+
+todo
+
+"""
+from collections import namedtuple
+
+Structure = namedtuple("Structure", ['nodelist', 'edgelist', 'adjacency'])
```
randomir: Why not just call them `nodes` and `edges`, instead of `nodelist` and `edgelist`? Is the `list` emphasis important, or any *listy* type suffices?

PR #77, opened 2018-02-17
Refactor @vartype_argument

randomir:
Support default argument values. Use `TypeError` consistently with Python's standard usage.

### DIFF
--- a/dimod/compatibility23.py
+++ b/dimod/compatibility23.py
@@ -20,8 +20,6 @@ if _PY2:
 
     zip_longest = itertools.izip_longest
 
-    RecursionError_ = RuntimeError
-
 else:
 
     range_ = range
@@ -38,8 +36,3 @@ else:
         return iter(d.keys())
 
     zip_longest = itertools.zip_longest
-
-    if sys.version_info.minor > 4:
-        RecursionError_ = RecursionError
-    else:
-        RecursionError_ = RuntimeError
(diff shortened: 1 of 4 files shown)


randomir:
FWIW, the `test_sampler_template_error` test (which induces infinite recursion and waits for `RecursionError`) is failing due to a bug (or at least an unwanted behaviour) in Python3 `inspect` module, [`getfullargspec()`](https://github.com/python/cpython/blob/master/Lib/inspect.py#L1082) function (called from [`getargspec()`](https://github.com/python/cpython/blob/master/Lib/inspect.py#L1051), which is used in the `vartype_argument` decorator around `BinaryQuadraticModel.__init__()`).

Namely, when `getfullargspec()` tries to call a helper function `_signature_from_callable` ([here](https://github.com/python/cpython/blob/master/Lib/inspect.py#L1116)), it catches the base `Exception`, recasting any error (including `RecursionError`) as `TypeError`. The standard (good) practice is to be as specific as possible when catching exceptions.

Note that `RecursionError` is somewhat of a misnomer, since it's *not necessarily caused by a recursion* -- but by a stack frame overflow -- which in this case happens when `_signature_from_callable()` is called.

So, due to all this, the test which expects `RecursionError`, gets `TypeError`, and fails.

We can abandon the feature this pull request brings (default values for constrained arguments with `@vartype_argument`), or we can refactor the base sampler that relies on a double (triple?) recursion that user has to break in his subclass. Or we can do both. :)

Default argument values are nice to have, but (of course) not critical. Regardless, it would probably be good to refactor that default infinite recursion.

randomir:
Replaced implicit sampler infinite loop testing with explicit.

Issue #82, opened 2018-02-21
use getfullargspec in python 3


arcondello:
In python 3.6
```
dimod/dimod/decorators.py:116: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() or inspect.getfullargspec()
```
Relevant documentation:
https://docs.python.org/3/library/inspect.html#inspect.getfullargspec
https://docs.python.org/3/library/inspect.html#inspect.getargspec

arcondello:
merged

PR #83, opened 2018-02-21
Add getargspec to compatibility23.py

arcondello:
#82


arcondello:
@randomir Looks like for our purposes we don't need the extra fields.

PR #93, opened 2018-02-22
Update docstring, example for Vartype Properties

JoelPasvolsky:
@arcondello review this one more carefully please.

### DIFF
--- a/dimod/binary_quadratic_model.py
+++ b/dimod/binary_quadratic_model.py
@@ -163,38 +163,27 @@ class BinaryQuadraticModel(object):
 
     @property
     def spin(self):
-        """:class:`.BinaryQuadraticModel`: The spin-valued version of the binary quadratic model.
+        """:class:`.BinaryQuadraticModel`: An instance of the Ising model subclass
+        of the :class:`.BinaryQuadraticModel` superclass, corresponding to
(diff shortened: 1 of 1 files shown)


dimod/binary_quadratic_model.py
```diff
@@ -222,38 +211,27 @@ def spin(self):
...
-            The energy will correspond.
-
-            >>> bqm = dimod.BinaryQuadraticModel({'a': 0, 'b': 0}, {('a', 'b'): -1}, 0.0, dimod.SPIN)
-            >>> bqm.binary.energy({'a': 0, 'b': 1})
-            1.0
-            >>> bqm.energy({'a': -1, 'b': +1})
-            1.0
+        """:class:`.BinaryQuadraticModel`: An instance of the QUBO model subclass
+        of the :class:`.BinaryQuadraticModel` superclass, corresponding to
+        a binary quadratic model with binary variables.
```
arcondello: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

In most cases the summary line should fit on a single line

JoelPasvolsky: @arcondello, this is the one that's more like an argument. The rest looks okay, so should be ready too merge. Thanks!

PR #103, opened 2018-02-26
Update exceptions

JoelPasvolsky:
@arcondello, note change of "Sampler" to "specified sampler" and "BQM/bqm" to "binary quadratic model" (alternatively can be BinaryQuadraticModel)


JoelPasvolsky:
change of "Sampler" to "specified sampler" --> looking at another place, maybe "Sampler" refers to the Sampler class... Wasn't clear to me from the exception

PR #108, opened 2018-02-27
Conform to boilerplate from template

bellert:


arcondello:
I think you're right. If you close the PR and open another one I think it will ditch my commit. Sorry!

Issue #123, opened 2018-03-02
from_futures should resolve on first response read


arcondello:

arcondello:
merged

Issue #121, opened 2018-03-01
Ignore "third state" of qubit in actual QPU samples


randomir:
`dimod.response.infer_vartype` expects sample values to be binary `{0,1}` or `{-1,1}`, but the real QPU/SAPI response also includes the "third state" for missing qubits.

We can filter those out (in `dimod.Response.from_futures`) or support them (in `infer_vartype`).

arcondello:
merged

PR #124, opened 2018-03-02
Update Response.from_futures to resolve on response read

arcondello:
Addresses #123 #121

### DIFF
--- a/dimod/response.py
+++ b/dimod/response.py
@@ -21,7 +21,7 @@ import itertools
 import concurrent.futures
 
 import numpy as np
-from six import itervalues
+from six import itervalues, iteritems
 
 from dimod.decorators import vartype_argument
 from dimod.vartypes import Vartype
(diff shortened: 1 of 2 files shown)


arcondello:
Ignore the failed travis build, it is some issue with `brew install python3`

dimod/response.py
```diff
@@ -347,31 +350,170 @@ def from_pandas(cls, samples_df, data_vectors, vartype=None, info=None):
...
+                         variable_labels, active_variables):
+
+        # first reset the futures to avoid recursion errors
+        self._futures = {}
 
         # `dwave.cloud.qpu.computation.Future` is not yet interchangeable with
         # `concurrent.futures.Future`, so we need to detect the kind of future
         # we're dealing with.
+        futures = list(futures)  # if generator
         if hasattr(futures[0], 'as_completed'):
```
randomir: We should check if `futures` is an empty list before accessing `[0]`. In the previous version there was a check, but I don't see it here.

PR #148, opened 2018-03-19
Update composed samplers & composites

JoelPasvolsky:
@arcondello For some of these examples I use pseudocode to not lengthen too much and rely on the reference examples for full implmentation


JoelPasvolsky:
@arcondello the failed example is just an illustrative snippet (it's introduced with "Given a sampler, `sampler1`, the composed sampler") so I added # doctest: +SKIP

JoelPasvolsky:
@arcondello

Issue #153, opened 2018-03-21
BinaryQuadraticResponse needs a place to store metadata


arcondello:

arcondello:
merged

Issue #139, opened 2018-03-14
Add `BinaryQuadraticModel.empty()` class method


arcondello:

arcondello:
merged

PR #154, opened 2018-03-21
Update BQM to allow for metadata in an `info` field and to be built from empty

arcondello:
#153 #139

### DIFF
--- a/dimod/binary_quadratic_model.py
+++ b/dimod/binary_quadratic_model.py
@@ -142,6 +142,36 @@ class BinaryQuadraticModel(object):
         self.add_variables_from(linear)
         self.add_interactions_from(quadratic)
 
+    @classmethod
+    def empty(cls, vartype):
+        """Creates an empty BinaryQuadraticModel.
+
+        Equivalent to
+
+        .. code-block:: python
+
+            BinaryQuadraticModel({}, {}, 0.0, vartype)
+
+        Args:
+            vartype (:class:`.Vartype`/str/set):
+                The variable type desired for the binary quadratic model. Accepted input values:
+
+                * :attr:`.Vartype.SPIN`, ``'SPIN'``, ``{-1, 1}``
+                * :attr:`.Vartype.BINARY`, ``'BINARY'``, ``{0, 1}``
+
+        Examples:
+
+            >>> bqm = dimod.BinaryQuadraticModel.empty(dimod.BINARY)
+            >>> bqm.linear
+            {}
+            >>> bqm.quadratic
+            {}
+            >>> bqn.offset
+            0.0
+
+        """
+        return cls({}, {}, 0.0, vartype)
+
     def __repr__(self):
         return 'BinaryQuadraticModel({}, {}, {}, {})'.format(self.linear, self.quadratic, self.offset, self.vartype)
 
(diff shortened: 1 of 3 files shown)


dimod/binary_quadratic_model.py
```diff
@@ -49,6 +49,10 @@ class BinaryQuadraticModel(object):
             * :class:`.Vartype.SPIN`, ``'SPIN'``, ``{-1, 1}``
             * :class:`.Vartype.BINARY`, ``'BINARY'``, ``{0, 1}``
 
+        **kwargs:
+            Any additional keyword parameters and their values will be stored in
```
JoelPasvolsky: Any additional keyword parameters and their values are stored in

dimod/binary_quadratic_model.py
```diff
@@ -123,17 +130,48 @@ class for the K4 complete graph, where the nodes have biases
...
 
     @vartype_argument('vartype')
-    def __init__(self, linear, quadratic, offset, vartype):
+    def __init__(self, linear, quadratic, offset, vartype, **kwargs):
         self.linear = {}
         self.quadratic = {}
         self.adj = {}
         self.offset = offset  # we are agnostic to type, though generally should behave like a number
         self.vartype = vartype
+        self.info = kwargs  # dump any additional kwargs into info
```
JoelPasvolsky: dump  --> place? Add? even "Put" would be nice

dimod/binary_quadratic_model.py
```diff
@@ -123,17 +130,48 @@ class for the K4 complete graph, where the nodes have biases
...
 
         # add linear, quadratic
         self.add_variables_from(linear)
         self.add_interactions_from(quadratic)
 
+    @classmethod
+    def empty(cls, vartype):
+        """Creates an empty BinaryQuadraticModel.
```
JoelPasvolsky: Create an empty binary quadratic model.

dimod/binary_quadratic_model.py
```diff
@@ -123,17 +130,48 @@ class for the K4 complete graph, where the nodes have biases
...
+
+        Equivalent to
+
+        .. code-block:: python
+
+            BinaryQuadraticModel({}, {}, 0.0, vartype)
+
+        Args:
+            vartype (:class:`.Vartype`/str/set):
+                The variable type desired for the binary quadratic model. Accepted input values:
```
JoelPasvolsky: Variable type for the binary quadratic model.

dimod/binary_quadratic_model.py
```diff
@@ -885,6 +923,11 @@ def update(self, bqm):
                 in the updating model are added to the corresponding values in
                 the updated model.
 
+            ignore_info (bool, optional, default=True):
+                If True, the info in the given bqm is ignored, otherwise
```
JoelPasvolsky: If True, `info` in the given model is ignored; otherwise .... updated with the given model's `info`

dimod/binary_quadratic_model.py
arcondello: I copied this from you! :laughing:

arcondello: I changed it in both places.

PR #159, opened 2018-03-29
Embedding

arcondello:
Merge the functionality of `dwave-embedding-utilities` [link](https://github.com/dwavesystems/dwave_embedding_utilities) into dimod


tests/test_embedding.py
```diff
@@ -0,0 +1,546 @@
...
+import numpy.testing as npt
+
+import dimod
+
+try:
+    import networkx as nx
+    _networkx = True
+except ImportError:
+    _networkx = False
+
```
randomir: Since `networkx` is required by a large number of tests below (almost half) which test the basic functionality of `dimod`, I would just put it as a requirement for running tests, and got rid of all `skipUnless` decorators. Unless there's some scenario I'm missing here.

arcondello: I like re-running the tests in a clean venv using `pip install .` as a check on the version ranges.

dimod/embedding/chain_breaks.py
```diff
@@ -0,0 +1,230 @@
...
+
+    for v, chain in iteritems(embedding):
+        vals = [sample[u] for u in chain]
+
+        if _all_equal(vals):
+            unembeded[v] = vals.pop()
+        else:
+            unembeded[v] = _most_common(vals)
+
+    yield unembeded
```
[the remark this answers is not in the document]
arcondello: I have added a short comment, but I agree it should be explained more thoroughly.

PR #165, opened 2018-04-06
Test and fix embedding chain break helpers

randomir:


arcondello:
Duplicates #166 (sorry @randomir !)

Issue #164, opened 2018-04-05
_all_equal fails for empty iterables


arcondello:
https://github.com/dwavesystems/dimod/blob/0f439633046acdf1162936ff694127b367c42a15/dimod/embedding/chain_breaks.py#L231

This should be fixed or at least the docstring should state the limitation.

PR #166, opened 2018-04-06
Address edge cases for embedding utility functions

arcondello:
* _most_common now throws a ValueError for empty sequences
* _all_equal now correctly returns True for empty sequences


arcondello:
Closes #164

PR #183, opened 2018-04-26
Add BinaryQuadraticModel.to_coo and BinaryQuadraticModel.from_coo

arcondello:


dimod/io/coo.py
```diff
@@ -0,0 +1,67 @@
...
+
+def dump(bqm, fp):
+    """Dump a binary quadratic model to a string in COOrdinate format."""
+    for triplet in _iter_triplets(bqm):
+        fp.write('%s\n' % triplet)
+
+
+def loads(s, bqm):
+    """Load a COOrdinate formatted binary quadratic model from a string."""
+    pattern = re.compile(r'^(\d+)[ ]+(\d)+[ ]+([+-]?([0-9]*[.])?[0-9]+)')
```
randomir: Few comments:
- second number is not captured properly if it has multiple-digits (`1 22 3`)
- third number (floating point) can have leading zeros (`0001.2`); scientific notation support?
- to support other whitespace characters (like tab), you could use `\s` instead of `[ ]`.
- `$` anchor is missing
- you might consider stripping whitespace, or adding `\s*` at the beginning and end of regex

dimod/io/coo.py
```diff
@@ -0,0 +1,76 @@
+import re
+
+from dimod.binary_quadratic_model import BinaryQuadraticModel
+
+_LINE_REGEX = r'^\s*(\d+)\s+(\d+)\s+([+-]?([0-9]*[.])?[0-9]+)\s*$'
```
randomir: Ok, so this addresses most of my previous comments. We could also constrain floats to non-zero first digit (for multiple digits), but that would probably just overly-complicate this simple regex, and `float()` will happily parse those anyway. Since COO is not really a specified format, we can be generous on what we parse.

Btw, you can also slightly optimize the parsing by compiling this regex in module scope, not function calls.

randomir: Btw2, you can simplify regex by using `\d` in place of `[0-9]`.

PR #190, opened 2018-05-24
Restructure intro, update BQM and its methods (add, remove, update)

JoelPasvolsky:


dimod/binary_quadratic_model.py
```diff
@@ -106,37 +133,43 @@ class assume that they are numeric.
...
+       set equal to their sequential labels and interactions are the
+       concatenations of the node pairs (e.g., 23 for u,v = 2,3).
+
+       >>> import dimod
+       >>> linear = {1: 1, 2: 2, 3: 3, 4: 4}
+       >>> quadratic = {(1, 2): 12, (1, 3): 13, (1, 4): 14,
+       ...              (2, 3): 23, (2, 4): 24,
+       ...              (3, 4): 34}
+       >>> offset = 0.0
+       >>> vt = dimod.BINARY
```
arcondello: Should spell our `vartype`. We aren't paying by the character :smile: . It keeps it consistent with the input argument names and ideally I should be able to know just from the name what each variable does.

dimod/binary_quadratic_model.py
```diff
@@ -106,37 +133,43 @@ class assume that they are numeric.
...
+       >>> bqm_k4 = dimod.BinaryQuadraticModel(linear, quadratic, offset, vt)
+       >>> bqm_k4.info = {'Complete K4 binary quadratic model.'}
+       >>> bqm_k4.info.issubset({'Complete K3 binary quadratic model.',
+       ...                       'Complete K4 binary quadratic model.',
+       ...                       'Complete K5 binary quadratic model.'})
+       True
+       >>> bqm_k4.adj.viewitems()   # Show all adjacencies  # doctest: +SKIP
```
arcondello: I would not call `viewitems()` directly, just use `items`, specially when skipping for the doctest

dimod/binary_quadratic_model.py
```diff
@@ -157,9 +190,10 @@ def __init__(self, linear, quadratic, offset, vartype, **kwargs):
 
     @classmethod
     def empty(cls, vartype):
-        """Create an empty BinaryQuadraticModel.
+        """Create an empty :class:`.BinaryQuadraticModel` instance.
```
arcondello: I am not sure this needs to be a hyperlink everywhere. Especially when it is a method of the mentioned class. The first line of a docstring is a bit special in that IDEs will use it to prompt the user when autocompleting or whatever. So I would avoid sphinx-specific syntax in the header line

dimod/binary_quadratic_model.py
JoelPasvolsky: Fixed

dimod/binary_quadratic_model.py
JoelPasvolsky: Fixed

dimod/binary_quadratic_model.py
JoelPasvolsky: Fixed

JoelPasvolsky:
@arcondello: dimod now passes doctest without failures

Issue #88, opened 2018-02-21
Add documentation for reference implementations


arcondello:

JoelPasvolsky:
PR #192

PR #192, opened 2018-05-29
Add reference samplers & composites

JoelPasvolsky:
@arcondello some time in the future I'll probably do a 2nd pass on this like I'm doing on the rest of dimod but I'd like to merge it, it's useful content, especially for new users


JoelPasvolsky:
Addresses Issue #88

PR #225, opened 2018-07-17
Add `to_numpy_vectors`, `from_numpy_vectors` to BQM

arcondello:
Add two new method `BinaryQuadraticModel.to_numpy_vectors` and `BinaryQuadraticModel.from_numpy_vectors`. This serves as a sparse alternative to `to_numpy_matrix` and `from_numpy_matrix`.

### DIFF
--- a/dimod/binary_quadratic_model.py
+++ b/dimod/binary_quadratic_model.py
@@ -49,6 +49,8 @@ from __future__ import absolute_import, division
 from collections import Sized, Container, Iterable
 from numbers import Number
 
+import numpy as np
+
 from six import itervalues, iteritems, iterkeys
 
 from dimod.decorators import vartype_argument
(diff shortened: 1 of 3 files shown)


dimod/binary_quadratic_model.py
```diff
@@ -1953,6 +1955,134 @@ def from_numpy_matrix(cls, mat, variable_order=None, offset=0.0, interactions=No
...
+            variable_order (iterable, optional):
+                If provided, labels the variables; otherwise, row/column indices are used.
+
+            dtype (:class:`numpy.dtype`, optional):
+                Data-type of the biases. By default, the data-type is inferred from the biases.
+
+            index_dtype (:class:`numpy.dtype`, optional):
+                Data-type of the indices. By default, the data-type is inferred from the labels.
+
+            sort_indices (bool, optional, default=False):
```
wbernoudy: It might be good to split this into two separate options. Sorting column-wise costs almost nothing, but sorting row-wise in a large dense problem can be expensive.

dimod/binary_quadratic_model.py
```diff
@@ -1953,6 +1955,134 @@ def from_numpy_matrix(cls, mat, variable_order=None, offset=0.0, interactions=No
...
+
+        try:
+            heads, tails, values = quadratic
+        except ValueError:
+            raise ValueError("quadratic should be a 3-tuple")
+
+        if variable_order is None:
+            variable_order = list(range(len(linear)))
+
+        linear = {v: bias for v, bias in zip(variable_order, linear)}
```
wbernoudy: Is it okay that the keys and values of these dicts will retain their `numpy` types? I believe this may cause unexpected behavior for the user (though it should be rare... perhaps something like submitting index arrays of type `np.uint8` and getting overflow later on could be a thing though), and can cause performance issues with dicts (https://stackoverflow.com/a/24705922).

dimod/binary_quadratic_model.py
arcondello: Hmm, I think in this case I am inclined to let the user sort it out themselves.

dimod/binary_quadratic_model.py
arcondello: Will fix

wbernoudy: My bad, I didn't notice the keys are actually Python, so the problems I talked about are not relevant.

PR #234, opened 2018-07-25
Add python3.4 compatibility

arcondello:


arcondello:
@randomir , I was able to speed up most of the osx builds by skipping the homebrew update step. We still need to update homebrew for python 3.7

PR #235, opened 2018-07-26
BSON encoding for BinaryQuadraticModel

wbernoudy:


dimod/binary_quadratic_model.py
```diff
@@ -1612,6 +1612,92 @@ def from_json(cls, obj):
...
 
+    def to_bson(self, fp=None):
+        """Convert a binary quadratic model to an effecient BSON format.
+
+        Args:
+            fp (file, optional):
+                A `.write()`-supporting `file object`_ to save the binary 
+                quadratic model encoded in BSON format to. If not provided, 
+                returns bytes.
+        """
```
randomir: Return value type should **never** depend on argument value. Besides being blasphemous, it's also asking for trouble later on, making API too "magical".

Instead, you should require `fp`, and if a user wants a byte stream instead of file, she can use [`io.BytesIO`](https://docs.python.org/3/library/io.html#io.BytesIO):

```
import io
buf = io.BytesIO()
bqm.to_bson(buf)
```

dimod/binary_quadratic_model.py
```diff
@@ -1612,6 +1612,92 @@ def from_json(cls, obj):
...
+    def from_bson(cls, obj):
+        """Deserialize a binary quadratic model from a BSON encoding.
+
+        Args:
+            obj: (bytes/file):
+                Either bytes or a  `.read()`-supporting `file object`_
+                that represents linear and quadratic biases using the encoding
+                used in `.to_bson`.
+        """
+
```
randomir: Same argument as for `to_json` -- why accept files **and** bytes? Choose one. (Otherwise the game of `isinstance` can escalate quickly.) Preferably, be consistent with the rest of the API and the complementary method in particular.

dimod/binary_quadratic_model.py
```diff
@@ -1612,6 +1612,92 @@ def from_json(cls, obj):
 
         return json.load(obj,  object_hook=bqm_decode_hook)
```
randomir: PR metacomment: I see the pattern of adding convenience to/from methods escalating quickly here.

To be fair, this `to_bson` does add a little bit of an optimization effort on top of `BSON.encode(bqm.to_dict())`, it's the need for BSON support in Ocean community I'm not certain about.

@arcondello Maybe it would make sense to consider having to/from convenience methods as some kind of BQM plugins, since BQM implementation is already pretty large, and it'll only grow over time.

dimod/binary_quadratic_model.py
```diff
@@ -1612,6 +1612,92 @@ def from_json(cls, obj):
...
+
+        Args:
+            fp (file, optional):
+                A `.write()`-supporting `file object`_ to save the binary 
+                quadratic model encoded in BSON format to. If not provided, 
+                returns bytes.
+        """
+                
+
+        import bson
```
randomir: Should we add `bson` dep to `all` extras?

dimod/binary_quadratic_model.py
wbernoudy: I wholeheartedly agree with your first statement. I chose to follow the interface of `to_json`/`from_json` as closely as possibly for consistency. 

I would prefer to keep `fp` optional because though passing in a buffer works fine, I think it's a little more clunky to use than these methods should be. 

Instead I propose always returning `bytes`, and if `fp` is provided, also writing those bytes to the file. If the user just cares about the file, they can ignore the return value. Thoughts?

dimod/binary_quadratic_model.py
arcondello: yes it should be added.

dimod/binary_quadratic_model.py
arcondello: This is already in a sense a plugin, the methods will fail unless the dependencies are installed. I am not sure I see an issue with the number of these things propagating within reason. Especially, as in this case, when they add actual functionality. The BQM object is afterall designed for convenience and transforming to and from other formats is one of its primary purposes.

dimod/binary_quadratic_model.py
arcondello: I don't think I agree so strongly, as is in evidence from `to_json/from_json`, but I am happy enough to make the change for the 0.7.0 release.

In terms of return value, I think my instinct is to always return bytes and the user can then write them to a file if they wish
```
bytes_ = bqm.to_bson()
with open('file') as fp:
    fp.write(bytes_)
```

arcondello: We can then make `to_json` return a json-serializable string always so the pattern would be
```
with open('file') as fp:
    fp.write(json.dumps(bqm.to_json())
```
Which might be less efficient but would be much simpler

arcondello: Unless the argument for passing in `fp` is that the memory footprint is less?

dimod/binary_quadratic_model.py
```diff
@@ -1612,6 +1612,92 @@ def from_json(cls, obj):
...
+
+        num_variables = len(self)
+        if num_variables > 2**16:
+            raise ValueError
+
+        variable_order = sorted(self.linear)
+        num_possible_edges = num_variables*(num_variables - 1) // 2
+        density = len(self.quadratic) / float(num_possible_edges)
```
arcondello: Don't need the `float` cast when `from __future__ import division`

dimod/binary_quadratic_model.py
randomir: Good point about local consistency. It's more important than global. (At least according to PEP-8.)

But I don't see your point on clunkiness. One might say it's clunky to need to open a file before calling `to_bson(fp)`. I'm not saying we should support filenames/paths as arguments, but that *there's beauty in having a method that does one thing, and does it well*.

Honestly, I don't see any benefit of supporting `fp`, i.e.:
```
with open('path') as fp:
    fp.write(bqm.to_bson())
```
vs.
```
with open('path') as fp:
    bqm.to_bson(fp)
```
On contrary, by not accepting `fp`, we have better separation of concerns.

dimod/binary_quadratic_model.py
randomir: In a very stretched sense. All methods still live inside the BQM class, inside one large file.

Even if formats conversion is one of the primary purposes of BQM (wasn't aware of that), I'll argue there's value in splitting those methods somehow, yanking them out of base base def. Just imagine the case of hundreds of conversion methods...

dimod/binary_quadratic_model.py
```diff
@@ -1606,6 +1606,88 @@ def from_json(cls, obj):
...
+            obj: bytes:
+                Byte string that represents linear and quadratic biases using 
+                the encoding used in `.to_bson`.
+        Returns:
+            :class:`.BinaryQuadraticModel`: The corresponding binary quadratic 
+            model.
+        """
+
+        import bson
+        import itertools
```
arcondello: this can/should be imported at the top since it's built in

dimod/binary_quadratic_model.py
```diff
@@ -1606,6 +1606,88 @@ def from_json(cls, obj):
...
+
+        doc = bson.BSON.decode(obj)
+
+        lin = np.frombuffer(doc["linear"], dtype=np.float32)
+        num_variables = len(lin)
+        vals = np.frombuffer(doc["quadratic_vals"], dtype=np.float32)
+        if doc["as_complete"]:
+            i, j = list(zip(*itertools.combinations(range(num_variables), 2)))
```
arcondello: not sure you need the additional `list` call

dimod/binary_quadratic_model.py
```diff
@@ -1606,6 +1606,88 @@ def from_json(cls, obj):
...
+        return bson.BSON.encode(doc)
+
+    @classmethod
+    def from_bson(cls, obj):
+        """Deserialize a binary quadratic model from a BSON encoding.
+
+        Args:
+            obj: bytes:
+                Byte string that represents linear and quadratic biases using 
+                the encoding used in `.to_bson`.
```
arcondello:
```
:meth:`to_bson`
```

dimod/binary_quadratic_model.py
wbernoudy: Correct

PR #236, opened 2018-07-30
to_json/from_json

arcondello:
**This is a backwards compatibility break**

Simplifies the `BinaryQuadraticModel.to_json` method to always return a string. We also remove a lot of boilerplate used to encode bqms with low memory footprint. That case will be covered by #235 .

We will also close #218 as redundant. In order to prevent having too many serialization methods, we don't include `to_dict`/`from_dict`.

### DIFF
--- a/dimod/binary_quadratic_model.py
+++ b/dimod/binary_quadratic_model.py
@@ -1487,6 +1487,63 @@ class BinaryQuadraticModel(Sized, Container, Iterable):
 
         return coo.load(obj, cls.empty(vartype))
 
+    def to_dict(self):
+        """Create a serializable JSON-style dict.
+
+        Returns:
+            dict: A python dictionary of JSON-serializable objects.
+
+        Examples:
+            >>> bqm = dimod.BinaryQuadraticModel.from_ising({0: 1, 2: .1}, {(0, 1): -1})
+            >>> bqm.to_dict()  # doctest: +SKIP
+            {'linear_terms': [{'bias': 1, 'label': 0},
+              {'bias': 0.1, 'label': 2},
+              {'bias': 0.0, 'label': 1}],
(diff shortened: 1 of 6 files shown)


arcondello:
@wbernoudy @randomir 

For `from_json` the default pattern is
```
bqm = BinaryQuadraticModel.from_json(json_string)
```
But I left in the
```
with open('file', 'r') as fp:
    bqm = BinaryQuadraticModel.from_json(fp)
```
pattern. because it doesn't really change the behaviour. It looks a bit odd though with the new `to_json`. Thoughts?

dimod/binary_quadratic_model.py
```diff
@@ -1546,24 +1532,28 @@ def to_json(self, fp=None):
...
+        return json.dumps(self, cls=DimodEncoder, sort_keys=True)
 
     @classmethod
     def from_json(cls, obj):
         """Deserialize a binary quadratic model from a JSON encoding.
 
         Args:
             obj: (str/file):
-                Either a string or a  `.read()`-supporting `file object`_
+                Either a string or a `.read()`-supporting `file object`_
```
randomir: Not sure about this.. On one hand, it's good that `BQM.from_json(bqm.to_json())` works as expected, but on the other, it's inconsistent with `to_json` interface.

randomir:
Re: `to/from_dict`, I would actually prefer to keep those to vanilla `to/from_json`, because they are a more general form. Once you have `dict` of standard objects, it's easy to store that in many other formats (i.e. you get json/bson/yaml/pickle/mongo/postgresql/mysql/redis/memcache/etc adapters with oneliners). I don't mind having performant, specialized converters in addition (as we do), but there's value in providing general base format.

PR #239, opened 2018-08-01
Add BQM.variables property

randomir:
BQM already acts as an iterable container, but `variables` property is more explicit and it supports efficient set operations.


randomir:
Yes, iterating variables with `bqm.linear` seems like an implementation detail. It's not entirely obvious at first that `linear` includes all variables.

PR #240, opened 2018-08-02
Add OrderedBinaryQuadraticModel

### LINKED ISSUE #238: Allow user-specified mapping type for `BinaryQuadraticModel`s
```
bqm = dimod.BinaryQuadraticModel(linear, quadratic, offset, vartype, mapping_type=OrderedDict)
```
This would affect `.linear` and `.quadratic`, possibly `.adj` as well.

arcondello:
Closes #238

### DIFF
--- a/dimod/binary_quadratic_model.py
+++ b/dimod/binary_quadratic_model.py
@@ -2198,12 +2198,16 @@ class BinaryQuadraticModel(Sized, Container, Iterable):
 
 
 class OrderedBinaryQuadraticModel(BinaryQuadraticModel):
-    """Consistently ordered varient of :class:`BinaryQuadraticModel`.
+    """Consistently ordered variant of :class:`BinaryQuadraticModel`.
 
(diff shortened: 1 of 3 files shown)


dimod/binary_quadratic_model.py
```diff
@@ -2195,3 +2195,33 @@ def from_pandas_dataframe(cls, bqm_df, offset=0.0, interactions=None):
             bqm.add_interaction(u, v, 0.0)
 
         return bqm
+
+
+class OrderedBinaryQuadraticModel(BinaryQuadraticModel):
+    """Consistently ordered varient of :class:`BinaryQuadraticModel`.
```
randomir: variant

Also, mention how are they ordered -- by order of insertion. Which is well defined if adding variables/interactions one by one (`.add_variable`), but not if constructed from an unordered mapping provided for `linear`/`quadratic` to init.

PR #242, opened 2018-08-02
WIP: coerce embedding into desired vartype

### LINKED ISSUE #204: dimod.embed_bqm: Couplings/biases out of j/h range after embedding
For example, running sampling on `qbsolv/tests/qubos/bqp100_1.qubo`, we get embedded BQM with coupling strengths from [-200, 200] range and biases from [-4, 8]. Both out of allowed range for a particular QPU solver.

The error message on server side is also confusing: it says out-of-range values are discarded and after that they are scaled.

### LINKED ISSUE #241: Embedding QUBO leads to out-of-range linear terms.
If I determine the correct scaling factor to scale a QUBO such that the Ising form fits in the QPU range, I should be able to scale the QUBO by this amount, embed it, and solve the Ising.  However, the embedding causes some linear QUBO terms to be out of range in the resulting Ising model.  As discussed with @arcondello.

https://github.com/dwavesystems/dimod/blob/8ebfffa42319aa4850cfc5a1c99a8711eac44722/dimod/embedding/transforms.py#L125

arcondello:
Closes #241 
Closes #204

### DIFF
--- a/tests/test_embedding_transforms.py
+++ b/tests/test_embedding_transforms.py
@@ -384,3 +384,25 @@ class TestEmbedBQM(unittest.TestCase):
 
         for v in itertools.chain(*embedding.values()):
             self.assertIn(v, target_bqm)
+
+    @unittest.skipUnless(_networkx, "No networkx installed")
+    def test_energy_range_embedding(self):
+        # start with an Ising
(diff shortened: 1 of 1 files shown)


arcondello:
Proposed solution, either:
1) Always embed the bqm in spin-space
2) Allow the user to specify in which space to embed

PR #233, opened 2018-07-25
Convert Response object to use numpy record array to store samples and data vectors

arcondello:
Closes #232

PR #243, opened 2018-08-02
Refactor Response

### LINKED ISSUE #92: Printing the Response object should print the variable labels if present

### LINKED ISSUE #127: Response.from_futures info_keys should have an 'all' option

### LINKED ISSUE #228: Mention `num_occurrences` in docs for `dimod.Response`
It's not clear from the docs how to get `num_occurrences`, alongside list of samples.

User's [comment here](https://support.dwavesys.com/hc/en-us/community/posts/360013878673/comments/360001418813).

### LINKED ISSUE #232: PendingDeprecationWarning
The Matrix subclass is being deprecated by numpy

### LINKED ISSUE #237: `Response.from_futures` parameter `info_keys` is unnecessarily inconvenient and inconsistent
https://github.com/dwavesystems/dimod/blob/3b4243f81cf0c6d9c1e66dae5f9ea75d4a438d48/dimod/response.py#L442

Currently, the `info_keys` parameter requires you to know which which keys in the return value from your future will correspond to which keys in the `Response.info` dict. This is quite inconvenient if you are writing code that is intended to be agnostic to an underlying sampler, as you may not know what "format" the sampler will return for information about the samples. Furthermore, this parameter is inconsistent with `Response.from_dicts`, which simply accepts a `info` parameter.

Proposed solution: change `info_keys=None` to `info_key="info"`, such that when the future returns, dimod will simply take the value of `future.result[info_key]` (if it exists) and stick it into the info of the newly created `Response`.

arcondello:
Closes #233 
Closes #228 
Closes #92 
Closes #127 
Closes #237 
Closes #232


arcondello:
@randomir @wbernoudy , I am still working on the docstrings, but I think the logic is mostly done. The main point of interest is the `Response` object in `response.py`. All the other changes are just adapting to that.

dimod/binary_quadratic_model.py
```diff
@@ -2073,11 +2074,14 @@ def from_numpy_vectors(cls, linear, quadratic, offset, vartype, variable_order=N
             :obj:`.BinaryQuadraticModel`
 
         Examples:
+            >>> import dimod
+            >>> import numpy as np
+            ...
             >>> linear_vector = np.asarray([-1, 1])
-            >>> quadratic_vectors = (np.asarray([0]), np.asarray([1]), np.asarray([-1.0])))
+            >>> quadratic_vectors = (np.asarray([0]), np.asarray([1]), np.asarray([-1.0]))
```
randomir: Sidenote: outer parens unnecessary.

dimod/response.py
```diff
@@ -13,737 +13,469 @@
...
+            >>> response.record.num_occurrences
+            array([1, 1, 1, 1])
 
         """
-        if self._futures:
-            self._resolve_futures(**self._futures)
+        if hasattr(self, '_future'):
+            self._resolve_future()
+        # subsequent calls will bypass the _future check
+        self.__dict__['record'] = record = self._record
```
randomir: Not sure this is a smart thing to do on properties (or in general).

dimod/response.py
```diff
@@ -13,737 +13,469 @@
...
-
-        if vartype is None:
-            vartype = infer_vartype(samples_matrix)
-
-        response = cls(samples_matrix, data_vectors=data_vectors,
-                       vartype=vartype, info=info, variable_labels=variable_labels)
-
-        return response
+        if n is None:
+            for sample, in self.data(['sample'], sorted_by=sorted_by, name=None):
```
randomir: Syntax error? Or at least runtime error?

dimod/response.py
```diff
@@ -13,737 +13,469 @@
...
         """
-        samples = iter(samples)
+        record = self.record
 
-        # get the first sample
-        first_sample = next(samples)
+        if fields is None:
+            # make sure that sample, energy is first
+            fields = ['sample', 'energy', 'num_occurrences'] + [field for field in record.dtype.fields
+                                                                if field not in {'sample', 'energy', 'num_occurrences'}]
```
randomir: Maybe keep a list of these default fields somewhere, class-local, or module-local?

dimod/response.py
arcondello: I am not sure what you mean here

dimod/response.py
```diff
@@ -74,14 +74,15 @@ class Response(Iterable, Sized):
         {'a': -1, 'b': 1} 1.0
 
     """
+    _REQUIRED_FIELDS = ['sample', 'energy', 'num_occurrences']
```
randomir: Even better, make it a `set` here, you use it for membership testing later on, either way.

arcondello: Need it to be ordered

PR #251, opened 2018-08-09
Fix illegal argument of change_vartype in Sampler.sample

kotarotanahashi:
**Issue**
* The illegal argument: `data_vector_offsets` of `change_vartype` is used in `Sampler.sample`
* There is no test to call `Sampler.sample` from `Sampler.sample_qubo` or `Sampler.sample_ising`

**Fix**
* Replaced  `data_vector_offsets` with `energy_offset` which is correct argument of `change_vartype`
* Added tests to call `Sampler.sample`


arcondello:
Good catch! Thank you!

PR #263, opened 2018-08-15
Add support for headers in BQM COO methods

### LINKED ISSUE #262: Support metadata in COO (like vartype)
**Application**
Currently, (de-)serialization of BQMs is partial and with loss of information - vartype is not stored/loaded.

It would be nice to have a complete (de-)serialization of BQM to COO format, like we already have for JSON and BSON. E.g.:
```
bqm = BinaryQuadraticModel.from_coo(fp)
bqm.to_coo(fp)
```

**Proposed Solution**
Support `vartype` via `p` or `c` in COO.

arcondello:
Closes #262

### DIFF
--- a/dimod/io/coo.py
+++ b/dimod/io/coo.py
@@ -18,7 +18,7 @@ import re
 
 from dimod.binary_quadratic_model import BinaryQuadraticModel
 
-_LINE_REGEX = r'^\s*(\d+)\s+(\d+)\s+([+-]?([0-9]*[.])?[0-9]+)\s*$'
+_LINE_REGEX = r'^\s*(\d+)\s+(\d+)\s+([+-]?(?:[0-9]*[.]):?[0-9]+)\s*$'
 """
 Each line should look like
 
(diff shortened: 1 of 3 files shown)


dimod/io/coo.py
```diff
@@ -27,58 +27,70 @@
...
     for line in fp:
-        match = pattern.search(line)
+        triplets.extend(pattern.findall(line))
+
+        vt = vartype_pattern.findall(line)
+        if vt:
+            if vartype is None:
+                vartype = vt[0]
+            elif vartype is not vt[0]:
+                raise ValueError("vartypes from headers and/or inputs do not match")
```
randomir: I would maybe allow kwarg to override the vartype from file.
Just like you're allowing vartype at the end of file to override vartype at the beginning.

arcondello: In this case I am allowing multiple vartype references but they all must be consistent.

randomir: What's the point of override if it has to be consistent? :)

PR #266, opened 2018-08-21
Add chain_break_fraction kwarg to unembed_response

### LINKED ISSUE #246: Feature Request: Throw warning when unembedding results in too many chain breaks
Proposed syntax:
```
resp = unembed_response(response, embedding, bqm, warning_threshold=.1)
```
In this case a `UserWarning` will be raised if more than 10% of the chains are broken.

arcondello:
Closes #246 

We do not throw a warning here, that is left to the calling method/composite to do.


arcondello:
The embedding composites will set it to be true

PR #273, opened 2018-09-05
Add dtype to BinaryQuadraticModel construction

### LINKED ISSUE #254: Feature Request: Add dtype to BinaryQuadraticModel construction
Right now the biases in `BinaryQuadraticModel`s can be arbitrary types which is convenient in many cases, however there are times that it would be nice to be able to specify a consistent type for biases.

**Proposed solution**
Add a `dtype` kwarg to the bqm construction methods.

```
bqm = BinaryQuadraticModel(linear, quadratic, offset, vartype, dtype=float)
```
The default value would be `dtype = object` which would be the current behaviour. In general we would assume that it is a callable (e.g. `float`, `int`, `np.int8`).

It would be nice if we also allowed [numpy data type](https://docs.scipy.org/doc/numpy/reference/arrays.dtypes.html) syntax so
```
bqm = BinaryQuadraticModel(linear, quadratic, offset, vartype, dtype='float64')
```
would be valid.

Subsequent `.add_variable` or `.add_interaction` calls would coerce to the specified datatype if possible.
```
>>> bqm = BinaryQuadraticModel.empty(dimod.SPIN, dtype=float)
>>> bqm.add_variable('a', 1)
>>> isinstance(bqm.linear['a'], float)
True
```

arcondello:
Closes #254 

Allow the binary quadratic model to have a single data type. Biases are coerced to the correct type when added to the model.

So far I have only changed the `.__init__` and `.empty` constructors. There will be a new issue generated to address the other construction methods.

### DIFF
--- a/dimod/binary_quadratic_model.py
+++ b/dimod/binary_quadratic_model.py
@@ -75,19 +75,19 @@ class BinaryQuadraticModel(Sized, Container, Iterable):
             the binary quadratic model and values the linear biases associated
             with these variables.
             A variable can be any python object that is valid as a dictionary key.
-            Biases are generally numbers but this is not explicitly checked.
+            Biases are generally numbers, see dtype below.
 
         quadratic (dict[(variable, variable), bias]):
             Quadratic biases as a dict, where keys are
             2-tuples of variables and values the quadratic biases associated
             with the pair of variables (the interaction).
             A variable can be any python object that is valid as a dictionary key.
-            Biases are generally numbers but this is not explicitly checked.
+            Biases are generally numbers, see dtype below.
             Interactions that are not unique are added.
 
         offset (number):
             Constant energy offset associated with the binary quadratic model.
-            Any input type is allowed, but many applications assume that offset is a number.
+            Offsets are generally numbers, see dtype below.
             See :meth:`.BinaryQuadraticModel.energy`.
 
         vartype (:class:`.Vartype`/str/set):
@@ -95,6 +95,11 @@ class BinaryQuadraticModel(Sized, Container, Iterable):
 
             * :class:`.Vartype.SPIN`, ``'SPIN'``, ``{-1, 1}``
             * :class:`.Vartype.BINARY`, ``'BINARY'``, ``{0, 1}``
+
+        dtype (type/function, optional, default=object):
+            The type used for biases in the binary quadratic model. If :class:`object`, then biases
+            can be of any type, though in general only numeric types will work. Otherwise dtype
+            is called on each bias to coerce it into the appropriate type.
 
         **kwargs:
             Any additional keyword parameters and their values are stored in
@@ -151,6 +156,9 @@ class BinaryQuadraticModel(Sized, Container, Iterable):
             adjacent to `v` (e.g. `u`) and values are quadratic biases associated
             with the pair of inner and outer keys (`u, v`).
 
+        dtype (type):
+            The bias type.
+
         info (dict):
             A place to store miscellaneous data about the binary quadratic model
             as a whole.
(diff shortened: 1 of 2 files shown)


dimod/binary_quadratic_model.py
```diff
@@ -194,20 +202,26 @@ class for the K4 complete graph, where the nodes have biases
...
+    def __init__(self, linear, quadratic, offset, vartype, dtype=object, **kwargs):
         self.linear = {}
         self.quadratic = {}
         self.adj = {}
-        self.offset = offset  # we are agnostic to type, though generally should behave like a number
+
+        if dtype is object:
+            self.offset = offset  # we are agnostic to type, though generally should behave like a number
+        else:
+            self.offset = dtype(offset)
```
m3ller: Is self.offset an actual number (the energy offset) or a data type?

dimod/binary_quadratic_model.py
```diff
@@ -224,6 +238,11 @@ def empty(cls, vartype):
                 * :attr:`.Vartype.SPIN`, ``'SPIN'``, ``{-1, 1}``
                 * :attr:`.Vartype.BINARY`, ``'BINARY'``, ``{0, 1}``
 
+            dtype (type/function, optional, default=object):
```
m3ller: Just out of curiosity, `empty(cls, vartype, dtype=object)` creates a BQM with no bias values, correct? Is the `dtype` here for the future, when the user decides to add in bias values later on?

dimod/binary_quadratic_model.py
```diff
@@ -496,7 +518,7 @@ def add_interaction(self, u, v, bias, vartype=None):
             bias (bias):
                 Quadratic bias associated with u, v. If u, v is already in the model, this value
                 is added to the current quadratic bias. Many methods and functions expect `bias` to
-                be a number but this is not explicitly checked.
+                to be numbers but this is not explicitly checked unless a dtype has been provided.
```
m3ller: Minor thing: there are two to's.  "... to to be numbers"

tests/test_binary_quadratic_model.py
```diff
@@ -1311,6 +1311,159 @@ def test_binary_property_relabel(self):
         self.assertIn('a', model.spin.linear)
         self.assertNotIn(0, model.spin.linear)
 
+    def test_dtype_basic_construction_float(self):
+        linear = {'a': 1, 'b': 1., 'c': np.int8(1), 'd': np.float(1)}
+        quadratic = {'ab': 1, 'bc': 1., 'cd': np.uint(16), 'da': np.float64(1)}
```
m3ller: In `linear` and `quadratic`, why are `b` and `bc` marked as "1.", while `a` and `ab` are "1"?

dimod/binary_quadratic_model.py
arcondello: An actual number.

dimod/binary_quadratic_model.py
arcondello: That's correct. The dtype is a property of the bqm as a whole, so future biases added are coerced to the appropriate type.

dimod/binary_quadratic_model.py
```diff
@@ -1286,23 +1311,23 @@ def binary_to_spin(linear, quadratic, offset):
...
 
         for u, bias in iteritems(linear):
-            h[u] = .5 * bias
+            h[u] = bias / 2
             linear_offset += bias
 
         for (u, v), bias in iteritems(quadratic):
 
-            J[(u, v)] = .25 * bias
+            J[(u, v)] = bias / 4
```
wbernoudy: This will allow for some strange behavior, for example:

```python
>>> qubo_bqm = dimod.BinaryQuadraticModel({}, {(0, 1): 3}, 0, vartype=dimod.BINARY, dtype=np.int64)
>>> ising_bqm = qubo_bqm.change_vartype(dimod.SPIN)
>>> ising_bqm
BinaryQuadraticModel({0: 0, 1: 0}, {(0, 1): 0}, 0.75, Vartype.SPIN)
```

Note that the coupler bias has been erased due to calling `np.int64(0.75)`. Perhaps more importantly,
```python
>>> type(ising_bqm.offset)
numpy.float64
```
Looks like `add_offset` needs to do the same coercion to the dtype that the other `add` methods do.

Issue #216, opened 2018-07-06
Add make_quadratic to the documentation


arcondello:
https://github.com/dwavesystems/dimod/blob/54303f184427d1b3a2741cfb7ea1b1263f68a361/dimod/higherorder.py#L82

arcondello:
Needs to be added to the read-the-docs as well as getting a documentation refactor.

PR #276, opened 2018-09-12
fix variables function

arcondello:
Add `fix_variables` function which in polynomial time can determine some variables in a bqm to fix.

This change also adds C++ code to dimod, necessitating some CI infrastructure changes.

This PR does not update the deploy in CI.


dimod/roof_duality/fix_variables.py
```diff
@@ -0,0 +1,68 @@
+from dimod.roof_duality._fix_variables import fix_variables_wrapper
+from dimod.vartypes import Vartype
+
+
+def fix_variables(bqm, sampling_mode=True):
+    """Determine assignments for some bqm variables.
```
JoelPasvolsky: BQM, and elsewhere

dimod/roof_duality/fix_variables.py
```diff
@@ -0,0 +1,68 @@
...
+
+    fix_variables uses maximum flow in the implication network to correctly fix variables (that is,
+    one can find an assignment for the other variables that attains the optimal value). The
+    variables that roof duality fixes will take the same values in all optimal solutions.
+
+    Args:
+        bqm (:obj:`.BinaryQuadraticModel`)
+            A binary quadratic model.
+
+        sampling_mode (bool, optional, default=True):
```
JoelPasvolsky: sampling_mode description is unclear. What is sampling mode (i.e., what does it do for true vs false)?

dimod/roof_duality/fix_variables.py
```diff
@@ -0,0 +1,68 @@
...
+        dict: Variable assignments for some bqm variables.
+
+    Examples:
+
+        >>> bqm = dimod.BinaryQuadraticModel.empty(dimod.SPIN)
+        >>> bqm.add_variable('a', 1.0)
+        >>> dimod.fix_variables(bqm)
+        {'a': -1}
+
+        There are two ground states, so no variables fixed.
```
JoelPasvolsky: In the next example, there

(Otherwise looks like it's for the previous example.)

arcondello:
Just realised I still need to add the Apache headers, will do so before merging

dimod/roof_duality/fix_variables.py
randomir: Agree, it's not clear you get multiple samples for the same ground state energy when `sampling_mode=True`.

arcondello:
closes #216

PR #280, opened 2018-09-22
Generate BQMs

arcondello:
Add generation functions for bqms.


dimod/generators/fcl.py
```diff
@@ -0,0 +1,155 @@
...
+    adj = {v: set() for v in nodes}
+    for u, v in edges:
+        if u in adj:
+            adj[u].add(v)
+        else:
+            adj[u] = {v}
+        if v in adj:
+            adj[v].add(u)
+        else:
+            adj[v] = {u}
```
m3ller: For lines 78-85, perhaps instead of if-else's, it might be better to use try-except's. That way, you won't need to search for `u` or `v` in `adj` on each iteration.

i.e.
```
try:
  adj[u].add(v)
except KeyError:
  adj[u] = {v}
```

m3ller: On second thought, this is just a style choice. It won't make a huge difference either way, so feel free to ignore the try-except suggestion.

arcondello: try-catch is faster for the try block and slower for the catch block. In this case I expect to hit the catch block fairly often

dimod/decorators.py
```diff
@@ -176,3 +179,63 @@ def new_f(*args, **kwargs):
         return new_f
 
     return _vartype_arg
+
+
+def graph_argument(*arg_names):
```
randomir: Too bad `vartype_argument` code had to be duplicated here.

dimod/generators/random.py
```diff
@@ -0,0 +1,91 @@
...
+            The low end of the range for the random biases.
+
+        high (float, optional, default=1.0):
+            The high end of the range for the random biases.
+
+        cls (:class:`.BinaryQuadraticModel`):
+            Binary quadratic model class to build from.
+
+    """
+    nodes, edges = graph
```
randomir: I don't see this working if `graph` is `int`.

dimod/generators/random.py
```diff
@@ -0,0 +1,91 @@
...
+            The low end of the range for the random biases.
+
+        high (float, optional, default=1):
+            The high end of the range for the random biases.
+
+        cls (:class:`.BinaryQuadraticModel`):
+            Binary quadratic model class to build from.
+
+    """
+    nodes, edges = graph
```
randomir: Same as above (`int` support).

dimod/generators/random.py
```diff
@@ -0,0 +1,91 @@
...
+
+@graph_argument('graph')
+def uniform(graph, vartype, low=0.0, high=1.0, cls=BinaryQuadraticModel):
+    """Generate a bqm with random biases and offset.
+
+    Biases and offset are drawn from a uniform distribution range (low, high).
+
+    Args:
+        graph (int/tuple[nodes, edges]/:obj:`~networkx.Graph`):
+            The graph to build the frustrated loops on. Either an integer n, interpreted as a
```
randomir: frustrated loops?

dimod/generators/fcl.py
```diff
@@ -0,0 +1,155 @@
...
+            Maximum number of failures to find a cycle before terminating.
+
+    .. [HJARTL] Hen, I., J. Job, T. Albash, T.F. Rønnow, M. Troyer, D. Lidar. Probing for quantum
+        speedup in spin glass problems with planted solutions. https://arxiv.org/abs/1502.01663v2
+
+    .. [KLH] King, A.D., T. Lanting, R. Harris. Performance of a quantum annealer on range-limited
+        constraint satisfaction problems. https://arxiv.org/abs/1502.02098
+
+    """
+    nodes, edges = graph
```
randomir: What if graph is `int`?

dimod/generators/random.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    nodes, edges = graph
+    return cls(((n, random.uniform(low, high)) for n in nodes),
+               ((u, v, random.uniform(low, high)) for u, v in edges),
+               random.uniform(low, high),
+               vartype)
+
+
+@graph_argument('graph')
+def randint(graph, vartype, low=0, high=1, cls=BinaryQuadraticModel):
```
randomir: Maybe combine with `uniform`, and parametrize on RNG function?

dimod/generators/fcl.py
arcondello: decorator handles this case

dimod/decorators.py
arcondello: we would look at making a generic arg coercer decorator, maybe if we get a third

Issue #281, opened 2018-09-27
Consider removing the boost dependency


arcondello:
The roof duality functions rely on several boost graph objects. This makes installing from sdist problematic, also building on read the docs. We should determine if we can remove the boost dependency without significantly affecting performance or the clarity of the code.

arcondello:
Upon further inspection, the two boost functions `strong_components` and `push_relabel_max_flow` would be difficult to replace without significant additional code needed. With the changes added by #284 I am closing this issue as being infeasible.

PR #284, opened 2018-09-28
Allow c++ extension build to fail gracefully.

arcondello:
If for some reason the `python setup.py build_ext` command fails (either no available compiler or boost is not installed), the roof_duality extension is ignored.

Related to #281

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,8 @@ import os
 
 from setuptools import setup
 from distutils.extension import Extension
+from distutils.command.build_ext import build_ext
+from distutils.errors import CCompilerError, DistutilsExecError, DistutilsPlatformError
 
 # add __version__, __author__, __authoremail__, __description__ to this namespace
 _PY2 = sys.version_info.major == 2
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -84,22 +86,93 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+
+if sys.platform == 'win32' and sys.version_info < (2, 7):
```
randomir: I guess we don't really need this block, since we supporting only 2.7+

PR #287, opened 2018-10-02
Update template (leap)

JoelPasvolsky:
@arcondello, added leap & friends


test.qubo
```diff
@@ -0,0 +1,15 @@
+# vartype=BINARY
```
randomir: I like how our little header is taking off :)

PR #290, opened 2018-10-11
SampleSet.to_serializable and SampleSet.from_serializable

arcondello:


dimod/io/json.py
```diff
@@ -68,6 +125,24 @@ def bqm_decode_hook(dct, cls=None):
     return dct
 
 
+def sampleset_decode_hook(dct, cls=None):
```
m3ller: What does `dct` and `cls` mean? Is it fairly obvious for people in the know?

arcondello: Yeah, I am definitely being sloppy here. These are super internal functions though, so it's just us who will feel the pain. I will generate an issue to get this cleaned up.

PR #303, opened 2018-10-24
Update bqm, samplers/composites, sampleset

JoelPasvolsky:
@arcondello I'll break this into parts to keep PR sizes tolerable

### DIFF
--- a/dimod/binary_quadratic_model.py
+++ b/dimod/binary_quadratic_model.py
@@ -166,6 +166,7 @@ class BinaryQuadraticModel(Sized, Container, Iterable):
        concatenations of the node pairs (e.g., 23 for u,v = 2,3).
 
        >>> import dimod
+       ...
        >>> linear = {1: 1, 2: 2, 3: 3, 4: 4}
        >>> quadratic = {(1, 2): 12, (1, 3): 13, (1, 4): 14,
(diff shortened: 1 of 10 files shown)


dimod/reference/samplers/exact_solver.py
```diff
@@ -30,7 +30,7 @@
 
 
 class ExactSolver(Sampler):
-    """A simple exact solver for testing and debugging.
+    """A simple exact solver for testing and debugging code using your local CPU.
```
randomir: I don't think we can emphasize enough this is only for testing and debugging on small-scale problems.

Also, I think it's worth mentioning here `ExactSolver` is a python implementation of exhaustive (brute force) search over all possible samples. (Remember how one of the "influencers" misinterpreted the list of samples returned by exact solver.)

dimod/binary_quadratic_model.py
```diff
@@ -166,6 +166,7 @@ class for the K4 complete graph, where the nodes have biases
        concatenations of the node pairs (e.g., 23 for u,v = 2,3).
 
        >>> import dimod
+       ...
```
randomir: I think it's safe to drop `import <module>` from examples in documentation of `<module>`.

arcondello: I agree

PR #310, opened 2018-10-26
Allow interactions/variables to be excluded when scaling bqms

### LINKED ISSUE #269: Allow scaling of subset of bqm biases
**Application**
In the case of embedded bqm's, it would be useful to be able to scale the original biases without scaling the chain biases

**Proposed Solution/Alternatives Considered**
Pattern A:
```
bqm.lock_interactions([('a', 'b'), ('b', 'c')])
bqm.scale(.5)
```

Pattern B:
```
bqm.scale(.5, exclude_interactions=[('a', 'b'), ('b', 'c')])
```

@conta877

arcondello:
Closes #269

### DIFF
--- a/dimod/binary_quadratic_model.py
+++ b/dimod/binary_quadratic_model.py
@@ -844,12 +844,18 @@ class BinaryQuadraticModel(Sized, Container, Iterable):
         """
         self.add_offset(-self.offset)
 
-    def scale(self, scalar):
+    def scale(self, scalar, ignored_variables=None, ignored_interactions=None):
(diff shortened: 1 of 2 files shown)


arcondello:
Specifically looking for feedback on the parameter name

m3ller:
Perhaps `ignored_variables` and `ignored_interactions`? I like "ignore" more than "exclude" because "exclude" seems like an active thing you're doing to the list, while "ignore" just means you won't be touching that list.

arcondello:
`skip_*` sounds boolean to me. I'll go with `ignored_*`

PR #311, opened 2018-11-06
Add BinaryQuadraticModel bias normalization method

wbernoudy:


arcondello:
It would be good if we could specify the ranges for linear quadratic as different and it scales to fit within both.

dimod/binary_quadratic_model.py
```diff
@@ -891,6 +891,40 @@ def scale(self, scalar):
         except AttributeError:
             pass
 
+    def normalize(self, a=1):
```
randomir: More descriptive name for `a`? Also, would asymmetric range make sense? `[a, b]`?

wbernoudy:
@arcondello That's correct, `bias_range` will be ignored (actually, it's ignored if even just one of linear or quadratic ranges are provided). The behavior is described in the docstring.

Issue #186, opened 2018-05-03
Ability to add variable and interaction data in BinaryQuadraticModel


arcondello:
For instance

```
import dimod

bqm = dimod.BinaryQuadraticModel.empty(dimod.SPIN)

bqm.add_variable(v, .5, data={'somedata': 0})
bqm.add_interaction(u, v, .5, data={'someotherdata': 1})
```

PR #323, opened 2018-12-17
Store BQM biases in a single data structure

arcondello:
We now used views for bqm.linear, bqm.adjacency, bqm.quadratic. This way we can keep all the biases in a single bqm._adj dict-of-dicts.

We also added functionality for keeping bias and interaction data see #186 though this is not fully implemented yet.

This supersedes some of the functionality in #312  which will need to be rebased.


arcondello:
This is now a backwards compatibility break as we have removed the OrderedBQM object.

dimod/views.py
```diff
@@ -72,3 +74,239 @@ class IndexValuesView(abc.ValuesView):
...
+            raise KeyError('{} cannot have an interaction with itself'.format(u))
+
+        adj = self._adj
+        
+        # we don't know what type we want the biases, so we require that the variables already
+        # exist before we can add an interaction between them
+        if u not in adj:
+            raise KeyError('{} is not already a variables in the binary quadratic model'.format(u))
+        if v not in adj:
+            raise KeyError('{} is not already a variables in the binary quadratic model'.format(v))
```
m3ller: Typo in error messages: `a variables` -> `a variable`

dimod/views.py
```diff
@@ -72,3 +74,239 @@ class IndexValuesView(abc.ValuesView):
...
+    """Faster items iteration"""
+    __slots__ = ()
+
+    def __iter__(self):
+        adj = self._mapping._adj
+        for u, v in self._mapping:
+            yield (u, v), adj[u][v].value
+
+class NeighbourView(abc.Mapping):
+    """Acts as a dictionary `{u: bias, ...}` for the neightbours of a variables `v`.
```
m3ller: typo: `neightbours` -> `neighbours`. `a variables` -> `a variable`

dimod/views.py
```diff
@@ -72,3 +74,239 @@ class IndexValuesView(abc.ValuesView):
...
+    __slots__ = '_adj', '_var'
+
+    def __init__(self, adj, v):
+        self._adj = adj
+        self._var = v
+
+    def __getitem__(self, v):
+        u = self._var
+        if u == v:
+            raise KeyError
```
m3ller: Perhaps there could be a message with the KeyError? Like `"{} is not a neighbour of itself".format(u)`

dimod/binary_quadratic_model.py
```diff
@@ -236,32 +246,40 @@ def __repr__(self):
...
+        return iter(self.adj)
 
     @property
     def variables(self):
         """Return binary quadratic model's variables as a dictionary view object."""
-        return viewkeys(self.linear)
+        if PY2:
+            return set(self.linear)
+        else:
+            return self.linear.keys()
```
randomir: Since `LinearView` is a mapping, we could add `.keys()` there (as a proxy to `adj.keys()`), and then here use:
```
return self.linear
```
Would that work?

dimod/views.py
```diff
@@ -72,3 +74,239 @@ class IndexValuesView(abc.ValuesView):
     def __iter__(self):
         # Inherited __init__ puts the Mapping into self._mapping
         return iter(self._mapping._data.flat)
+
+class Bias(object):
+    """Biases and data associated with a variable or intereraction."""
```
randomir: typo

dimod/views.py
```diff
@@ -72,3 +74,239 @@ class IndexValuesView(abc.ValuesView):
     def __iter__(self):
         # Inherited __init__ puts the Mapping into self._mapping
         return iter(self._mapping._data.flat)
+
+class Bias(object):
```
randomir: Wrapping all biases with a class seems awfully slow (especially in the average case where the bias is a primitive type like `int` or `float`).

dimod/views.py
```diff
@@ -72,3 +74,239 @@ class IndexValuesView(abc.ValuesView):
...
+
+            for v in neigh:
+                if u == v:
+                    # not adjacent to itself
+                    continue
+
+                if v not in seen:
+                    yield (u, v)
+
+            seen.add(u)
```
randomir: It would be nice if iteration would have a guaranteed/stable order, i.e. if it would step through sorted pairs of a triangular matrix. That would also resolve the O(N^3) time complexity this iterator has.

dimod/binary_quadratic_model.py
```diff
@@ -526,64 +543,34 @@ def add_interaction(self, u, v, bias, vartype=None):
...
 
+        # so that they exist
+        self.add_variable(u, 0.0)
+        self.add_variable(v, 0.0)
+
         if vartype is not None and vartype is not self.vartype:
             if self.vartype is Vartype.SPIN and vartype is Vartype.BINARY:
                 # convert from binary to spin
-                bias /= 4.
+                bias /= 4
```
randomir: Thanks for removing the hanging dots here and below. My eyes are thankful. :)

dimod/binary_quadratic_model.py
```diff
@@ -236,32 +246,40 @@ def __repr__(self):
...
+                return False
+
+            if self.offset != other.offset:
+                return False
+
+            if self.linear != other.linear:
+                return False
+
+            return self.adj == other.adj
+        except AttributeError:
```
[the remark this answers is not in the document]
arcondello: If `bqm == 1` say. I originally checked `isinstance(other, BinaryQuadraticModel)` but this seems more permissive.

dimod/binary_quadratic_model.py
arcondello: That would work, things like
`1 in bqm.variables` would work fine, but if you did `print(bqm.variables)` it might be confusing.

dimod/binary_quadratic_model.py
arcondello: If the biases are floats it actually speeds things up to have the dots! But it also would promote types to floats which is bad behaviour.

arcondello:
So, following up on our previous conversation, we could speed things up by storing the variables in _adj as upper triangular matrix, sorting on the hash. We then would keep only a single reference to each of the biases which would eliminate the need for the `Bias` class. This would speed things up, but we would lose the per-bias data field. I'll run some experiments before merging.

arcondello:
Right, upon further investigation, storing upper triangular means we can't in general do fast discover on interactions for `bqm.adj`

PR #327, opened 2018-12-19
WIP: Speed up the appveyor build by only building cpp extensions when they change

arcondello:


arcondello:
I have no idea if this works because of how appveyor does caching....

PR #330, opened 2018-12-19
Add HigherOrderComposite

### LINKED ISSUE #270: extracting whether constraints introduced by make_quadratic are satisfied
under dimod.higher_order, we can add a function that takes the bqm object and the response and checks whether the constraints are satisfied for the samples. Can add that as a field to the solution object.

can have a bool field to remove/keep the extra penalty information after/during this step. 

Already have the function ready to do that - let me know if a PR is wanted.

conta877:
Need to write docs and tests.
takes a HUBO or Hising and solves it upon penalty reductions.

right now its assuming there is a single way of reducing a HUBO to a Hising. in reality there may be many ways and we should think about treating this like chain breaks resolutions... there can be a function that defines what kind of reduction user wants.

Closes #270


dimod/higherorder.py
```diff
@@ -200,21 +283,68 @@ def _prod(iterable):
...
-    return sum(_prod(sample[v] for v in variables) * bias
-               for variables, bias in poly.items())
+
+    try:
+        _ = len(sample[0])
+        D = len(sample)
+        return sum(_prod_D([sample[:, v] for v in variables], D) * bias
+                   for variables, bias in poly.items())
+
+    except:
```
arcondello: This should catch a specific exception. Also it looks like you're using the `_ = len(sample[0])` as the test to decide which format to do. In that case you should just use an `if` branch.

conta877: yeah this was me avoiding numpy...

dimod/higherorder.py
```diff
@@ -200,21 +283,68 @@ def _prod(iterable):
...
 
         poly (dict):
-            Polynomial as a dict of form {term: bias, ...}, where `term` is a tuple of
-            variables and `bias` the associated bias.
+            Polynomial as a dict of form {term: bias, ...}, where `term` is a 
+            tuple of variables and `bias` the associated bias.
 
     Returns:
-        float: The energy of the sample.
+        float or list: The energy of the sample(s).
```
arcondello: `float/list` to be picked up correctly by readthedocs. You could also consider using https://github.com/dwavesystems/dimod/blob/96ef78f53b7762782491bdd635a2c2dd1d6c75d2/dimod/sampleset.py#L39 to extend to samples-like

dimod/higherorder.py
```diff
@@ -200,21 +283,68 @@ def _prod(iterable):
...
+        poly[new_tup] = v
+    for k, v in j.items():
+        new_tup = tuple(sorted((label_dict[vidx] for vidx in list(k))))
+        poly[new_tup] = v
+    return poly
+
+
+def create_poly(h, J):
+    """ given h,J creates a single polynomial dict.
+    all h's will turn into tuples.
```
arcondello: First line of a docstring should be on a single line to readthedocs to work

dimod/higherorder.py
```diff
@@ -200,21 +283,68 @@ def _prod(iterable):
...
+
+    Args:
+        h (dict): a dict of linear variables
+        J (dict): a dict of quadratic and higher order variables
+
+    Returns
+        dict: a higher order problem dict containing both linear,
+    quadratic and n-order terms.
```
arcondello: needs to be indented for sphinx to pick it up

dimod/higherorder.py
```diff
@@ -23,8 +23,92 @@
...
+            data[name] = record[name]
+
+    data['penalty_satisfaction'] = penalty_vector
+    response.info['reduction'] = bqm.info['reduction']
+    response.info['penalty_stength'] = penalty_strength
+    return SampleSet(data, original_variables, response.info,
+                     response.vartype)
+
+
+def check_penalty_satisfaction(response, bqm):
```
arcondello: I think this is a misleading function name because it returns a vector. I would expect check to return a bool or throw an exception.

dimod/higherorder.py
```diff
@@ -23,8 +23,92 @@
...
+        response = self.child.sample(bqm, **parameters)
+
+        return polymorph_response(response, h, J, offset, bqm,
+                                  penalty_strength=penalty_strength)
+
+
+def polymorph_response(response, h, J, offset, bqm, penalty_strength=None):
+    record = response.record
+    original_variables = set(
+        sorted(set(h.keys()) | set(v for key in J.keys() for
```
arcondello: In python3 you can't sort unlike types so this will fail for `h = {'a': -1, 0: 1}`

dimod/higherorder.py
```diff
@@ -23,8 +23,92 @@
...
+    def parameters(self):
+        param = self.child.parameters.copy()
+        param['penalty_strength'] = []
+        return param
+
+    @property
+    def properties(self):
+        return {'child_properties': self.child.properties.copy()}
+
+    def sample(self, h, J, offset=0, penalty_strength=1.0, **parameters):
```
arcondello: this should be `sample_ising` to use the`ComposedSampler` ABC.

I do like that it's a composite. If it is used nested in a stack than in some sense it just becomes identity.

dimod/higherorder.py
```diff
@@ -23,8 +23,92 @@
 from dimod.binary_quadratic_model import BinaryQuadraticModel
 from dimod.decorators import vartype_argument
 from dimod.vartypes import Vartype
+import dimod
```
arcondello: Things should be imported by name into the namespace rather than using `dimod.`. Makes the import behaviour more consistent.

dimod/higherorder.py
```diff
@@ -23,8 +23,92 @@
...
+from dimod.response import SampleSet
+import numpy as np
 
-__all__ = ['make_quadratic']
+__all__ = ['make_quadratic','HigherOrderComposite']
+
+
+class HigherOrderComposite(dimod.ComposedSampler):
```
arcondello: Should consider moving this under `dimod/reference/composites`, leaving the utility functions here.

arcondello:
Looks good so far! Most of my comments are just nitpicks. Thanks @conta877 !

dimod/higherorder.py
conta877: so I kinda realized that we end up cheating here...
the main function name is now sample_ising but under the hood we are calling sample for the children.

do all composites have self.sample for sure?

dimod/reference/composites/higherordercomposite.py
```diff
@@ -0,0 +1,231 @@
...
+from dimod.core.composite import ComposedSampler
+from dimod.higherorder import make_quadratic, poly_energy
+from dimod.response import SampleSet
+
+__all__ = ['HigherOrderComposite']
+
+
+class HigherOrderComposite(ComposedSampler):
+    """Composite to reduce variables of a higher order problem.
+   Inherits from :class:`dimod.ComposedSampler`.
```
arcondello: You probably got this line from somewhere else but it's not necessary

dimod/reference/composites/higherordercomposite.py
```diff
@@ -0,0 +1,231 @@
...
+    def sample_ising(self, h, J, offset=0, penalty_strength=1.0,
+                     keep_penalty_variables=False,
+                     discard_unsatisfied=False, **parameters):
+        """
+
+        Args:
+            h (dict): linear biases corresponding to the HUBO form
+            J (dict): higher order biases corresponding to the HUBO form
+            offset (float, optional): constant energy offset
+            penalty_strength (float, optional): default is None, if provided,
```
arcondello: Not clear what this does from the discription

dimod/reference/composites/higherordercomposite.py
```diff
@@ -0,0 +1,231 @@
...
+def penalty_satisfaction(response, bqm):
+    """ Given a sampleSet and a bqm object, will calculate the
+
+    Args:
+        response (:obj:`.SampleSet`): Samples corresponding to provided bqm
+        bqm (:class:`.BinaryQuadraticModel`): a bqm object that contains
+            its reduction info.
+
+    Returns:
+        (np.array): a binary array of penalty satisfaction information.
```
arcondello:
```
:obj:`numpy.ndarray`
```

dimod/reference/composites/higherordercomposite.py
```diff
@@ -0,0 +1,231 @@
...
+                                  keep_penalty_variables=keep_penalty_variables,
+                                  discard_unsatisfied=discard_unsatisfied)
+
+
+def penalty_satisfaction(response, bqm):
+    """ Given a sampleSet and a bqm object, will calculate the
+
+    Args:
+        response (:obj:`.SampleSet`): Samples corresponding to provided bqm
+        bqm (:class:`.BinaryQuadraticModel`): a bqm object that contains
```
arcondello: should be `obj` not `class`

dimod/reference/composites/higherordercomposite.py
```diff
@@ -0,0 +1,231 @@
...
+    return poly
+
+
+def polymorph_response(response, h, J, bqm, offset=0,
+                       penalty_strength=None,
+                       keep_penalty_variables=True,
+                       discard_unsatisfied=False):
+    """ Given a response of a penalized hubo, will convert recalculate take
+    care of penalty information
+    Args:
```
arcondello: Space here betore Args

dimod/higherorder.py
conta877: I don't know what you are suggesting here

arcondello: In dimod we have a notion of `sample-like`, which can be a lot of things. If we're going to extend from dict to also allow lists/numpy array then we should use the full sample-like notion. See [as_samples](https://docs.ocean.dwavesys.com/projects/dimod/en/latest/reference/generated/dimod.sampleset.as_samples.html#dimod.sampleset.as_samples). So any 'sample-like' object would be a valid input as samples. This would also allow us to do `poly_energies`.

arcondello: I can also change this later

dimod/higherorder.py
```diff
@@ -103,8 +104,9 @@ def make_quadratic(poly, strength, vartype=None, bqm=None):
 
     Examples:
 
+        >>> import dimod
```
arcondello: We're moving away from importing for the main package in examples. If you were importing from numpy say you would have an import, but `import dimod` is kind of redundant.

dimod/higherorder.py
```diff
@@ -193,28 +198,97 @@ def _reduce_degree(bqm, poly, vartype, scale):
     return _reduce_degree(bqm, new_poly, vartype, scale)
 
 
+def create_poly(linear, quadratic):
+    """ given h,j creates a single polynomial dict.
```
arcondello: See above about docstrings.

```
"""Creates a polynomial dict.

... other stuff
"""
```

dimod/higherorder.py
```diff
@@ -193,28 +198,97 @@ def _reduce_degree(bqm, poly, vartype, scale):
     return _reduce_degree(bqm, new_poly, vartype, scale)
 
 
+def create_poly(linear, quadratic):
```
arcondello: `quadratic` should probably not be renamed if it can contain higher order terms

dimod/higherorder.py
```diff
@@ -193,28 +198,97 @@ def _reduce_degree(bqm, poly, vartype, scale):
...
+    val = [1.] * dim
+    for v in iterable:
+        val *= v
+    return val
+
+
 def poly_energy(sample, poly):
-    """Create a binary quadratic model from a higher order polynomial.
+    """
+    Calculate energy of a sample that is a solution to a higher order problem
```
arcondello: See above about docstrings

```
"""Calculates energy of a sample from a higher order polynomial.

... other stuff
"""
```
or something like that

dimod/higherorder.py
```diff
@@ -193,28 +198,97 @@ def _reduce_degree(bqm, poly, vartype, scale):
...
+    provided by poly.
 
     Args:
-        sample (dict):
-            Sample for which to calculate the energy, formatted as a dict where keys
-            are variables and values are the value associated with each variable.
+        sample (dict or (list,np.array)):
```
arcondello:
```
(dict/list/:obj:`numpy.ndarray`)
```

dimod/higherorder.py
```diff
@@ -193,28 +198,97 @@ def _reduce_degree(bqm, poly, vartype, scale):
...
 
         poly (dict):
-            Polynomial as a dict of form {term: bias, ...}, where `term` is a tuple of
-            variables and `bias` the associated bias.
+            Polynomial as a dict of form {term: bias, ...}, where `term` is a
+            tuple of variables and `bias` the associated bias.
 
     Returns:
-        float: The energy of the sample.
+        float/list: The energy of the sample(s).
```
arcondello: This should be consistent and always return a float.

```
def poly_energy(sample, poly):
    try:
        energy, = poly_energies(sample, dict)
    except ValueError:
        raise ValueError('poly_energy accepts a single sample. For multiple '
                         'samples use poly_energies')
    return energy
```
or something like that would work

dimod/higherorder.py
```diff
@@ -193,28 +198,97 @@ def _reduce_degree(bqm, poly, vartype, scale):
...
+            where keys are variables and values are the value associated with 
+            each variable. If formatted as a list or np.array, the terms in 
+            poly dict must be provided as variable order.
+
+        poly (dict):
+            Polynomial as a dict of form {term: bias, ...}, where `term` is a 
+            tuple of variables and `bias` the associated bias.
+
+    Returns:
+        float/list: The energy of the sample(s).
```
arcondello: Should always return the same type, a list or a numpy array

dimod/higherorder.py
```diff
@@ -193,28 +198,97 @@ def _reduce_degree(bqm, poly, vartype, scale):
...
+    if not isinstance(sample[0], Number):
+        raise ValueError('poly_energy accepts a single sample. For multiple '
+                         'samples use poly_energies')
+    else:
+        return poly_energies(sample, poly)
+
+
+def poly_energies(samples_like, poly):
+    """
+    Calculate energy of sample(s) that are solutions to a higher order problem
```
arcondello: See above comments about docstrings

dimod/higherorder.py
```diff
@@ -193,28 +198,97 @@ def _reduce_degree(bqm, poly, vartype, scale):
...
+    provided by poly.
+
+    Args:
+        sample (samples_like):
+            Sample(s) for which to calculate the energy, formatted as a dict
```
arcondello: Can probably just link to where samples-like is described elsewhere. Maybe something like
```
            samples (samples_like):
                A collection of raw samples. `samples_like` is an extension of NumPy's array_like
                structure. See :func:`.as_samples`.
```

dimod/reference/composites/higherordercomposite.py
```diff
@@ -0,0 +1,249 @@
...
+from dimod.core.composite import ComposedSampler
+from dimod.higherorder import make_quadratic, poly_energies
+from dimod.response import SampleSet
+
+__all__ = ['HigherOrderComposite']
+
+
+class HigherOrderComposite(ComposedSampler):
+    """Reduces a HUBO to bqm by introducing penalties. Energies of the returned
+   samples do not include the penalties.
```
arcondello: The second sentence should go on a later line.

```
"""Reduces a HUBO to bqm by introducing penalties.

Energies of the returned samples do not include the penalties.

...
```

dimod/reference/composites/higherordercomposite.py
```diff
@@ -0,0 +1,249 @@
...
+        return param
+
+    @property
+    def properties(self):
+        return {'child_properties': self.child.properties.copy()}
+
+    def sample_ising(self, h, J, offset=0, penalty_strength=1.0,
+                     keep_penalty_variables=False,
+                     discard_unsatisfied=False, **parameters):
+        """ Takes in linear variables in h and quadratic and higher order
```
arcondello: See above about docstrings

dimod/reference/composites/higherordercomposite.py
```diff
@@ -0,0 +1,249 @@
...
+
+        return polymorph_response(response, h, J, bqm,
+                                  offset=offset,
+                                  penalty_strength=penalty_strength,
+                                  keep_penalty_variables=keep_penalty_variables,
+                                  discard_unsatisfied=discard_unsatisfied)
+
+
+def penalty_satisfaction(response, bqm):
+    """ Given a sampleSet and a bqm object, will calculate the
```
arcondello: well calculate the?

dimod/higherorder.py
```diff
@@ -133,7 +135,10 @@ def make_quadratic(poly, strength, vartype=None, bqm=None):
 
 
 def _reduce_degree(bqm, poly, vartype, scale):
+    """
+    helper function of make_quadratic
 
+    """
```
arcondello: Kind of redundant I think. Could be one line `"""Helper function for make_quadratic"""`.

google styleguide on docstring:
https://github.com/google/styleguide/blob/gh-pages/pyguide.md#383-functions-and-methods
another example:
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

dimod/higherorder.py
```diff
@@ -193,28 +195,92 @@ def _reduce_degree(bqm, poly, vartype, scale):
...
-    return sum(_prod(sample[v] for v in variables) * bias
-               for variables, bias in poly.items())
+    if isinstance(sample, dict):
+        return poly_energies(sample, poly)
+
+    if not isinstance(sample[0], Number):
+        raise ValueError('poly_energy accepts a single sample. For multiple '
+                         'samples use poly_energies')
+    else:
+        return poly_energies(sample, poly)
```
arcondello: This should always return a float
```
try:
    energy, = poly_energies(sample, poly)
except ValueError:
    raise ValueError("more than one sample or something like that here")
return energy
```

dimod/higherorder.py
```diff
@@ -193,28 +194,89 @@ def _reduce_degree(bqm, poly, vartype, scale):
...
+    Returns:
+        list/:obj:`numpy.ndarray`: The energy of the sample(s).
+
+    """
+    sample, labels = as_samples(samples_like)
+    idx, label = zip(*enumerate(labels))
+    labeldict = dict(zip(label, idx))
+
+    num_samples = len(sample)
+    return sum(_prod_d([sample[:, labeldict[v]] for v in variables],
```
arcondello: Wouldn't this return a single value?

arcondello: I guess not seeing as the tests pass... Good on numpy

Issue #332, opened 2018-12-20
Export SampleSet object to Pandas df


arcondello:
**Application**
It would be useful for Pandas users to be able to export the `SampleSet`/`Response` objects into a pandas dataframe.

**Proposed Solution**
```
samples = dimod.ExactSolver().sample_ising(h, J)
df = samples.to_pandas_dataframe()
```

PR #333, opened 2018-12-20
WIP: markov-network

arcondello:
Before I finished off documentation and tests I wanted to make sure that this meets the requirements.

I have provided the following example to see how it lines up with https://github.com/dwavesystems/markov-network#example
```
            >>> fixed = {'n01': 1, 'n03': 0. }
            >>> edges = [['n01', 'n02', 3.0, 4.0, -6.0, 2.0], 
                         ['n02', 'n04', 5.0, -1.0, -4.0, 2.0], 
                         ['n03', 'n04', 1.0, 6.0, -6.0, 2.0], 
                         ['n03', 'n05', -4.0, -3.0, 6.0, 5.0]]
            ...
            >>> # generate the markov network in a form that can be solved
            >>> network = dimod.BinaryQuadraticModel.from_markov_random_field(edges)
            ...
            >>> # next we want to clamp/fix some of the variables
            >>> clamped = network.copy()
            >>> clamped.fix_variables(fixed)
            ...
            >>> # sample
            >>> samples = dimod.ExactSolver().sample(clamped)
            >>> samples.first
            Sample(sample={'n02': 0, 'n04': 1, 'n05': 0}, energy=-1.0, num_occurrences=1)
```

One piece that is missing is the export to pandas dataframe. See #332


dimod/binary_quadratic_model.py
```diff
@@ -2048,6 +2048,43 @@ def from_qubo(cls, Q, offset=0.0):
...
+            >>> network = dimod.BinaryQuadraticModel.from_markov_random_field(edges)
+            ...
+            >>> # next we want to clamp/fix some of the variables
+            >>> clamped = network.copy()
+            >>> clamped.fix_variables(fixed)
+            ...
+            >>> # sample
+            >>> samples = dimod.ExactSolver().sample(clamped)
+            >>> samples.first
+            Sample(sample={'n02': 0, 'n04': 1, 'n05': 0}, energy=-1.0, num_occurrences=1)
```
jberwald: We'll have to include the additional step where the fixed variables are added back in so that the "full problem" is visible to the user on output. That might be best accomplished in a custom `to_dataframe` step.

tests/test_binary_quadratic_model.py
```diff
@@ -1947,3 +1947,13 @@ def test_bson_sparse(self):
         dec_bqm = dimod.BinaryQuadraticModel.from_bson(enc_bqm)
 
         self.assertEqual(bqm, dec_bqm)
+
+class TestMarkovRandomField(unittest.TestCase):
+    def test_energy_single_interaction(self):
+        edges = [(0, 1, .5, -.1, 3, 0.7)]
+        bqm = dimod.BinaryQuadraticModel.from_markov_random_field(edges)
+
+        self.assertAlmostEqual(bqm.energy({0: 1, 1: 1}), edges[0][2])
```
jberwald: Why 'almost'? Just floating point differences?

dimod/binary_quadratic_model.py
```diff
@@ -2048,6 +2048,43 @@ def from_qubo(cls, Q, offset=0.0):
...
+            >>> edges = [['n01', 'n02', 3.0, 4.0, -6.0, 2.0], 
+                         ['n02', 'n04', 5.0, -1.0, -4.0, 2.0], 
+                         ['n03', 'n04', 1.0, 6.0, -6.0, 2.0], 
+                         ['n03', 'n05', -4.0, -3.0, 6.0, 5.0]]
+            ...
+            >>> # generate the markov network in a form that can be solved
+            >>> network = dimod.BinaryQuadraticModel.from_markov_random_field(edges)
+            ...
+            >>> # next we want to clamp/fix some of the variables
+            >>> clamped = network.copy()
```
jberwald: Why `copy`?

tests/test_binary_quadratic_model.py
arcondello: yes

dimod/binary_quadratic_model.py
arcondello: `fix_variables` acts on the BQM in-place, so we need to copy it first if we want to keep the original bqm (in this case it doesn't matter)

dimod/binary_quadratic_model.py
arcondello: Yeah, I just added a simple `to_dataframe` method (see #334 ). For now it would be simple to append the df, but in the future we want to add a
```
samples = FixVariablesComposite(SimulatedAnnealingSampler()).sample(bqm)
```
which adds them back in to the record.

tests/test_binary_quadratic_model.py
arcondello: it might actually work, just a habit for energy calculations :slightly_smiling_face:

dimod/binary_quadratic_model.py
jberwald: You'd need to retain the original (non-fixed) `bqm` to do this, correct?  

Would this functionality be wrapped in a simple function? It's getting a bit convoluted for your average user off the street. :)

arcondello: The composite above would handle that for you (I edited it to make it a bit more clear). The idea it that the ising problem send to `SimulatedAnnealingSampler` would (in that example) have a/b clamped, but the returned `samples` would have those values reinserted

arcondello: I have expanded these thoughts in #335

arcondello:
Thanks for the comments! The first point is fundamentally philosophical so I will address the technical ones in this comment.
* Because the phi and quadratic-edge formats are not part of a known standard, I think that should be part of a wrapper provided to EigenMed. I would be happy to write it.
* I am completely happy to have this block release, the good news is that it is being written today and should be up by the end of the week.
* As was mentioned above, see #334
* It does, I provided an example to Jesse before that had a bug but the above is correct.

murraythom:
Thanks for your request for a Markov Network object. Regarding MNs vs BQMs, I agree we want to improve transparency for developers who work regularly with MNs. I think we'll look to see if some of that can be done with common MN programming tools. If there are any developers who work regularly with Markov Networks who can point us to any existing open source projects that would be helpful!

spreinhardt:
Excellent, thanks for taking these requirements into account.  I’ll look for evidence on the first point; it’s not just philosophical.

On 12/20/18, 2:50 PM, "Alexander Condello" <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks for the comments! The first point is fundamentally philosophical so I will address the technical ones in this comment.
·         Because the phi and quadratic-edge formats are not part of a known standard, I think that should be part of a wrapper provided to EigenMed. I would be happy to write it.
·         I am completely happy to have this block release, the good news is that it is being written today and should be up by the end of the week.
·         As was mentioned above, see #334<https://github.com/dwavesystems/dimod/pull/334>
·         It does, I provided an example to Jesse before that had a bug but the above is correct.

jberwald:
It's possible that "transparency" is the wrong term for what we're looking to provide for the user. I interpret Steve's request not as "philosophy" but as a desire to provide the user with the best OO standard: 
1. "dealing with BQMs is an unnecessary conceptual overload": This is not philosophy, I see it is a request for *encapsulation* and *abstraction*. Let's only define what is needed for the situation. In particular, this principle of OO programming asks that the developer only include aspects that are relevant to the particular situation. Are BQM's relevant? Is an `ExactSolver` something the user *needs* to understand? Or could we encapsulate these in some syntactic sugar? Speaking of...
2. "slight bit of syntactic sugar": A spoonful of polymorphism is all that we need here. For instance, let `MarkovNetwork` inherit from `NetworkX` (for graph stuff) and `dimod` (for sampling work) and you can then overload/redefine the `sample` method in whatever way is necessary for a MarkovNetwork. For example, `MarkovNetwork(param1, param2, token=mytoken).sample([list of edges], {dict of forced vars})` would call the QPU and return a DF. This is simple, elegant, and the user never needs to know that a BQM or QUBO was created. 
3 "often struck, in a negative way, by how many concepts/terms I'm expected to grok to do almost anything": A clear signal/sanity check that abstraction and encapsulation may not be working the way people think. 

IMHO, creating `BinaryQuadraticModel.from_markov_random_field()` completely misses the power of encapsulation and polymorphism. We want to empower a developer to stay in their space by enabling them to extend our packages with common patterns. We do not want to force them to come into our space and have to figure out how to manipulate a `BinaryQuadraticModel` and then fiddle around with a `Response` object.  Choosing a "common" open source tool isn't much better, and sounds like a very heavy-weight solution. 

Alright, that's my $0.02. As usual, happy to answer any questions. :)

arcondello:
Appreciate the $0.02! I want to make sure that we make the best code possible.

For a user who desires to solve their markov network problem on a QPU, understanding BQMs is relevant. Since concepts like sampling and fixing variables are properties of the BQM, not concepts of the markov network, it makes sense to include them in a binary quadratic model object. The point here is not to pretend that a BQM is a markov network, the method `from_markov_random_field` *derives* a BQM from an input that defines a markov network. For your specific example `MarkovNetwork(param1, param2, token=mytoken).sample([list of edges], {dict of forced vars})` includes two notions, 'token' and 'sample' that are QPU and BQM properties respectively, not concepts associated with markov networks.

I do think that it would make sense to define what that input should be. I have followed the syntax from the experimental code that defines a markov network in term of a list-of-lists, but I would happily take suggestions of better input formats. I do not think though that creating a `MarkovNetwork` object is something that fits within Ocean where we try whenever possible to use existing formats/frameworks/libraries. My preference would be to use an existing format/library but failing that a 'native' python object that encodes the relevant information in a simple way is best.

This is also just the first pull request dealing with these concepts. If you know of an existing library or framework that is commonly used in the markov network community that we could extend using this logic then please let me know!

jberwald:
@arcondello, In this case the `sample` method happens to be "BQM-derived". My point was that in developing extensions it is important to have a common framework, so I used the term `sample` to indicate that I want to transform my QUBO into QMI. The best-in-class tools for this allow for the developer to subclass a base class with such a method, and all that is expected as data moves through a pipeline is that each class has such a common method.  Since `sample` seems ubiquitous in dwave code, it's the clear option. Libraries like `sklearn` use `fit`, `transform`, and `fit_transform` so that large pipelines can fit seamlessly together. The closest analogy for us would likely be `sample` == `fit_transform`. 

I entirely agree that it is not ideal to create a `MarkovNetwork` object within Ocean (or dimod for that matter). That's why we initially submitted the code as a separate (extension) package (with dimod and Ocean dependencies). I like many of the changes/simplifications you've proposed, but I'm still not clear on why keeping it a a separate package is so undesirable.

jberwald:
I think it's necessary to address the notion that a user must understand BQM's to use a D-Wave package like MarkovNetwork, and I'm worried I'm missing something basic or fundamental. Ultimately, I really want to create the simplest interface for a user based on canonical Python data structures. The whole point of creating a wrapper class is so that a Markov *Network* user can think only in graphical terms. (I consider NetworkX graphs canonical given their ubiquity in scientific code.)

Here is an example from another software land: In Scala Spark, a `CountVectorizer` is used to count the number of times a word occurs in a document. A user can send call `CountVectorizer.fit()` on  a large Dataset with millions of rows containing text (`Dataset` ~= strongly typed list in this example). The user needs to know nothing of the underlying data structures that are being accessed, and on the other side they get out  another Dataset containing the word counts they are looking for. Underneath the hood, an entire optimized SQL engine is being accessed. But the user never needs to see that in most cases.

I view the task of creating a MarkovNetwork interface similarly. The user should be able to send in a graph plus a list of forced nodes and receive a list of node values in return. It’s easy to dive in to view the BQM’s if the user wants, and maybe some parameter settings will require knowledge of the underlying data structures, which is only natural.  But in the end, the default call should look something like this: `node_val_dict = MarkovNetwork().sample(graph, forced)`. 

@arcondell, if you think this is not possible then there must something fundamental about how dimod/Ocean operates that I do not understand.

arcondello:
Closed by https://github.com/dwavesystems/dwave_networkx/pull/92

Issue #268, opened 2018-08-28
Aggregate unique samples (num_occurrences++) on Response construction


randomir:
**Application**
Right now `Response.from_samples` uses samples verbatim, even if they are identical (same values, same energies) -- although each `Sample` constructed has `num_occurrences` field.

**Proposed Solution**
It would be useful to provide an option to combine identical samples, only summing up the number of occurrences.

**Alternatives Considered**
Aggregate/unique-ify in user code, prior `Response` construction. Implementation on `dimod` side will probably be more efficient and overall convenient for users.

arcondello:
I remember the counter-case, sometimes the order of samples matters, and aggregating them would eliminate that. I think an intermediate solution would be to add a `aggregate_samples()` method that can be called by the user or the sampler author, but in general we would not always keep them aggregated.

arcondello:
Yet another approach would be to aggregate automatically but for the samplers for which order matters, add the order into the `.info` field.

So `[[0, 0], [1, 1], [1, 0], [0, 0]]` would become `[[0, 0], [1, 1], [1, 0]]` with `.info['order'] = [0, 1, 2, 0]`

PR #334, opened 2018-12-20
Add SampleSet.to_pandas_dataframe method

### LINKED ISSUE #332: Export SampleSet object to Pandas df

arcondello:
Closes #332

### DIFF
--- a/tests/test_sampleset.py
+++ b/tests/test_sampleset.py
@@ -21,6 +21,13 @@ import numpy as np
 import dimod
 
 
+try:
+    import pandas as pd
+    _pandas = True
+except ImportError:
+    _pandas = False
+
+
 class TestSampleSet(unittest.TestCase):
     def test_from_samples(self):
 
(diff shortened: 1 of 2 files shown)


arcondello:
Closes #334

tests/test_sampleset.py
```diff
@@ -158,3 +165,16 @@ def test_functional_str(self):
...
+class TestSampleSet_to_pandas_dataframe(unittest.TestCase):
+    def test_simple(self):
+        samples = dimod.SampleSet.from_samples(([[-1, 1, -1], [-1, -1, 1]], 'abc'),
+                                               dimod.SPIN, energy=[-.5, .5])
+        df = samples.to_pandas_dataframe()
+
+        other = pd.DataFrame([[-1, 1, -1, -.5, 1], [-1, -1, 1, .5, 1]],
+                             columns=['a', 'b', 'c', 'energy', 'num_occurrences'])
+
+        pd.testing.assert_frame_equal(df, other, check_dtype=False)
```
jberwald: Please consider adding a test for duplicate samples. When these occur they will produce duplicate rows in the DF. For example if `sample1 = {'a':1, 'b':0}` is sampled  twice by QC it should show up in the sample set with  `num_occurrences =2`. I find that this is not always true (no clue why), and that I sometimes get two separate instances of `sample1` in a returned Response object. I have resorted to a groupby in pandas to account for this and get the correct `num_occurrences`.

arcondello: I think the best approach would be to add an aggregator to the `SampleSet` object that can be invoked before `to_pandas_dataframe`. See #268

PR #336, opened 2018-12-20
Add smear_vartype to embed_bqm function.

### LINKED ISSUE #204: dimod.embed_bqm: Couplings/biases out of j/h range after embedding

### LINKED ISSUE #241: Embedding QUBO leads to out-of-range linear terms.

arcondello:
Closes #241
Closes #204

### DIFF
--- a/dimod/embedding/transforms.py
+++ b/dimod/embedding/transforms.py
@@ -35,7 +35,8 @@ __all__ = ['embed_bqm',
            ]
 
 
-def embed_bqm(source_bqm, embedding, target_adjacency, chain_strength=1.0):
+def embed_bqm(source_bqm, embedding, target_adjacency, chain_strength=1.0,
+              smear_vartype=None):
     """Embed a binary quadratic model onto a target graph.
 
     Args:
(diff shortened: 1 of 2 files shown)


dimod/embedding/transforms.py
```diff
@@ -54,6 +55,10 @@ def embed_bqm(source_bqm, embedding, target_adjacency, chain_strength=1.0):
             Magnitude of the quadratic bias (in SPIN-space) applied between variables to create chains. Note
             that the energy penalty of chain breaks is 2 * `chain_strength`.
 
+        smear_vartype (:class:`.Vartype`, optional, default=None):
+            Determine whether the bqm is embedded in SPIN or BINARY space. By default the embedding
+            is done according to the given source_bqm.
```
randomir: If I understand this correctly, the `smear_vartype` is actually something like `target_bqm_vartype`, or `result_vartype`, or `embedded_bqm_vartype`, or something along those lines? I find `smear_vartype` to be very confusing and unclear..

arcondello: I added some more explanation but it would be hard to fully explain without giving an example and that would take up a lot of docstring space...

Issue #335, opened 2018-12-20
FixedVariablesComposite


arcondello:
**Application**
Sometimes we know in advance some of the values in the solution (see [fix_variable](https://docs.ocean.dwavesys.com/projects/dimod/en/latest/reference/generated/dimod.BinaryQuadraticModel.fix_variable.html#dimod.BinaryQuadraticModel.fix_variable)). We could wrap this functionality in a composite to make 'nicer' `SampleSet` objects.

**Proposed Solution**
```
h = {'a': -1, 'b': 0, 'c': 0}
J = {('a', 'b'): -1, ('b', 'c'): -1}
samples = FixedVariablesComposite(ExactSolver()).sample_ising(h, J, fixed={'a': -1, 'b': +1})

assert 'a' in samples.variables
assert 'b' in samples.variables
```

@conta877

PR #338, opened 2018-12-21
Add FixedVariableComposite

conta877:
docs and tests are missing

### DIFF
--- a/dimod/reference/composites/fixedvariable.py
+++ b/dimod/reference/composites/fixedvariable.py
@@ -5,6 +5,7 @@ import numpy as np
 
 __all__ = ['FixedVariableComposite']
 
+
 class FixedVariableComposite(dimod.ComposedSampler):
 
     def __init__(self, child_sampler):
@@ -31,20 +32,19 @@ class FixedVariableComposite(dimod.ComposedSampler):
         child = self.child
         if fixed_variables is None:
             fixed_variables = fix_variables(source_bqm)
+
         bqm = source_bqm.copy()
+        bqm.fix_variables(fixed_variables)
 
-        for v, val in fixed_variables.items():
-            bqm.fix_variable(v, val)
-
-        if len(bqm.variables)==0:
-            return _fixed_response(fixed_variables,bqm.offset,bqm.vartype)
+        if len(bqm.variables) == 0:
+            return _fixed_response(fixed_variables, bqm.offset, bqm.vartype)
 
         response = child.sample(bqm, **parameters)
-        
+
         return _release_response(response, bqm, fixed_variables)
 
 
-def _fixed_response(fixed_variables,offset,vartype):
+def _fixed_response(fixed_variables, offset, vartype):
     variables = fixed_variables.keys()
 
     samples = np.asarray([list(fixed_variables.values())])
@@ -53,7 +53,8 @@ def _fixed_response(fixed_variables,offset,vartype):
     num_variables = len(fixed_variables)
 
     datatypes = [('sample', np.dtype(np.int8), (num_variables,)),
-                 ('energy', energy.dtype),('num_occurrences',np.dtype(np.int8))]
+                 ('energy', energy.dtype),
+                 ('num_occurrences', np.dtype(np.int8))]
 
     data = np.rec.array(np.empty(1, dtype=datatypes))
 
@@ -61,13 +62,12 @@ def _fixed_response(fixed_variables,offset,vartype):
     data.energy = energy
     data.num_occurrences = 1
 
-    return SampleSet(data, variables, 
-                     {'fixed_variables':fixed_variables},
+    return SampleSet(data, variables,
+                     {'fixed_variables': fixed_variables},
                      vartype)
 
 
 def _release_response(response, bqm, fixed_variables):
-
     original_variables = list(bqm.variables)
     record = response.record
     samples = np.asarray(record.sample)
--- a/dimod/reference/composites/fixedvariable.py
+++ b/dimod/reference/composites/fixedvariable.py
@@ -16,16 +16,19 @@
 """
 A composite that fixes the variables provided and removes them from the
 bqm object before sending to its child sampler.
+
+See `Ocean Glossary <https://docs.ocean.dwavesys.com/en/latest/glossary.html>`_ for explanations
+of technical terms in descriptions of Ocean tools.
 """
 import numpy as np
 from dimod.response import SampleSet
 
-import dimod
+from dimod import ComposedSampler
 
 __all__ = ['FixedVariableComposite']
 
 
-class FixedVariableComposite(dimod.ComposedSampler):
+class FixedVariableComposite(ComposedSampler):
     """Composite to fix variables of a problem to provided, assigned values
 
     Inherits from :class:`dimod.ComposedSampler`.
@@ -34,7 +37,7 @@ class FixedVariableComposite(dimod.ComposedSampler):
     accordingly. Returned samples include the fixed variable
 
     Args:
-       sampler (:class:`dimod.Sampler`):
+       sampler (:obj:`dimod.Sampler`):
             A dimod sampler
 
     Examples:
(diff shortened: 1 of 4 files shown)


dimod/reference/composites/fixedvariable.py
```diff
@@ -0,0 +1,103 @@
...
+    def properties(self):
+
+        return {'child_properties': self.child.properties.copy()}
+
+    def sample(self, source_bqm, fixed_variables=None,
+               **parameters):
+
+        child = self.child
+        if fixed_variables is None:
+            fixed_variables = fix_variables(source_bqm)
```
arcondello: Not sure we want to do this automatically. Maybe toggled by a `use_roof_duality` flag?

dimod/reference/composites/fixedvariable.py
```diff
@@ -0,0 +1,103 @@
...
+        bqm = source_bqm.copy()
+
+        for v, val in fixed_variables.items():
```
arcondello: We have a new method `fix_variables` that you can use.

dimod/reference/composites/fixedvariable.py
```diff
@@ -0,0 +1,103 @@
...
+            bqm.fix_variable(v, val)
+
+        if len(bqm.variables)==0:
+            return _fixed_response(fixed_variables,bqm.offset,bqm.vartype)
```
arcondello: couldn't you just generate an empty response? Then pass it through `_release_response`?

dimod/reference/composites/fixedvariable.py
conta877: ok i'll change this

dimod/reference/composites/fixedvariable.py
conta877: I assume if they are calling fixedvariablecomposite without the fixed_variables they want us to decide

dimod/reference/composites/fixedvariable.py
conta877: i feel like this is more efficient... but may not be.

arcondello: It is, but for a case that's silly. So I don't feel like we have a responsibility to optimize it....

dimod/reference/composites/fixedvariable.py
arcondello: Ok, fair point. Though the case could be made that we would want to run that even if they do provide us with some variables to fix.

conta877: this way we allow both options... if someone is experimenting its a nice feature to be able to define your own fixed variables without any decision making on our end.

dimod/reference/composites/fixedvariable.py
conta877: its written already?

dimod/reference/composites/fixedvariable.py
```diff
@@ -0,0 +1,103 @@
...
+                                  'energy'})
+
+    data = np.rec.array(np.empty(num_samples, dtype=datatypes))
+
+    data.sample = samples
+    data.energy = energy
+    for name in record.dtype.names:
+        if name not in {'sample', 'energy'}:
+            data[name] = record[name]
+    response.info['fixed_variables'] = fixed_variables
```
arcondello: They already provided this information, so is it necessary to give it to them again?

arcondello:
Closes #335

arcondello:
This will need to be added to the docs as well. See https://github.com/dwavesystems/dimod/blob/master/docs/reference/ref_samplers_composites.rst#L74 (you'll want to look at the 'raw' view, not the rendered one). This will probably cause the doctests to fail because the docs don't have access to `DWaveSampler`

dimod/reference/composites/fixedvariable.py
```diff
@@ -0,0 +1,235 @@
...
+#
+# ================================================================================================
+"""
+A composite that fixes the variables provided and removes them from the
+bqm object before sending to its child sampler.
+"""
+import numpy as np
+from dimod.response import SampleSet
+
+import dimod
```
arcondello: Things should be imported by name like above `from dimod.response import SampleSet`

dimod/reference/composites/fixedvariable.py
```diff
@@ -0,0 +1,235 @@
...
+class FixedVariableComposite(dimod.ComposedSampler):
+    """Composite to fix variables of a problem to provided, assigned values
+
+    Inherits from :class:`dimod.ComposedSampler`.
+
+    Fixes variables of a bqm and modifies linear and quadratic terms
+    accordingly. Returned samples include the fixed variable
+
+    Args:
+       sampler (:class:`dimod.Sampler`):
```
arcondello: This should be
```
:obj:`dimod.Sampler`
```
It might be incorrect elsewhere, but might as well fix it now while you're changing some things.

dimod/reference/composites/fixedvariable.py
```diff
@@ -0,0 +1,235 @@
...
+            A dimod sampler
+
+    Examples:
+       This example uses :class:`.FixedVariableComposite` to instantiate a
+       composed sampler that submits a simple Ising problem to a sampler.
+       The composed sampler fixes a variable and modifies linear and quadratic
+       biases according.
+
+       >>> from dwave.system.samplers import DWaveSampler
```
arcondello: This example uses a library potentially not available to the user. I think we should just show it wrapping the exact solver but maybe we should talk about that.

dimod/reference/composites/fixedvariable.py
```diff
@@ -0,0 +1,235 @@
...
+       >>> from dimod.binary_quadratic_model import BinaryQuadraticModel
+       >>> from dimod.reference.composites.fixedvariable import FixedVariableComposite
+       >>> import dimod
+       >>> sampler = FixedVariableComposite(DWaveSampler())
+       >>> h = {1: -1.1, 4: -0.5}
+       >>> J = {(1,4):-0.5}
+       >>> bqm = BinaryQuadraticModel(linear= h,quadratic = J ,offset=0,
+       >>> vartype=dimod.SPIN)
```
arcondello: I think this syntax won't work witn the extra `>>>` for the second line, but I could be wrong

dimod/reference/composites/fixedvariable.py
```diff
@@ -0,0 +1,235 @@
...
+       >>> response = sampler.sample(bqm,fixed_variables={1:-1})
+       >>> for sample in response.samples():    # doctest: +SKIP
+       ...     print(sample)
+       ...
+       fixed variables =  {1: -1}
+       {4: -1, 1: -1}
+
+    See `Ocean Glossary <https://docs.ocean.dwavesys.com/en/latest/glossary.html>`_ for explanations
+    of technical terms in descriptions of Ocean tools.
```
arcondello: If you want to include this link it should probably be at the module level rather than in every docstring.

dimod/reference/composites/fixedvariable.py
```diff
@@ -0,0 +1,235 @@
...
+
+    """
+
+    def __init__(self, child_sampler):
+        self._children = [child_sampler]
+
+    @property
+    def children(self):
+        """list: Children property inherited from :class:`dimod.Composite` class.
```
arcondello: Don't think this property needs  a docstring

dimod/reference/composites/fixedvariable.py
```diff
@@ -0,0 +1,235 @@
...
+
+        See `Ocean Glossary <https://docs.ocean.dwavesys.com/en/latest/glossary.html>`_ for explanations of technical terms in descriptions of Ocean tools.
+
+        """
+
+        return self._children
+
+    @property
+    def parameters(self):
+        """dict[str, list]: Parameters in the form of a dict.
```
arcondello: ditto this one

dimod/reference/composites/fixedvariable.py
```diff
@@ -0,0 +1,235 @@
...
+
+        See `Ocean Glossary <https://docs.ocean.dwavesys.com/en/latest/glossary.html>`_ for explanations of technical terms in descriptions of Ocean tools.
+
+        """
+        # does not add or remove any parameters
+        return self.child.parameters.copy()
+
+    @property
+    def properties(self):
+        """dict: Properties in the form of a dict.
```
arcondello: I don't think this property needs a docstring

dimod/reference/composites/fixedvariable.py
```diff
@@ -0,0 +1,235 @@
...
+                Binary quadratic model to be sampled from.
+
+            fixed_variables (dict):
+                A dictionary of variable assignments.
+
+            **parameters:
+                Parameters for the sampling method, specified by the child sampler.
+
+        Returns:
+            :class:`dimod.SampleSet`
```
arcondello:
```
:obj:`dimod.SampleSet`
```

tests/test_fixedvariablecomposite.py
```diff
@@ -0,0 +1,57 @@
...
+                               quadratic={(1, 4): -0.6},
+                               offset=0,
+                               vartype=dimod.SPIN)
+
+        fixed_variables = {1: -1}
+        sampler = FixedVariableComposite(ExactSolver())
+        response = sampler.sample(bqm, fixed_variables=fixed_variables)
+
+        self.assertDictEqual(dict(response.first.sample), {4: -1, 1: -1})
+        self.assertAlmostEquals(response.first.energy,1.2)
```
arcondello:
```

        self.asserEqual(response.first.sample, {4: -1, 1: -1})
        self.assertAlmostEquals(response.first.energy, 1.2)
```
I think that first line should work, second line has a pep8 issue

dimod/reference/composites/fixedvariable.py
```diff
@@ -0,0 +1,235 @@
...
+
+    Examples:
+       This example uses :class:`.FixedVariableComposite` to instantiate a
+       composed sampler that submits a simple Ising problem to a sampler.
+       The composed sampler fixes a variable and modifies linear and quadratic
+       biases according.
+
+       >>> from dwave.system.samplers import DWaveSampler
+       >>> from dimod.binary_quadratic_model import BinaryQuadraticModel
+       >>> from dimod.reference.composites.fixedvariable import FixedVariableComposite
```
arcondello: `from dimod import FixedVariableComposite` will work

PR #339, opened 2018-12-21
ignore offset for scale and ignore stuff options for normalize in bqm

conta877:


dimod/binary_quadratic_model.py
```diff
@@ -828,6 +829,8 @@ def scale(self, scalar, ignored_variables=None, ignored_interactions=None):
             ignored_interactions (iterable[tuple], optional):
                 As an iterable of 2-tuples. Biases associated with these interactions are not scaled.
 
+            ignore_offset (bool): if True, the offset is not scaled
```
arcondello:
```
        ignore_offset (bool, default=False):
            If True, the offset is not scaled
```

arcondello:
Just a tiny doc fix, otherwise looks good!

PR #353, opened 2019-01-04
Benchmarking

### LINKED ISSUE #351: Consider setting up a benchmarking suite
[air speed velocity](https://asv.readthedocs.io/en/stable/index.html) is used by numpy/scipy and seems promising.

**Alternatives Considered**
* Adding this much infrastructure might be overkill at this point. 
* We might want to benchmark Ocean in a more integrated way, so this might be better in the [sdk](https://github.com/dwavesystems/dwave-ocean-sdk).

arcondello:
Add benchmarks using [airspeed velocity](https://asv.readthedocs.io/).

Closes #351


arcondello:
I played around trying to add this to CI but I ran into some problems. The main one is deciding what to benchmark. Obvious candidates are comparing to last tag, comparing to master, comparing the last 10 commits. But I think we need to think about that more. For now I'll just add the benchmarks without continuous integration.

PR #370, opened 2019-01-11
SampleSet.__str__

### LINKED ISSUE #366: Nicer print for SampleSets
**Application**
The current `print(sampleset)` is hard to read. It would be good to have a nicer print form.

arcondello:
The code is not complete, it needs a refactor, but I wanted to get early feedback. Right now it looks like:
```
>>> sampleset = dimod.RandomSampler().sample_ising({v: .3 for v in range(10)}, {}, num_reads=20) 
>>> print(sampleset)
     0   1   2   3   4   5   6  ..   9               energy  num_occurrences
0   -1  +1  +1  -1  -1  -1  -1  ..  +1                 -1.2                1
1   +1  +1  -1  +1  -1  -1  -1  ..  -1                 -1.2                1
2   -1  +1  +1  +1  -1  -1  -1  ..  -1                 -1.2                1
3   -1  +1  +1  +1  -1  -1  -1  ..  -1                 -0.6                1
4   +1  +1  -1  +1  -1  -1  -1  ..  +1  -0.5999999999999999                1
5   -1  -1  -1  -1  -1  +1  -1  ..  +1  -0.5999999999999999                1
6   -1  -1  +1  +1  -1  -1  -1  ..  -1  -0.5999999999999999                1
7   +1  -1  -1  -1  -1  +1  +1  ..  +1  -0.5999999999999999                1
8   -1  -1  +1  -1  +1  +1  -1  ..  -1                  0.0                1
9   +1  -1  -1  -1  +1  +1  -1  ..  +1                  0.0                1
10  -1  +1  +1  +1  -1  -1  -1  ..  +1                  0.0                1
11  -1  +1  +1  +1  -1  +1  -1  ..  -1                  0.0                1
12  -1  +1  +1  -1  +1  +1  +1  ..  +1   0.5999999999999999                1
13  -1  -1  +1  +1  +1  +1  -1  ..  -1   0.5999999999999999                1
14  -1  +1  -1  +1  -1  +1  +1  ..  +1                  0.6                1
15  +1  +1  -1  -1  +1  -1  -1  ..  +1                  0.6                1
16  +1  -1  +1  +1  -1  -1  -1  ..  +1                  0.6                1
17  -1  -1  +1  +1  +1  +1  +1  ..  -1                  1.2                1
18  +1  -1  +1  +1  +1  +1  -1  ..  -1                  1.2                1
19  -1  +1  +1  +1  +1  +1  +1  ..  +1                  1.8                1

[ 20 rows, 10 variables ]
>>> dimod.serialization.format.SampleSetFormatter.max_rows=10
>>> dimod.serialization.format.SampleSetFormatter.max_columns=7
>>> print(sampleset)
     0   1  ..   9               energy  num_occurrences
0   -1  +1  ..  +1                 -1.2                1
1   +1  +1  ..  -1                 -1.2                1
2   -1  +1  ..  -1                 -1.2                1
3   -1  +1  ..  -1                 -0.6                1
4   +1  +1  ..  +1  -0.5999999999999999                1
..  ..  ..  ..  ..                   ..               ..
19  -1  +1  ..  +1                  1.8                1

[ 20 rows, 10 variables ]
```

Closes #366

### DIFF
--- a/dimod/serialization/format.py
+++ b/dimod/serialization/format.py
@@ -1,0 +1,143 @@
+# Copyright 2019 D-Wave Systems Inc.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+# ================================================================================================
+from itertools import islice
+
+import numpy as np
+
+import dimod
+
+
+class PassVar(object):
+    def __str__(self):
+        return '..'
+
+
+class SampleSetFormatter(object):
+
+    min_column_width = 2
+    column_space = 2
+
+    max_rows = 40
+    max_columns = 12  # 17
+    header = True
+    footer = True
+    index = True
+
+    def __init__(self, sampleset):
+
+        self.sampleset = sampleset
+
+        # sampleset's size/shape/fields won't change, so we can build the (empty)
+        # array
+
+        nrows = min(self.max_rows,
+                    (len(sampleset) +  # one row for header
+                        self.header +  # two for footer
+                        2*self.footer))
+        nrows -= 2*self.footer  # two rows for footer
+
+        ncols = min(self.max_columns,
+                    (len(sampleset.variables) +
+                        len(sampleset.record.dtype.names) - 1 +
+                        self.index))
+
+        self.char = char = np.full((nrows, ncols), '', dtype=object)
+
+        # get columns
+        self.fields = fields = [field for field in sampleset.record.dtype.names
+                                if field != 'sample']
+
+        max_variables = self.max_columns - len(fields) - self.index
+        if len(sampleset.variables) > max_variables:
+            variables = sampleset.variables[:max_variables-2]
+            variables.append(PassVar())
+            variables.append(sampleset.variables[-1])
+        else:
+            # we can do all the variables
+            variables = sampleset.variables
+        variable_labels = list(map(str, variables))
+        self.variables = variables
+
+        if self.header:
+            char[0, :] = ['']*self.index + variable_labels + fields
+
+    def to_string(self):
+        sampleset = self.sampleset
+        num_samples = len(sampleset)
+        char = self.char
+        nrows, ncols = char.shape
+
+        variables = self.variables
+        fields = self.fields
+
+        if sampleset.vartype is dimod.SPIN:
+            def _str(datum, v):
+                if isinstance(v, PassVar):
+                    return str(v)
+                return '+1' if datum.sample[v] > 0 else '-1'
+        else:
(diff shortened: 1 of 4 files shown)


dimod/serialization/format.py
```diff
@@ -0,0 +1,143 @@
...
+class SampleSetFormatter(object):
+
+    min_column_width = 2
+    column_space = 2
+
+    max_rows = 40
+    max_columns = 12  # 17
+    header = True
+    footer = True
+    index = True
```
randomir: Instead of class variables (which in this case are basically global variables), I would model the formatter/pretty printer according to [pprint](https://docs.python.org/3/library/pprint.html) module.

arcondello: For documentation we want to be able to set some formatting options once (in the docstring setup) and have them persist. I don't see how that can be done with pprint?

I was following [pandas'](https://pandas.pydata.org/pandas-docs/stable/options.html#overview) pattern. But we could use [numpy's](https://docs.scipy.org/doc/numpy/reference/generated/numpy.set_printoptions.html) instead if you think it's cleaner.

arcondello: They more-or-less amount to the same thing though.

dimod/serialization/format.py
```diff
@@ -0,0 +1,143 @@
...
+        self.sampleset = sampleset
+
+        # sampleset's size/shape/fields won't change, so we can build the (empty)
+        # array
+
+        nrows = min(self.max_rows,
+                    (len(sampleset) +  # one row for header
+                        self.header +  # two for footer
+                        2*self.footer))
+        nrows -= 2*self.footer  # two rows for footer
```
m3ller: Why do you only subtract `n_rows` with the footer and not the header?

dimod/serialization/format.py
```diff
@@ -0,0 +1,143 @@
...
+
+        ncols = min(self.max_columns,
+                    (len(sampleset.variables) +
+                        len(sampleset.record.dtype.names) - 1 +
```
m3ller: It would be nice to know what the `-1` is accounting for. Zero-indexing?

dimod/serialization/format.py
```diff
@@ -0,0 +1,143 @@
...
+            row.extend(str(getattr(datum, field)) for field in fields)
+
+            char[ci, :] = row
+            ci += 1
+
+            if ci >= nrows:
+                break
+
+        if num_samples > nrows - self.header:
+            char[-2, :] = '..'
```
m3ller: Instead of `'..'`, perhaps use `PassVar()`?

dimod/serialization/format.py
```diff
@@ -0,0 +1,143 @@
...
+        ci = 1*self.header
+        for i, datum in enumerate(sampleset.data()):
+            row = [str(i)]*self.header
+            row.extend(_str(datum, v) for v in variables)
+            row.extend(str(getattr(datum, field)) for field in fields)
+
+            char[ci, :] = row
+            ci += 1
+
+            if ci >= nrows:
```
m3ller: It is not obvious to me why you need `ci`. Why can't the `ci` in lines 103-106 be replaced with `i+1`? 

Maybe things would be clearer to me if I knew what `ci` stands for. I think it's "column index"; if so, it would be nice as a comment.

arcondello:
Updated to more closely match [numpy's](https://docs.scipy.org/doc/numpy/reference/generated/numpy.set_printoptions.html) pattern.

Issue #358, opened 2019-01-07
Scaling composite


conta877:
**Application**
Most hamiltonians for real life applications are not defined within the linear/quadratic ranges of the QPU, making scaling a neccesity. In addition, the QPU has precision limitations that needs to be taken into account that get amplified further during scaling. 

**Proposed Solution**
a scaling composite that takes care of scaling problems and dealing with precision issues. Since HigherOrderComposite is in the making, this composite ideally should be able to handle HUBOs.

conta877:
this can be closed!

PR #372, opened 2019-01-14
ScaleComposite

conta877:
a composite to scale variables/interactions/offset of a problem.

Closes https://github.com/dwavesystems/dwave-system/issues/1

### DIFF
--- a/dimod/reference/composites/scalecomposite.py
+++ b/dimod/reference/composites/scalecomposite.py
@@ -203,7 +203,10 @@ class ScaleComposite(ComposedSampler):
         # handle HUBO
         child = self.child
         if scalar is None:
-            scalar = _calc_norm_coeff(h, J, bias_range, quadratic_range)
+            scalar = _calc_norm_coeff(h, J, bias_range, quadratic_range,
+                                      ignored_variables=ignored_variables,
+                                      ignored_interactions=ignored_interactions
+                                      )
         h_sc = dict(h)
         J_sc = dict(J)
         if scalar != 1:
@@ -221,7 +224,8 @@ class ScaleComposite(ComposedSampler):
         return response
 
 
-def _calc_norm_coeff(h, J, bias_range, quadratic_range):
+def _calc_norm_coeff(h, J, bias_range, quadratic_range,ignored_variables=None,
+                               ignored_interactions=None):
     """Helper function to calculate normalization coefficient"""
 
     def parse_range(r):
@@ -242,8 +246,10 @@ def _calc_norm_coeff(h, J, bias_range, quadratic_range):
     lin_range, quad_range = map(parse_range, (linear_range,
                                               quadratic_range))
 
-    lin_min, lin_max = min_and_max(h.values())
-    quad_min, quad_max = min_and_max(J.values())
+    lin_min, lin_max = min_and_max([v for k, v in h.items()
+                                    if k not in ignored_variables])
+    quad_min, quad_max = min_and_max([v for k, v in J.items()
+                                      if k not in ignored_interactions])
 
     inv_scalar = max(lin_min / lin_range[0], lin_max / lin_range[1],
                      quad_min / quad_range[0], quad_max / quad_range[1])
--- a/dimod/reference/composites/scalecomposite.py
+++ b/dimod/reference/composites/scalecomposite.py
@@ -1,0 +1,284 @@
+# Copyright 2018 D-Wave Systems Inc.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


conta877:
implements some of the requests in #358

dimod/reference/composites/scalecomposite.py
```diff
@@ -0,0 +1,284 @@
...
+                     ignore_offset=False, **parameters):
+        """ Scale and sample from the problem provided by h, J, offset
+
+        if scalar is not given, problem is scaled based on bias and quadratic
+        ranges.
+
+        Args:
+            h (dict): linear biases corresponding to the HUBO form
+
+            J (dict): higher order biases corresponding to the HUBO form
```
arcondello: This should be less HUBO-specific. It might be worth just calling out at the end that this supports HUBOs

dimod/reference/composites/scalecomposite.py
```diff
@@ -0,0 +1,284 @@
...
+    def __init__(self, child_sampler):
+        self._children = [child_sampler]
+
+    @property
+    def children(self):
+        return self._children
+
+    @property
+    def parameters(self):
+        return self.child.parameters.copy()
```
arcondello: Need to add the parameters here, `scalar`, `bias_range`, etc

dimod/reference/composites/scalecomposite.py
```diff
@@ -0,0 +1,284 @@
...
+        bqm_copy = bqm.copy()
+        if scalar is None:
+            bqm_copy.normalize(bias_range, quadratic_range,
+                               ignored_variables=ignored_variables,
+                               ignored_interactions=ignored_interactions,
+                               ignore_offset=ignore_offset)
+        else:
+            bqm_copy.scale(scalar, ignored_variables=ignored_variables,
+                           ignored_interactions=ignored_interactions,
+                           ignore_offset=ignore_offset)
```
arcondello: In this case you don't need to recalculate the energies, you can just scale them

dimod/reference/composites/scalecomposite.py
```diff
@@ -0,0 +1,284 @@
...
+
+       >>> linear = {'a': -4.0, 'b': -4.0}
+       >>> quadratic = {('a', 'b'): 3.2}
+       >>> sampler = dimod.ScaleComposite(dimod.SimulatedAnnealingSampler())
+       >>> response = sampler.sample_ising(linear, quadratic, scalar=0.5,
+       ...                ignored_interactions=[('a','b')],num_reads = 100)
+       >>> print(response.aggregate().record) # doctest: +SKIP
+       >>> response = sampler.sample_ising(linear, quadratic, scalar=0.5,
+       ...                                              num_reads = 100)
+       >>> print(response.aggregate().record) # doctest: +SKIP
```
arcondello: We're about to make new pretty printing (see #370 ), so maybe just remove most of this example. I would also use `ExactSolver` because the return is deterministic.

```
       >>> linear = {'a': -4.0, 'b': -4.0}
       >>> quadratic = {('a', 'b'): 3.2}
       >>> sampler = dimod.ScaleComposite(dimod.ExactSolver())
       >>> sampleset = sampler.sample_ising(linear, quadratic, scalar=0.5,
       ...                ignored_interactions=[('a','b')],num_reads = 100)
```

dimod/reference/composites/scalecomposite.py
conta877: I can't really just rescale them. there are options to ignore certain variables and interactions.

dimod/reference/composites/scalecomposite.py
conta877: I cannot use exactsolver as an example for scalecomposite since scaling changes the probabilities. The outputs would look exactly the same and the examples would be meaningless.

tests/test_scalecomposite.py
```diff
@@ -0,0 +1,314 @@
...
+                                        ignored_interactions=[
+                                            ('a', 'b', 'c')],
+                                        ignored_variables=['a'], num_reads \
+                                            =100)
+
+        samples = [(s, e) for s, e, in
+                   response.aggregate().data(['sample', 'energy'],
+                                             sorted_by='num_occurrences',
+                                             reverse=True)]
+        self.assertIn(samples[0][0], [{'a': 1, 'b': -1, 'c': 1},
```
arcondello: This is not guaranteed. `SimulatedAnnealingSampler` should not be used for testing

dimod/reference/composites/scalecomposite.py
arcondello: Fair point, might just be better to skip showing the output then

PR #379, opened 2019-01-21
Cythonize BQM.energies

arcondello:
Use cython to speed up energy calculations, falling back on numpy if the extension if not built.

Test script:
```
import timeit

import networkx as nx
import numpy as np

import dimod

num_variables = 100
num_samples = 1000

samples = np.ones((num_samples, num_variables), dtype=np.int8)

G = nx.complete_graph(num_variables)

bqm = dimod.BQM.from_ising({}, {edge: -1 for edge in G.edges})

N = 1000

contime = timeit.timeit('bqm.to_numpy_vectors()', 'from __main__ import bqm', number=N)
print('numpy conversion time', contime)

cytime = timeit.timeit('bqm.energies(samples, _use_cpp=True)', 'from __main__ import bqm, samples', number=N)
print('cython', cytime)

nptime = timeit.timeit('bqm.energies(samples, _use_cpp=False)', 'from __main__ import bqm, samples', number=N)
print('numpy', nptime)

print('cython/numpy', cytime/nptime)
```
Output:
```
numpy conversion time 2.7262168080196716
cython 7.390854711004067
numpy 50.06847931200173
cython/numpy 0.14761492285291822
```

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -28,8 +28,9 @@ cache:
   - '%PYTHON%\Lib\site-packages -> requirements.txt'
   - '%AppData%\pip-cache'
 
-  # cache the .pyd file
+  # cache the .pyd files
   - dimod\roof_duality\ -> dimod\roof_duality\_fix_variables.pyx, dimod\roof_duality\src\compressed_matrix.hpp, dimod\roof_duality\src\fix_variables.cpp, dimod\roof_duality\src\fix_variables.hpp, dimod\roof_duality\fix_variables.py, dimod\roof_duality\__init__.py
+  - dimod\cyutils\ -> dimod\cyutils\_energy.pyx
 
 install:
   - "%PYTHON%\\python.exe -m pip install -r requirements.txt --cache-dir %AppData%\\pip-cache"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,15 +96,20 @@ jobs:
     steps: 
       - checkout
 
-      - run: 
-          name: install pyenv
-          command: |
-            brew install pyenv
-
       - run: &install-boost-osx-template
           name: install boost
           command: |
             brew install boost
+
+      - run: &install-openmp-osx-template
+          name: install openmp
+          command: |
+            brew install libomp
+
+      - run: &install-pyenv-osx-template
+          name: install pyenv
+          command: |
+            brew install pyenv
 
       - restore_cache:
           keys:
@@ -275,12 +280,11 @@ jobs:
     steps: 
       - checkout
 
-      - run: 
-          name: install pyenv
-          command: |
-            brew install pyenv
-
       - run: *install-boost-osx-template
+
+      - run: *install-openmp-osx-template
+
+      - run: *install-pyenv-osx-template
 
       - restore_cache:
           keys:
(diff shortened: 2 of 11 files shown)


arcondello:
Running the same benchmark but with a 10-node complete graph and only one sample resulted in the following benchmarks on my system
```
numpy conversion time 0.038362087012501433
cython 0.18449774000328034
numpy 0.07943365999381058
cython/numpy 2.322664472688987
```
and these on circle-ci
```
numpy conversion time 0.00040504799926566193
cython 1.4814547020005193
numpy 0.0008503760000166949
cython/numpy 1742.1172539811034

```

arcondello:
When I ssh into the circleci session and `export OMP_NUM_THREADS=1` it goes back to 'normal' performance, otherwise it is still very slow. See `num_threads` in [prange docs](https://cython.readthedocs.io/en/latest/src/userguide/parallelism.html).

arcondello:
Unset
```
numpy conversion time 0.0005985349998809397
cython 3.5206295359967044
numpy 0.0011095899972133338
cython/numpy 3172.910304561636
```
`OMP_NUM_THREADS=36` (the number of CPUs in circle-ci)
```
numpy conversion time 0.00045068300096318126
cython 3.504014004000055
numpy 0.0016239319957094267
cython/numpy 2157.734445320354
```
`OMP_NUM_THREADS=4`
```
numpy conversion time 0.000588616996537894
cython 0.1617555379998521
numpy 0.0011004350017174147
cython/numpy 146.99236006434296
```
`OMP_NUM_THREADS=1`
```
numpy conversion time 0.000440096999227535
cython 0.0014873429972794838
numpy 0.0014783549995627254
cython/numpy 1.0060797289686285
```

arcondello:
It is currently not building correctly on OSX. I was able to get it to work by following:
* install llvm instructions from https://stackoverflow.com/questions/28010801/compiling-parallelized-cython-with-clang
* symbolic link instructions from https://stackoverflow.com/questions/44579856/ld-library-not-found-for-lomp

To do this I also need to use different linking arguments for clang/gcc (`-lomp`/`-fopenmp`). There doesn't (yet) seem to be an elegent way to determine in `setup.py` whether we are using clang (see https://bugs.python.org/issue35244)

arcondello:
I was also able to get it to work by following https://iscinumpy.gitlab.io/post/omp-on-high-sierra/ which seems cleaner

arcondello:
Was able to resolve the compiler issues.

arcondello:
I would still like to merge this in but needing to overwrite an environmental variable for it to work properly in docker containers seems to be a problem. Documenting some approaches for those who are interested:

A) Try to determine if in a docker container and limit the number of threads using `num_threads` in [prange docs](https://cython.readthedocs.io/en/latest/src/userguide/parallelism.html). Of course, there doesn't seem to be a 'nice' way to do that, [link](https://stackoverflow.com/questions/20010199/how-to-determine-if-a-process-runs-inside-lxc-docker), and it's docker-specific.

B) Create an opt-in approach
* Decide using a function argument, e.g. `bqm.energies(samples, num_threads=2)` where `None` would let openmp choose, defaulting to 1. This just kicks the decision up stream though and does not make for portable code in my opinion.
* Decide at install time, e.g. `pip install dimod[openmp]` whether to use the cythonized parallel code or just use the current numpy implementation. However we probably would not want to pass that up to the sdk, so only people who install dimod directly would benefit from the faster calculations.
* Some sort of global variable :nauseated_face: , though that is essentially what the current `OMP_NUM_THREADS` environmental variable is.

PR #384, opened 2019-01-22
Bson serialization bug fixes2

### LINKED ISSUE #381: `bqm.to_serializable(use_bytes=True)` produces an offset with numpy type
**Description**
The offset value produced by `bqm.to_serializable(use_bytes=True)` is a numpy type, which is not generally serializable.

**Steps To Reproduce**
```python
>>> bqm = dimod.BinaryQuadraticModel.from_ising({}, {"ab": 1})
>>> type(bqm.to_serializable(use_bytes=True)["offset"])
numpy.float32
```

**Expected Behavior**
All the types produced in `bqm.to_serializable(use_bytes=True)` should be bson serializable.

### LINKED ISSUE #382: Calling `to_serializable(use_bytes=True)` on a bqm with 0 or 1 variables results in a ` ZeroDivisionError`
**Description**
If `to_serializable(use_bytes=True)` is called on a bqm with 0 or 1 variables, a `ZeroDivisionError` is raised.

**Steps To Reproduce**
```python
>>> bqm = dimod.BinaryQuadraticModel.from_ising({0: 1}, {})
>>> bqm.to_serializable(use_bytes=True)
...traceback...
ZeroDivisionError: integer division or modulo by zero
```

**Expected Behavior**
All bqms should be serializable without crashing.

wbernoudy:
Closes #381 
Closes #382

### DIFF
--- a/dimod/serialization/bson.py
+++ b/dimod/serialization/bson.py
@@ -38,6 +38,7 @@ def bqm_bson_encoder(bqm):
         index_dtype=index_dtype,
         sort_indices=as_complete,
         variable_order=variable_order)
+    off = float(off)
 
     if as_complete:
         vals = np.zeros(num_possible_edges, dtype=np.float32)
--- a/dimod/serialization/bson.py
+++ b/dimod/serialization/bson.py
@@ -29,7 +29,7 @@ def bqm_bson_encoder(bqm):
         index_dtype = np.uint16
 
     variable_order = sorted(bqm.linear)
-    num_possible_edges = num_variables*(num_variables - 1) // 2
+    num_possible_edges = max(num_variables*(num_variables - 1) // 2, 1)
     density = len(bqm.quadratic) / num_possible_edges
     as_complete = density >= 0.5
 
--- a/dimod/serialization/bson.py
+++ b/dimod/serialization/bson.py
@@ -28,7 +28,7 @@ def bqm_bson_encoder(bqm):
     if num_variables <= 2**16:
         index_dtype = np.uint16
 
-    variable_order = sorted(bqm.linear)
+    variable_order = list(bqm.variables)
     num_possible_edges = max(num_variables*(num_variables - 1) // 2, 1)
     density = len(bqm.quadratic) / num_possible_edges
     as_complete = density >= 0.5
--- a/dimod/serialization/bson.py
+++ b/dimod/serialization/bson.py
@@ -21,7 +21,7 @@ import numpy as np
 from dimod.binary_quadratic_model import BinaryQuadraticModel
 
 
-def bqm_bson_encoder(bqm, bytes_type=bytes):
+def bqm_bson_encoder(bqm, bytes_type=bytes, bias_dtype=np.float32):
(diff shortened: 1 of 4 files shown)


dimod/serialization/bson.py
```diff
@@ -13,31 +13,43 @@
...
 
 from dimod.binary_quadratic_model import BinaryQuadraticModel
 
+from six import PY2
+
 
 def bqm_bson_encoder(bqm):
     """todo"""
+
+    if PY2:
```
arcondello: Ok, I see what this is trying to do, but it definitely needs at least a comment explaining

dimod/serialization/bson.py
```diff
@@ -13,31 +13,43 @@
...
     index_dtype = np.uint32
     if num_variables <= 2**16:
         index_dtype = np.uint16
 
-    variable_order = sorted(bqm.linear)
-    num_possible_edges = num_variables*(num_variables - 1) // 2
-    density = len(bqm.quadratic) / num_possible_edges
+    variable_order = list(bqm.variables)
+    num_possible_edges = max(num_variables*(num_variables - 1) // 2, 1)
+    density = len(bqm.quadratic) / float(num_possible_edges)
```
arcondello: you might want to `from __future__ import division` while you're at it

dimod/serialization/bson.py
```diff
@@ -52,19 +64,26 @@ def mul(a, b):
...
         "variable_order": variable_order,
         "index_dtype": np.dtype(index_dtype).str,
     }
 
     if not as_complete:
-        doc["quadratic_head"] = i.tobytes()
-        doc["quadratic_tail"] = j.tobytes()
+        ii, jj = i.tobytes(), j.tobytes()
+        if PY2:
+            ii, jj = bson.Binary(ii), bson.Binary(jj)
```
arcondello: It might be better to make a function `tobytes` in `dimod/compatibility23.py` or something like that, just to make it a little cleaner.

tests/test_serialization_bson.py
```diff
@@ -0,0 +1,92 @@
+import unittest
```
arcondello: need apache header

dimod/serialization/bson.py
```diff
@@ -13,31 +13,43 @@
...
     num_variables = len(bqm)
 
     index_dtype = np.uint32
     if num_variables <= 2**16:
         index_dtype = np.uint16
 
-    variable_order = sorted(bqm.linear)
-    num_possible_edges = num_variables*(num_variables - 1) // 2
-    density = len(bqm.quadratic) / num_possible_edges
+    variable_order = list(bqm.variables)
```
wbernoudy: If we want serialization to be unique (ie return the same for different "orderings" of bqm.variables), we should have a consistent ordering of the variables labels here.

arcondello: `sorted(bqm.variables, key=hash)`

arcondello:
Marking as WIP for now, just comment here when you want me to take a look again!

Issue #391, opened 2019-01-30
Support higher order polynomials


arcondello:
Right now in dimod there is 'dimod/higherorder.py` which supports higher-order Ising problems. But as this is proving to be a more common use case, we should consider creating a proper `HUBO` (or similar)
class.

We could also define a Sampler ABC similar to `Structured`, to signal samplers/composites that support HUBOs. These would then implement a `sample_hubo` method.

conta877:
this also effects #330 and #372

PR #392, opened 2019-01-30
TruncateComposite

conta877:
a new composite to trim response


arcondello:
Can be simplified if/when #394 is addressed

dimod/reference/composites/trimcomposite.py
```diff
@@ -0,0 +1,169 @@
...
+        if not (isinstance(n, int) and n >= 0):
+            raise ValueError('number of samples requested must be >0. '
+                             'receieved {}'.format(n))
+
+        child = self.child
+        response = child.sample(bqm, **parameters)
+
+        return _trim(response, n, aggregate)
+
+    def sample_ising(self, h, J, n=None, aggregate=False,
```
arcondello: Can be simplified when/if #391 is addressed

tests/test_truncatecomposite.py
```diff
@@ -0,0 +1,64 @@
...
+        # this is actually just a smoke test, needs better testing in the
+        # future...
+        h = {'a': -4.0, 'b': -4.0, 'c': 0}
+        J = {('a', 'b'): 3.2}
+
+        sampler = TruncateComposite(ExactSolver(), 6, aggregate=True)
+        samples = sampler.sample_ising(h, J)
+
+        # we should see 2**n rows
+        self.assertEqual(len(samples), 6)
```
m3ller: In the TruncateComposite doc-string, the parameter `n (int)` is described as `MAXIMUM number of rows in the returned sample set`. In all the unit tests in `TestSample`, the `n` chosen was `<= the number of available samples`. It would nice to see a test case where `n > number of available samples` - is this possible? Or is the number of available samples theoretically infinite?

Another unit test that I think would be useful is when `n = 0`. By `n` I am referring to TruncateComposite's truncate_kwargs' `n`.

tests/test_truncatecomposite.py
```diff
@@ -0,0 +1,64 @@
...
+
+class TestSample(unittest.TestCase):
+    def test_sampleset_shorter(self):
+        h = {'a': -4.0, 'b': -4.0, 'c': 0}
+        J = {('a', 'b'): 3.2}
+
+        sampler = TruncateComposite(ExactSolver(), 10)
+        samples = sampler.sample_ising(h, J)
+
+        # we should see 2**n rows
```
m3ller: It's not super obvious to me what `n` is referring to. Based on context, I believe it's `2**len(h)`, but I'm not sure because I know TruncateComposite has a parameter `n` as well.

tests/test_truncatecomposite.py
arcondello: The first unittest covers the n > number of samples case. But I should include a test for 0

tests/test_truncatecomposite.py
arcondello: that's right, but those comments were erroneous and confusing.

tests/test_truncatecomposite.py
m3ller: Yes, I saw the first unittest, `test_10`. To clarify, I was curious to see if `sampler.sample_ising(h, J)` (used in the TestSample class unittests) can be called for the `n > number of samples` case.

PR #396, opened 2019-02-03
Support higher-order binary problems

### LINKED ISSUE #373: Erroneous embedding reference in docs
**Description**
Remove embedding reference from documentation

https://github.com/dwavesystems/dimod/edit/master/docs/reference/bqm_functions.rst

### LINKED ISSUE #388: Deprecation warning: np.sum(generator)
https://github.com/dwavesystems/dimod/blob/26f0abd2eb2e57cc2a5ccc8c7d2aa6202d7dbed4/dimod/higherorder.py#L317
```
DeprecationWarning: Calling np.sum(generator) is deprecated, and in the future will give a different result. Use np.sum(np.from_iter(generator)) or the python sum builtin instead.
  return np.sum(set(key) == set(key_tmp) for key_tmp in key_list)
```

### LINKED ISSUE #391: Support higher order polynomials

arcondello:
This is a work-in-progress but there is enough code to start to get an idea of it.

Closes #391 
Closes #388
Closes #373


arcondello:
One thing that occurs to me is that using frozenset under the hood rules out `a^2` for instance.

dimod/reference/composites/higherordercomposites.py
```diff
@@ -0,0 +1,351 @@
...
+
+    @property
+    def properties(self):
+        return {'child_properties': self.child.properties.copy()}
+
+    def sample_poly(self, poly, scalar=None, bias_range=1, poly_range=None,
+                    ignored_terms=None, **parameters):
+        """Scale and sample from the given binary polynomial.
+
+        Ff scalar is not given, problem is scaled based on bias and polynomial
```
conta877: 'Ff' ?

dimod/reference/composites/higherordercomposites.py
```diff
@@ -0,0 +1,351 @@
...
+        if scalar is not None:
+            poly.scale(scalar, ignored_terms=ignored_terms)
+        else:
+            poly.normalize(bias_range=bias_range, poly_range=poly_range,
+                           ignored_terms=ignored_terms)
+
+            # we need to know how much we scaled by, which we can do by looking
+            # at the biases
+            try:
+                v = next((v for v, bias in original.items() if bias))
```
conta877: what if the first one is in ignored terms? - tad worried that tests don't catch this

dimod/higherorder/polynomial.py
```diff
@@ -0,0 +1,527 @@
...
+        any sequence.
+
+        >>> poly = dimod.BinaryPolynomial({'a': -1, 'ab': 1}, dimod.BINARY)
+        >>> poly['ab']
+        1
+        >>> poly['ba']
+        1
+        >>> poly[{'a', 'b'}]
+        1
+        >>> poly[('a', 'b')]
```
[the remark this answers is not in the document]
arcondello: Not sure what you mean? It is contrast to the previous (set) example.

dimod/reference/composites/higherordercomposites.py
arcondello: Good catch!

arcondello: It worked because if we detected ignored terms we just recalculate the energies anyway, but I have updated it to calculate the scalar correctly in case that logic changes later.

PR #399, opened 2019-02-05
Add SampleSet.truncate method

### LINKED ISSUE #394: SampleSet.trim or similar
Something like
```
samples.trim(n, 'energy')
```
which would return a new sampleset containing the first n rows of the sampleset ordered by energy.

We could then do things like
```
samples.aggregate().trim(n)
```

arcondello:
Closes #394

### DIFF
--- a/tests/test_sampleset.py
+++ b/tests/test_sampleset.py
@@ -243,17 +243,17 @@ class TestSampleSet(unittest.TestCase):
 
     def test_truncate(self):
         bqm = dimod.BinaryQuadraticModel.from_ising({v: -1 for v in range(100)}, {})
-        samples = dimod.SampleSet.from_samples_bqm(np.triu(np.ones(100)), bqm.binary)
+        samples = dimod.SampleSet.from_samples_bqm(np.tril(np.ones(100)), bqm.binary)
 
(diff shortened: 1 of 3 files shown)


tests/test_sampleset.py
```diff
@@ -241,6 +241,34 @@ def test_from_bqm_single_sample(self):
         self.assertEqual(samples,
                          dimod.SampleSet.from_samples(([-1, 1], 'ab'), dimod.SPIN, energy=1))
 
+    def test_truncate(self):
+        bqm = dimod.BinaryQuadraticModel.from_ising({v: -1 for v in range(100)}, {})
+        samples = dimod.SampleSet.from_samples_bqm(np.triu(np.ones(100)), bqm.binary)
+
+        # by default should be in reverse order
```
conta877: for this, shouldn't the bqm need to prepped with v:1

arcondello: I do not understand the question. The order stored in the underlying array is in reverse energy order (because of the bqm). So truncating by energy then iterating by energy low-to-high should give low energy solutions (mostly all 1s)

arcondello: I understand now, will fix

PR #409, opened 2019-02-15
Do not keep unset linear biases in memory

arcondello:
In the case that the linear biases are not specified by the user and are assumed to be 0, we don't store them in the underlying `BQM._adj` dict-of-dicts.

### DIFF
--- a/dimod/binary_quadratic_model.py
+++ b/dimod/binary_quadratic_model.py
@@ -442,7 +442,10 @@ class BinaryQuadraticModel(abc.Sized, abc.Container, abc.Iterable):
         # is much faster
         _adj = self._adj
         if v in _adj:
-            _adj[v][v] += bias
+            if v in _adj[v]:
+                _adj[v][v] += bias
+            else:
+                _adj[v][v] = bias
         else:
             _adj[v] = {v: bias}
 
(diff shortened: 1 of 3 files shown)


jberwald:
This seems like it could affect the `to_networkx_graph()` method, though I'm not familiar enough either with the implementation to know. Here's my concern: If biases are transferred to the graph data via linear terms, but the nodes are absent, will the node data be missing (even though `bias = 0` it may still be desirable to have that info)?

arcondello:
Hey @jberwald , it shouldn't affect `to_networkx_graph` because that method uses `BQM.linear` and `BQM.quadratic`. Even in the case that it used `BQM.adj`, this change affects the underlying `BQM._adj`, *not* the public attributes. So
```
>>> bqm = dimod.BinaryQuadraticModel.empty(dimod.SPIN)
>>> bqm.add_interaction({('a', 'b'): -1})
>>> print(bqm.linear)
{'a': 0, 'b': 0}
```
just like before. The only difference is we are not actually storing those 0's in memory, we generate them on read.

We could then do
```
>>> bqm.add_variable('a', 1)
>>> print(bqm.linear)
{'a': 1, 'b': 0}
```
and in this case the 1 associated with 'a' is kept.

arcondello:
I have changed the commit message to make that clearer.

PR #425, opened 2019-03-02
Support simplejson encoding

arcondello:


dimod/binary_quadratic_model.py
```diff
@@ -1755,6 +1755,10 @@ def to_serializable(self, use_bytes=False, bias_dtype=np.float32,
 
         return doc
 
+    def _asdict(self):
+        # support simplejson encoding
+        return self.to_serializable()
+
```
randomir: Yes, that's it. Unfortunately that doesn't work for stdlib's `json` module, although `json` is based on `simplejson` (I wasn't aware they omitted some parts. Or maybe they just forked an older version of `simplejson`.. Don't know, will have to check.)

randomir:
Just to follow-up on my previous comment, for all the interested parties. :smile: 

CPython's `json` module is based on `simplejson==2.0.9`.

The functionality I was proposing in this PR - support for `_asdict` encoder method (introduced for named tuple serialization via `namedtuple_as_object` parameter) - was introduced in `simplejson==2.2.0` in 2011.

Just for a reference, the version included in stdlib (fork of `simplejson` 2.0.9) is from 2009-02-18. :scream:

PR #426, opened 2019-03-04
QOL improvements for BQM generators

### LINKED ISSUE #371: Problem generators should accept seeds
**Application**
I love the problem generators in `dimod.generators`. It would be great if they accepted a seed so they could be reproducible!

**Proposed Solution**
Accept a `seed` parameter as an int, and pass that into `random.Random()`.

### LINKED ISSUE #424: chimera_anticluster should accept a working chimera graph
**Application**
make it more consistent with the rest of generators :)

arcondello:
Closes #371 
Closes #424

### DIFF
--- a/dimod/generators/chimera.py
+++ b/dimod/generators/chimera.py
@@ -61,7 +61,7 @@ def chimera_anticluster(m, n=None, t=4, multiplier=3.0,
 
     """
     if seed is None:
-        seed = numpy.random.randint(65536)
+        seed = numpy.random.randint(2**32, dtype=np.uint32)
     r = numpy.random.RandomState(seed)
 
     m = int(m)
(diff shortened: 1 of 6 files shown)


arcondello:
@wbernoudy pointed out that I should use [numpy.random.RandomState](https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.RandomState.html). Will update

wbernoudy:
We should consider using `numpy.random.RandomState` when a seed is passed instead of seeding the global `random` state, as this will preserve expected behavior of `random` from the user's point of view.

dimod/generators/chimera.py
```diff
@@ -35,33 +40,80 @@ def chimera_anticluster(m, n=None, t=4, multiplier=3.0, cls=BinaryQuadraticModel
...
+
+        seed (int, optional, default=None):
+            Random seed.
+
     Returns:
         :obj:`.BinaryQuadraticModel`: spin-valued binary quadratic model.
 
     """
+    if seed is None:
+        seed = numpy.random.randint(65536)
```
wbernoudy: This should be 2^32 not 2^16.

PR #432, opened 2019-03-12
Add k-combinations generator

arcondello:
Creates a bqm that is minimized when k of n variables are equal to 1.

Note that the function has not been added to the online documentation, see #433.

### DIFF
--- a/dimod/generators/constraints.py
+++ b/dimod/generators/constraints.py
@@ -1,0 +1,107 @@
+# Copyright 2019 D-Wave Systems Inc.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+# =============================================================================
+import itertools
+
+try:
+    import collections.abc as abc
+except ImportError:
+    import collections as abc
+
+from dimod.binary_quadratic_model import BinaryQuadraticModel
+from dimod.decorators import graph_argument, vartype_argument
+from dimod.vartypes import BINARY
+
+__all__ = 'combinations',
+
+
+def combinations(n, k, strength=1, vartype=BINARY):
+    r"""Generate a bqm that is minimized when k of n variables are selected.
+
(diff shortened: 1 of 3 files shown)


dimod/generators/constraints.py
```diff
@@ -0,0 +1,107 @@
...
+try:
+    import collections.abc as abc
+except ImportError:
+    import collections as abc
+
+from dimod.binary_quadratic_model import BinaryQuadraticModel
+from dimod.decorators import graph_argument, vartype_argument
+from dimod.vartypes import BINARY
+
+__all__ = 'combinations',
```
randomir: I noticed your increasing usage of this *implicit tuple assignment to `__all__`* pattern, and it's killing me. :)

If you hate `__all__` as a list (*the standard* pattern), use tuples - but please make them explicit. :)

arcondello: Well if it's the *standard* fine, but my way is still much prettier

randomir: Well, it's the standard way for a reason: https://zen-of-python.info/explicit-is-better-than-implicit.html#2 :)

arcondello: Feel like I could make a case for readability counts if I had to

randomir: Not sure I follow - are you suggesting that trailing comma is more readable than an explicit tuple or list?

arcondello: To be clear I will go back to the ugly list way! But consider this to be my official desired exception to python standards

randomir: Readability impact:
```
__all__ = ['combinations']
# vs
__all__ = 'combinations',
```

Gotcha impact:
```
__all__ = 'combinations',
          'ignored'

# to work, you need the ugly escape:
__all__ = 'combinations',\
          'ignored'
```

```
# small input difference, major impact
__all__ = 'combinations',
__all__ = 'combinations'
```

arcondello: I was going to argue that my way saves from
```
__all__ = ['hello', 'world']
__all__ = ['hello' 'world']
```
but turns out
```
a = 'hello' 'world'
```
is valid in python, who knew

arcondello: Also the last case would throw an error

randomir: What case would throw an error?

randomir: And regarding the "who knew" part; that's actually a pattern I like to use :)

https://github.com/dwavesystems/dwave-hybrid/blob/master/hybrid/flow.py#L637

arcondello: I meant `__all__ = 'combinations'` or `__all__ = ['hello' 'world']`, assuming `helloworld` was not an object in the module.
But I mis-remembered, it only throws an error if you try to `from example import *` and the `__all__` is example has something not actually present.

PR #436, opened 2019-03-13
Make the sidebar external links into hyperlinks

arcondello:


arcondello:
Kept it the same as you had it!

PR #439, opened 2019-03-15
Make SampleSet.samples return an iterable

### LINKED ISSUE #407: Return iterable from SampleSet.samples
It would be nice for `.samples()` to also return a view which would be iterable **and indexable**, so we could write:
```
initial = forward_answer.samples()[i5]
```

arcondello:
Closes #407

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -23,9 +23,10 @@ from dimod.views.samples import SamplesArray
 class TestSamplesArray(unittest.TestCase):
     # SampleSet.samples
     def setUp(self):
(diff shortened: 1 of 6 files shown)


arcondello:
Force push to rebase against #440

tests/test_views.py
```diff
@@ -0,0 +1,87 @@
...
+import unittest
+
+import numpy as np
+
+from dimod.views.samples import SamplesArray
+
+
+class TestSamplesArray(unittest.TestCase):
+    # SampleSet.samples
+    def setUp(self):
```
m3ller: Cool! I didn't know about setUp(..) and tearDown(..).

Issue #213, opened 2018-07-05
Document and test bqm_index_labels


arcondello:
https://github.com/dwavesystems/dimod/blob/38032409d73031b58f08bf27074539dd3803c32f/dimod/decorators.py#L29

The `bqm_index_labels` decorator exists and is used by `ExactSolver` but it is not referenced in the documentation and there are no unittests.

PR #443, opened 2019-03-21
Reorganize utilities docs for dimod

arcondello:
Addresses the documentation part of #213


dimod/decorators.py
```diff
@@ -34,7 +40,13 @@
 
 
 def bqm_index_labels(f):
-    """todo
+    """Decorator to convert a bqm to index-labels and relabel the sample set
+    output.
+
+    Designed to be applied to :meth:`.Sampler.sample`. Expects the wrapped
```
JoelPasvolsky: Applied to :meth:`.Sampler.sample`.

dimod/decorators.py
```diff
@@ -127,9 +139,12 @@ def _index_label(sampler, bqm, **kwargs):
 
 
 def bqm_structured(f):
-    """todo
+    """Decorator to raise an error if the given bqm does not match the sampler's
+    structure.
 
-    makes sure bqm has the appropriate structure
+    Designed to be applied to :meth:`.Sampler.sample`. Expects the wrapped
```
JoelPasvolsky: Applied to :meth:`.Sampler.sample` 
and elsewhere if occurs more.

dimod/decorators.py
arcondello: No, that is not correct. It is designed to be applied to that method, but it could be used elsewhere

PR #445, opened 2019-03-25
SampleSet.append method

arcondello:
Also update `FixedVariableComposite` to make use of the new method.

### DIFF
--- a/dimod/sampleset.py
+++ b/dimod/sampleset.py
@@ -1002,8 +1002,8 @@ class SampleSet(abc.Iterable, abc.Sized):
 
         return type(self)(record, self.variables, self.info, self.vartype)
 
-    def append(self, samples_like, sort_labels=True):
-        """Create a new sampleset with the given samples added.
+    def append_variables(self, samples_like, sort_labels=True):
+        """Create a new sampleset with the given variables with values added.
 
         Empty sample sets cannot be appended to.
 
@@ -1029,7 +1029,7 @@ class SampleSet(abc.Iterable, abc.Sized):
             ...                                           {'a': +1, 'b': +1}],
             ...                                          dimod.SPIN,
             ...                                          energy=[-1.0, 1.0])
-            >>> new = sampleset.append({'c': -1})
+            >>> new = sampleset.append_variables({'c': -1})
             >>> print(new)
                a  b  c energy num_oc.
             0 -1 +1 -1   -1.0       1
@@ -1042,7 +1042,7 @@ class SampleSet(abc.Iterable, abc.Sized):
             ...                                         {'c': +1, 'd': +1}],
             ...                                        dimod.SPIN,
             ...                                        energy=[-2.0, 1.0])
-            >>> new = sampleset.append(another)
+            >>> new = sampleset.append_variables(another)
             >>> print(new)
                a  b  c  d energy num_oc.
             0 -1 +1 -1 +1   -3.0       1
--- a/dimod/sampleset.py
+++ b/dimod/sampleset.py
@@ -1014,9 +1014,9 @@ class SampleSet(abc.Iterable, abc.Sized):
                 case the given order will be maintained.
 
         Returns:
-            :obj:`.SampleSet`: A new sample set with the samples appended. Note
-            that the returned sample set's energies are taken from the base
-            sample set.
+            :obj:`.SampleSet`: A new sample set with the samples appended. If
+            `samples_like` is another sample set then the returned sample set
+            has the sum of the energies.
 
         Examples:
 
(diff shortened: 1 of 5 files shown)


arcondello:
Ok, I think I will make it a function, not a method and then rename is `hstack`. I will probably also rename alias `concatenate` with `vstack`

arcondello:
As summarized in the commit message, hstack (if we want to be consistent with numpy) is actually a different notion than the logic encoded in this function. It would probably require each of the stacked sample-like to have the same size for instance. Also we would need to decide how to resolve the meta data.

dimod/sampleset.py
```diff
@@ -997,6 +1002,90 @@ def aggregate(self):
...
+        new_samples = np.hstack((self.record.sample, samples))
+
+        new = type(self).from_samples((new_samples, new_variables),
+                                      self.vartype,
+                                      info=dict(self.info),  # make a copy
+                                      sort_labels=sort_labels,
+                                      **self.data_vectors)
+
+        if isinstance(samples_like, SampleSet):
+            new.record.energy += samples_like.record.energy
```
randomir: It's weird to special-handle energies, out of all info vectors. What about `num_occurrences`, for example?

dimod/sampleset.py
```diff
@@ -997,6 +1002,90 @@ def aggregate(self):
 
         return type(self)(record, self.variables, self.info, self.vartype)
 
+    def append_variables(self, samples_like, sort_labels=True):
```
randomir: `append_variables` could work. I still like `hstack`, the only problem is one could expect the semantics to extend to info vectors too. (If we went with `hstack`, I would ditch the special case of shape mismatch when adding just one sample.)

dimod/sampleset.py
```diff
@@ -997,6 +1002,90 @@ def aggregate(self):
...
+                sample or should match the length of the sample set. See
+                :func:`.as_samples` for what is allowed to be `samples_like`.
+
+            sort_labels (bool, optional, default=True):
+                If true, returned :attr:`.SampleSet.variables` will be in
+                sorted-order. Note that mixed types are not sortable in which
+                case the given order will be maintained.
+
+        Returns:
+            :obj:`.SampleSet`: A new sample set with the samples appended. If
```
randomir: Again, when I read "samples appended" - the only possible interpretation of that in my head is samples are appended vertically. While, here, samples are actually *combined* on a sample-by-sample basis.

docs/reference/sampleset.rst
```diff
@@ -50,6 +50,7 @@ Methods
    :toctree: generated/
 
    SampleSet.aggregate
+   SampleSet.append
```
randomir: Leftover

dimod/sampleset.py
```diff
@@ -997,6 +1002,90 @@ def aggregate(self):
...
+        if any(v in variables for v in labels):
+            msg = "Appended samples cannot contain variables in sample set"
+            raise ValueError(msg)
+
+        new_variables = list(variables) + labels
+        new_samples = np.hstack((self.record.sample, samples))
+
+        new = type(self).from_samples((new_samples, new_variables),
+                                      self.vartype,
+                                      info=dict(self.info),  # make a copy
```
randomir: Do you want to actually `deepcopy` here?

dimod/sampleset.py
arcondello: Adding `num_occurences` does not make sense, energies is meaningful at least. Other option is to just ignore them all

dimod/sampleset.py
arcondello: Good question. In my mind that's on the user, but I suppose there is no harm in deepcopying

dimod/sampleset.py
arcondello: Yes, ditching the single-sample case is what I did when I went to implement hstack. But since adding/multipling/joining are all valid options depending on the vector content, I think we will need to leave that to the user.

dimod/sampleset.py
arcondello: Will change the language

dimod/sampleset.py
conta877: I don't agree with ignoring all.
energy is the only thing that changes with fixedvariables. num_occurrences should stay the same

arcondello: Energy does not change in fixed variables because we use the offset

dimod/sampleset.py
conta877: for this case, adding just one sample is pretty important. that would be the default use case

dimod/sampleset.py
conta877: agree with @randomir here

dimod/sampleset.py
randomir: Actually, ignoring all is the only option that makes sense in the `append_variables` method.

dimod/sampleset.py
arcondello: Yes, I think hstack is a different function with different logic. These two actions are complementary and trying to integrate them into a single function would just be confusing.

dimod/sampleset.py
conta877: ok I get it. samples on their own don't exist anyway. yes ignore all ftw

dimod/sampleset.py
randomir: Since we already treat `SampleSet` as an immutable object, deepcopying makes more sense here (despite the performance penalty). Just consider user has mutable values in `info` dict... Now after the append, they would get a fresh copy of `SampleSet`, but the values in their `info` dicts would point to the same container(s).

dimod/sampleset.py
randomir: Agreed. If we want to handle add-one-sample-to-many logic (we do), name `append_variables` is better.

dimod/sampleset.py
arcondello: Also fixed this in other methods.

PR #460, opened 2019-04-19
Architecture-agnostic coordinate converter and checkerboard gauge transformations.

joseppinilla:
This is an implementation of the transformations proposed in [Adachi's paper](https://arxiv.org/abs/1510.06356).

However, in order to be able to perform those transformations, regardless of the target graph (chimera, pegasus) I had to come up with 2 things:

1. Have a way of retrieving the information of the underlying architecture.
I did this by passing the full graph (_target_graph_dnx_) as a dwave_networkx graph. However, it would be great to have a way of getting all that information from a Structured Sampler. In fact, only the _graph_ attribute is needed.

2. I had to create an "architecture-agnostic" _coordinates_ class that returns a qubit's "shore" and "tile".
This would be better suited for _dwave_networkx_ but had to implement it here for proof of concept.

### DIFF
--- a/dimod/reference/composites/__init__.py
+++ b/dimod/reference/composites/__init__.py
@@ -14,6 +14,7 @@
 #
 # ================================================================================================
 
+from dimod.reference.composites.checkerboard_transform import *
 from dimod.reference.composites.fixedvariable import FixedVariableComposite
 from dimod.reference.composites.higherordercomposites import *
 from dimod.reference.composites.roofduality import RoofDualityComposite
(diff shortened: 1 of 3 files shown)


joseppinilla:
Will do. Thanks!

About this:
> Also I think for the coordinate-agnostic, you should check out the 'nice' coordinates introduced in [dwavesystems/dwave_networkx#95](https://github.com/dwavesystems/dwave_networkx/pull/95) and revised in [dwavesystems/dwave_networkx#106](https://github.com/dwavesystems/dwave_networkx/pull/106).

In the coordinate-agnostic class I used the `pegasus_to_nice` and `nice_to_pegasus` for various reasons.
e.g. to get the `tile` from `labels==int`. I had to do `int` -> `tuple` -> `nice`, from which I can get `t`, `i` and `j`.

arcondello:
Migrated to dwave-system

PR #462, opened 2019-04-19
Add RANr generator

JoelPasvolsky:


JoelPasvolsky:
So why is it I didn't just generalize to a RANk generator? 
image

arcondello:
Can you add the citation for the definition?

dimod/generators/random.py
```diff
@@ -133,3 +133,72 @@ def randint(graph, vartype, low=0, high=1, cls=BinaryQuadraticModel,
...
+    Args:
+        r (int):
+            Order of the RANr problem.
+
+        graph (int/tuple[nodes, edges]/:obj:`~networkx.Graph`):
+            The graph to build the binary quadratic model (BQM) for. Either an
+            integer n, interpreted as a complete graph of size n, or
+            a nodes/edges pair, or a NetworkX graph.
+
+        vartype (:class:`.Vartype`/str/set):
```
arcondello: RANr problems really only make sense for SPIN-valued BQMs. I think this should just always return SPIN.

JoelPasvolsky: Done

PR #464, opened 2019-04-19
Add sampleset slice

### LINKED ISSUE #463: Add SampleSet.slice()
Modeled after Python's `slice`, we could make picking top/bottom/middle n samples easier.

`SampleSet.truncate(n)` could be expressed as `SampleSet.slice(n)`.

Top `n` samples by energy would be `SampleSet.slice(-n, None)`.

Reversing the sample set, according to a custom `weirdness` vector would be expressed with: `SampleSet.slice(None, None, -1, sorted_by='weirdness')`.

randomir:
Closes #463.

### DIFF
--- a/dimod/sampleset.py
+++ b/dimod/sampleset.py
@@ -1187,6 +1187,88 @@ class SampleSet(abc.Iterable, abc.Sized):
         return type(self)(record, self.variables, copy.deepcopy(self.info),
                           self.vartype)
 
+    def slice(self, *slice_args, sorted_by='energy'):
+        """Create a new SampleSet with rows sliced according to standard `slice`
+        syntax.
+
+        Args:
+            start (int, optional, default=None):
+                Start index for `slice`.
+
+            stop (int):
+                Stop index for `slice`.
+
+            step (int, optional, default=None):
+                Step value for `slice`.
+
+            sorted_by (str/None, optional, default='energy'):
+                Selects the record field used to sort the samples before
+                truncating. Note that sample order is maintained in the
+                underlying array.
+
+        Returns:
+            :obj:`.SampleSet`
+
+        Examples:
+
+            >>> import numpy as np
+            ...
+            >>> sampleset = dimod.SampleSet.from_samples(np.diag(range(1, 11)), dimod.BINARY, energy=range(10))
+            >>> print(sampleset)
+               0  1  2  3  4  5  6  7  8  9 energy num_oc.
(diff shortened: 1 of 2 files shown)


dimod/sampleset.py
```diff
@@ -1187,6 +1187,89 @@ def truncate(self, n, sorted_by='energy'):
...
+
+            stop (int):
+                Stop index for `slice`.
+
+            step (int, optional, default=None):
+                Step value for `slice`.
+
+            sorted_by (str/None, optional, default='energy'):
+                Selects the record field used to sort the samples before
+                truncating. Note that sample order is maintained in the
```
JoelPasvolsky: Users who don't know the structure with an underlying array will be confused over whether the original or sorted order is the order that will be maintained going forward. "The original sample order is maintained after slicing." seems clearer to me.

dimod/sampleset.py
```diff
@@ -1187,6 +1187,89 @@ def truncate(self, n, sorted_by='energy'):
...
+
+            >>> # every second sample in between (skip the top and the bottom 3)
+            >>> print(sampleset.slice(3, -3, 2))
+               0  1  2  3  4  5  6  7  8  9 energy num_oc.
+            0  0  0  0  1  0  0  0  0  0  0      3       1
+            1  0  0  0  0  0  1  0  0  0  0      5       1
+            ['BINARY', 2 rows, 2 samples, 10 variables]
+
+        """
+        sorted_by = kwargs.pop('sorted_by', 'energy')
```
arcondello: I think we should throw an error if there is an unknown kwarg argument. I know
```
def f(*args, akwarg=False): pass
```
is invalid syntax in python2 but to keep the behaviour consistent with python3 and to prevent a very hard-to-find bug like
```
sampleset.slice(6, sortedby='weirdness')
```
I think we should add
```
if kwargs:
   # throw same error as python3
    raise TypeError('slice got an unexpected keyword argument {!r}'.format(kwargs.pop())
```
or similar.

randomir: Sounds good, will update.

arcondello:
:+1: Ready to merge?

randomir:
Not yet.

dimod/sampleset.py
randomir: You're right, thanks, I fixed this.

randomir: re: `slice`, I want to refer users to the [`slice`](https://docs.python.org/3/library/functions.html#slice) built-in.

PR #472, opened 2019-04-26
Prevent from_numpy_vectors from silently truncating 

JoelPasvolsky:
I think the risk a user inadvertently creates mismatched-length vectors and gets a BQM that is silently truncated by 
```
quadratic = {(variable_order[u], variable_order[v]): float(bias)
                     for u, v, bias in zip(heads, tails, values)}
``` 
, which will be hard to diagnose, especially for large numbers of variables, is high and the inconvenience of maybe having to set shorten vectors that would have been happily trimmed is not a big deal.


JoelPasvolsky:
Needs #473 to pass unit testing

arcondello:
@JoelPasvolsky  You'll need to rebase

JoelPasvolsky:
Rebased

PR #478, opened 2019-05-07
Support pickling sample sets

### LINKED ISSUE #474: Pickling SampleSets fails for sample sets with unresolved futures
When pickling a `SampleSet` with an unresolved future, the sample set should wait until the future is resolved

### LINKED ISSUE #475: SampleSet.resolve() method or similar
It would be useful to have a method that can be used to ensure that a sample set is resolved. This would be useful for #474

arcondello:
Closes #474 
Closes #475

### DIFF
--- a/dimod/sampleset.py
+++ b/dimod/sampleset.py
@@ -580,6 +580,16 @@ class SampleSet(abc.Iterable, abc.Sized):
 
         return (self.record.sample == other.record.sample[:, other_idx]).all()
 
+    def __getstate__(self):
+        """Ensure that any futures are resolved before pickling."""
+        self.resolve()
+        return {attr: getattr(self, attr)
+                for attr in self.__slots__ if hasattr(self, attr)}
+
+    def __setstate__(self, state):
+        for attr, obj in state.items():
+            setattr(self, attr, obj)
+
     def __repr__(self):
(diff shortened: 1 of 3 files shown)


arcondello:
Probably this should have been two separate PRs, I can separate them if needed.

dimod/sampleset.py
```diff
@@ -580,6 +574,16 @@ def __eq__(self, other):
...
 
+    def __getstate__(self):
+        """Ensure that any futures are resolved before pickling."""
+        self.resolve()
+        return {attr: getattr(self, attr)
+                for attr in self.__slots__ if hasattr(self, attr)}
+
+    def __setstate__(self, state):
+        for attr, obj in state.items():
+            setattr(self, attr, obj)
```
randomir: Could you just defer pickling to super methods, after the future is resolved?

arcondello: I tried that, but because we're using `__slots__` there is no `super().__getstate__`. We could maybe use `__reduce__` but that's getting a bit deep and I wasn't able to get it to work in a way I felt confident about edge cases

arcondello: Also to support python2 format which doesn't support `__slots__` at all.

randomir: I see. OK, let's go with this than.

PR #484, opened 2019-05-22
Add TrackingComposite

### LINKED ISSUE #481: A sampler for composites that will not sample but give access to modified problem
It would be cool to have a "sampler" that does not do sampling but gives access to the problem that is generated by the composite. 

I'm thinking something like
```
sampler = Composite(SaidSampler())
response = sampler.sample_ising(h,J,**kwargs)
response.bqm
response.h
response.j
```

and such.

arcondello:
Closes #481

### DIFF
--- a/dimod/reference/composites/__init__.py
+++ b/dimod/reference/composites/__init__.py
@@ -20,4 +20,5 @@ from dimod.reference.composites.roofduality import RoofDualityComposite
 from dimod.reference.composites.scalecomposite import *
 from dimod.reference.composites.spin_transform import *
 from dimod.reference.composites.structure import *
+from dimod.reference.composites.tracking import TrackingComposite
 from dimod.reference.composites.truncatecomposite import TruncateComposite
--- a/dimod/reference/composites/tracking.py
+++ b/dimod/reference/composites/tracking.py
@@ -29,7 +29,7 @@ except ImportError:
 
 from dimod.core.composite import ComposedSampler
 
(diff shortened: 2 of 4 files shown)


conta877:
whats the behavior when its sprinkled between composites? should there be a test for it?

arcondello:
The information is tracked on the composite object. So they don't interact with each other.

arcondello:
It does make things a bit awkward for nested composites, for instance
```
>>> sampler = RoofDualityComposite(TruncateComposite(TrackingComposite(ExactSolver())))
```
the inputs would need to be accessed with
```
>>> sampler.child.child.input
```
or, of course you can do
```
>>> tracking_sampler = TrackingComposite(ExactSolver())
>>> sampler = RoofDualityComposite(TruncateComposite(tracking_sampler))
>>> tracking_sampler.input
```

conta877:
yeah... so then why are we doing a composite for tracking instead of a sampler object? I thought the idea was to be able to nest

arcondello:
Because you still can access the nested objects, and because one is often interested in what happened between two composites.

conta877:
I feel like we are in agreement here. I just think that there should be a test for what

```
>>> tracking_sampler = TrackingComposite(ExactSolver())
>>> sampler = RoofDualityComposite(TruncateComposite(tracking_sampler))
>>> tracking_sampler.input
```

and this use case should be in docs

Issue #487, opened 2019-05-29
NullSampler/Trackercomposite won't work with composites that want a structured sampler


conta877:
Composites under dwave_system want structuredSampler as children making the nullsampler and the trackercomposite not work with them.

arcondello:
See also #111

PR #488, opened 2019-05-31
Add child_structure_dfs function

### LINKED ISSUE #111: SpinTransform should propagate Structured inheritance

arcondello:
This partly addresses #487, the EmbeddingComposite and similar will need to be updated to use `child_structure_dfs`.

See also #111

### DIFF
--- a/dimod/utilities.py
+++ b/dimod/utilities.py
@@ -370,17 +370,17 @@ def child_structure_dfs(sampler, seen=None):
     """
     seen = set() if seen is None else seen
 
-    if id(sampler) not in seen:
+    if sampler not in seen:
         try:
             return sampler.structure
         except AttributeError:
             # hasattr just tries to access anyway...
             pass
 
-    seen.add(id(sampler))
+    seen.add(sampler)
 
     for child in sampler.children:
-        if id(child) in seen:
+        if child in seen:
             continue
 
         try:
--- a/dimod/utilities.py
+++ b/dimod/utilities.py
@@ -20,7 +20,12 @@ import itertools
 
 from six import iteritems, itervalues
 
-__all__ = ['ising_energy', 'qubo_energy', 'ising_to_qubo', 'qubo_to_ising']
+__all__ = ['ising_energy',
+           'qubo_energy',
+           'ising_to_qubo',
+           'qubo_to_ising',
+           'child_structure_dfs',
+           ]
 
 
 def ising_energy(sample, h, J, offset=0.0):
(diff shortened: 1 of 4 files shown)


dimod/utilities.py
```diff
@@ -340,3 +345,48 @@ def resolve_label_conflict(mapping, old_labels=None, new_labels=None):
...
+
+    Returns:
+        :class:`~collections.namedtuple`: A named tuple of the form
+        `Structure(nodelist, edgelist, adjacency)`, where the 3-tuple values
+        are the :attr:`.Structured.nodelist`, :attr:`.Structured.edgelist`
+        and :attr:`.Structured.adjacency` attributes of the first structured
+        sampler encountered.
+
+    """
+    seen = set() if seen is None else seen
```
randomir: Why is `seen` needed for DFS? Do we want to support chains of composites with cyclic references (`a = A(B(a))`)? Those seem infinitely-evaluating, thus invalid by default.

dimod/utilities.py
```diff
@@ -340,3 +345,48 @@ def resolve_label_conflict(mapping, old_labels=None, new_labels=None):
...
+
+    if id(sampler) not in seen:
```
randomir: `id` not needed, you could just add `sampler` object to the set -- that's cheap, only reference is added. And membership testing is just as cheap, since sampler's hashes are equal to their `id`s already.

dimod/utilities.py
```diff
@@ -340,3 +345,48 @@ def resolve_label_conflict(mapping, old_labels=None, new_labels=None):
             # don't need to add it to intermediate_to_new because it is a self-label
 
     return old_to_intermediate, intermediate_to_new
+
+
+def child_structure_dfs(sampler, seen=None):
+    """Find the structure of a composed sampler by a depth first search on the
+    children.
```
randomir: What if there are multiple structured children? And one structure is closer to root, and makes more sense, but DFS returns some other structure from a deeper child?

dimod/utilities.py
arcondello: You're right, for some reason I thought it would not be hashable.

dimod/utilities.py
arcondello: There are lots of ways to resolve there being multiple structured children. It is up to the implementer to choose their algorithm. This one is very explicitly depth-first-search so they are opting into the behaviour.

dimod/utilities.py
arcondello: It would be possible to construct a set of samplers/composites for which you can get cycles, though I agree all of the cases I can think of for which that's true are arbitrary and would not be useful. But since it's possible, and keeping the set of samplers available is cheap, why not?

randomir: Well, it makes the code more complicated and it's uglier. :smile: 
But that's fine if we actually want to handle cyclic refs.

dimod/utilities.py
randomir: Sure, if each composite can choose the algorithm it uses, and this is only one strategy, all good.

PR #497, opened 2019-06-14
SampleSet/BQM.to_serializable now handles numpy variable labels

### LINKED ISSUE #496: sampleset.to_serializable to check variable indices
when using to_serializable, seems like it doesn't check the type of variable indices.
if np.int64 is sent it as variable indices, sampleset.to_serializable() remains not bson compatible. 

would be nice if we can check that and convert during the process

arcondello:
Closes #496 

I am genuinely a bit unsure about this one, I am worried that there are too many edge cases and it might be more explicit to simply require that the user provide serializable variable labels.

The argument for this approach is that hashing treats them as equivalent anyway so for instance
```
>>> {np.int64(1): 1, 1: 2, Fraction(1): 3}
{1: 3}
```
but it might be astonishing that serializing and deserializing does not return the same type for the variable labels.

### DIFF
--- a/dimod/variables.py
+++ b/dimod/variables.py
@@ -41,6 +41,23 @@ class CallableDict(abc.Callable, dict):
         return self[v]
 
 
+def iter_serialize_variables(variables):
+    # want to handle things like numpy numbers and fractions that do not
+    # serialize so easy
+    for v in variables:
+        if isinstance(v, Integral):
+            yield int(v)
+        elif isinstance(v, Number):
+            yield float(v)
+        elif isinstance(v, str):
+            yield v
+        # we want Collection, but that's not available in py2.7
+        elif isinstance(v, (abc.Sequence, abc.Set)):
+            yield tuple(iter_serialize_variables(v))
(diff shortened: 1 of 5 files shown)


dimod/variables.py
```diff
@@ -116,6 +120,25 @@ def count(self, v):
         # everything is unique
         return int(v in self)
 
+    def to_serializable(self):
+        def iter_variables(variables):
```
randomir: Maybe turn this into a utility function (de-dup from BQM too)?

arcondello: I consider the one in `Variables` the 'definitive' version and the BQM to be the duplication. In the future, I think the BQM will also use the Variables object (e.g. #477) so for now I have left the duplication.

However you are correct that in the mean time we could make this a utility function, probably in `bqm/serialization/utils.py`.

arcondello: I did make it a utility, but I kept in the `Variables` namespace so that we can remove it later without seriously affecting backwards compatibility.

PR #499, opened 2019-06-14
Bug fix energy calculation when offset is not a float/int

### LINKED ISSUE #498: When offset is mutable, energy calculation modifies the bqm
in energy calculation offset is assumed to be a float/int however an np.ndarray can also act as one. 
hence if such offset is passed, the bqm is constantly modified as the bqm.energy() is used.

conta877:
fixes #498


conta877:
thanks @pau557 for bug report!
i want cudos for good rebasing.

PR #506, opened 2019-07-09
dimod refresh part 1: restructure & intro update

JoelPasvolsky:


docs/introduction.rst
```diff
@@ -2,50 +2,34 @@
 Introduction
 ============
 
-The dimod API provides a binary quadratic :term:`model` (BQM) class that contains
-:term:`Ising` and quadratic unconstrained binary optimization (\ :term:`QUBO`\ ) models
-used by :term:`sampler`\ s such as the D-Wave system.
-It provides utilities for constructing new samplers and :term:`composed sampler`\ s.
-It also provides useful functionality for working with these models and samplers.
-Its reference examples include several samplers and composed samplers.
+The dimod API provides a binary quadratic model (BQM) class that contains
```
arcondello: I think here it shouldn't be "the dimod API", maybe just "dimod"

docs/introduction.rst
```diff
@@ -66,34 +50,52 @@ illustrative purposes, and converts the reduced-variables model to QUBO formulat
...
 
-
 Samplers and Composites
 -----------------------
 
-*Samplers* are processes that sample from low energy states of a problem’s objective function.
-A binary quadratic :term:`model` (BQM) :term:`sampler` samples from low energy states in
-models such as those defined by an Ising equation or a QUBO problem and returns an
-iterable of samples, in order of increasing energy. A dimod
+:term:`Sampler`\ s are  processes that sample from low-energy states of a problem’s objective function.
```
arcondello: The trailing backslashes raise a lot of errors for my (and assumedly the customer's) linter. It would be good if we could avoid them whenever possible.

docs/introduction.rst
JoelPasvolsky: Updated to dimod.

docs/introduction.rst
JoelPasvolsky: I do try to minimize use. Consider making a PR to improve the linter here: https://github.com/PyCQA/pylint/

PR #509, opened 2019-07-15
dimod refresh part 2: BQMs, sampleset, decorators, asserts

JoelPasvolsky:


dimod/binary_quadratic_model.py
```diff
@@ -120,12 +120,12 @@ class BinaryQuadraticModel(abc.Sized, abc.Container, abc.Iterable):
         >>> bqm = dimod.BinaryQuadraticModel({0: 1, 1: -1, 2: .5},
         ...                                  {(0, 1): .5, (1, 2): 1.5},
         ...                                  1.4,
-        ...                                  dimod.SPIN)
+        ...                                  dimod.Vartype.SPIN)
```
arcondello: why this change? SPIN and BINARY are in the top-level namespace

JoelPasvolsky: Yes, and I wanted to show that with the two consecutive examples using both supported formats rather than just the one

dimod/sampleset.py
```diff
@@ -646,7 +650,22 @@ def first(self):
 
     @property
     def info(self):
-        """Dict of information about the :class:`SampleSet` as a whole."""
+        """Dict of information about the :class:`SampleSet` as a whole.
+
+        Examples:
+           This example shows the type of information that might be returned by
```
arcondello: I am worried that this example is too specific. The sample set is used all over the place the info will not always be the same.

dimod/sampleset.py
```diff
@@ -1178,16 +1218,17 @@ def lowest(self, rtol=1.e-5, atol=1.e-8):
                           self.vartype)
 
     def truncate(self, n, sorted_by='energy'):
-        """Create a new SampleSet with rows truncated after n.
+        """Create a new sample set with up to n rows.
```
arcondello: I am worried that it is not clear that it takes the first n rows.

dimod/sampleset.py
```diff
@@ -1257,23 +1298,23 @@ def slice(self, *slice_args, **kwargs):
             8  0  0  0  0  0  0  0  0  1  0      8       1
             9  0  0  0  0  0  0  0  0  0  1      9       1
             ['BINARY', 10 rows, 10 samples, 10 variables]
-
+            >>>
```
arcondello: Why add the carrots?

dimod/utilities.py
```diff
@@ -350,28 +350,35 @@ def resolve_label_conflict(mapping, old_labels=None, new_labels=None):
...
+        sampler found.
 
     Raises:
         ValueError: If no structured sampler is found.
 
+    Examples:
+
+    >>> import dimod
+    >>> sampler = dimod.StructureComposite(dimod.ExactSolver(),[0, 1], [(0, 1)])
+    >>> print(dimod.child_structure_dfs(sampler).nodelist)
```
arcondello: Probably we want this example to show the structure being nested. Maybe ```sampler = dimod.TrackingComposite(dimod.StructureComposite(dimod.ExactSolver(), [0, 1], [(0, 1)])))```

dimod/sampleset.py
JoelPasvolsky: The intro sentence is pretty strongly worded to that effect: "type of information that might be returned"

dimod/sampleset.py
JoelPasvolsky: I feel that with both (1) the function name, "truncate" and (2) the following description, "Does not return  any rows above this limit in the original" that the error probability is sufficiently low. If you are highly worried, I can change

dimod/sampleset.py
JoelPasvolsky: Keeps the whole thing in one example in RTD, with a single scroll bar, as opposed to partial examples, each with a scroll bar. (And doctest broke when I had ellipses.)

dimod/utilities.py
JoelPasvolsky: Yes, thanks for this suggestion, I like it

dimod/sampleset.py
arcondello: Probably it would be better to just shorten the width then. E.g.
```
            >>> import numpy as np
            ...
            >>> sampleset = dimod.SampleSet.from_samples(np.diag(range(1, 11)),
            ...                                                                              dimod.BINARY, energy=range(10))
            >>> print(sampleset)
               0  1  2  3  4  5  6  7  8  9 energy num_oc.
            0  1  0  0  0  0  0  0  0  0  0      0       1
            1  0  1  0  0  0  0  0  0  0  0      1       1
            2  0  0  1  0  0  0  0  0  0  0      2       1
            3  0  0  0  1  0  0  0  0  0  0      3       1
            4  0  0  0  0  1  0  0  0  0  0      4       1
            5  0  0  0  0  0  1  0  0  0  0      5       1
            6  0  0  0  0  0  0  1  0  0  0      6       1
            7  0  0  0  0  0  0  0  1  0  0      7       1
            8  0  0  0  0  0  0  0  0  1  0      8       1
            9  0  0  0  0  0  0  0  0  0  1      9       1
            ['BINARY', 10 rows, 10 samples, 10 variables]
```

dimod/utilities.py
arcondello: Even better on multiple lines
```
sampler = dimod.TrackingComposite(
    dimod.StructureComposite(
        dimod.ExactSolver(), [0, 1], [(0, 1)])))
```

JoelPasvolsky: Yes, I had changed to 
```
>>> import dimod
    >>> sampler = dimod.TrackingComposite(dimod.StructureComposite(dimod.ExactSolver(),
    ...                                   [0, 1], [(0, 1)]))
 >>>
```

arcondello: That still exceeds the 80 char width by a fair margin (and introduces scroll bars)

dimod/sampleset.py
JoelPasvolsky: That will get rid of the scrolls but will keep the separation into partial examples

dimod/utilities.py
JoelPasvolsky: How about:
```
>>> sampler = dimod.TrackingComposite(dimod.StructureComposite(
...                 dimod.ExactSolver(), [0, 1], [(0, 1)]))
```

arcondello: I think I am attached to my example because it shows the 3-depth of the nesting clearly (I think). But your call.

JoelPasvolsky: Ah, I hadn't considered that aspect so I preferred two docstring lines to three but that is nice so will change.

dimod/sampleset.py
arcondello: I like the partial examples personally, but ymmv.

JoelPasvolsky: Our other examples are self-contained so I'd go with consistency barring some improvement to clarity. 
"ymmv": my daily new acronym

randomir: @JoelPasvolsky, I vote for no prompts between these examples. It's much nicer to have them self-contained. If you don't like the series of boxes on output, you can extract the comment line as a one line description of the example.

JoelPasvolsky: Okay, looks like we have mob rule in sampleset. I split it without the comment lines:
image

PR #510, opened 2019-07-15
BQM Serialization version 3.0.0

arcondello:
Remove the numpy dependency for the binary quadratic model serialization. The sample set will be similarly updated.

In order to make this more "python agnostic", perhaps we should change some of the metadata. We could do something like

```
{"schema_version": "3.0.0",
 "basetype": "BinaryQuadraticModel"
 "encoder": {"library": "dimod",  # "encoder" holds all the stuff specific to the encoding library
             "version": "0.8.14",
             "type": "BinaryQuadraticModel"  # the subclass, if any
             },
 ...
 }
```
We could alternatively consider just ditching that information. Though I do think it can be useful in the case of storing in a database for benchmarking etc.

### DIFF
--- a/dimod/binary_quadratic_model.py
+++ b/dimod/binary_quadratic_model.py
@@ -1760,19 +1760,22 @@ class BinaryQuadraticModel(abc.Sized, abc.Container, abc.Iterable):
 
         num_interactions = len(irow)
 
-        doc = {"basetype": "BinaryQuadraticModel",
-               "type": type(self).__name__,
-               "version": {"dimod": __version__,
-                           "bqm_schema": schema_version},
-               "variable_labels": variables,
-               "variable_type": self.vartype.name,
-               "info": self.info,
-               "offset": float(offset),
-               "index_type": np.dtype(index_dtype).str,
-               "bias_type": np.dtype(bias_dtype).str,
-               "shape": [num_variables, num_interactions],
-               "use_bytes": bool(use_bytes)
-               }
+        doc = {
+            # metadata
+            "type": type(self).__name__,
+            "version": {"bqm_schema": schema_version},
+            "use_bytes": bool(use_bytes),
+            "index_type": np.dtype(index_dtype).str,
+            "bias_type": np.dtype(bias_dtype).str,
+
+            # bqm
+            "num_variables": num_variables,
+            "num_interactions": num_interactions,
+            "variable_labels": variables,
+            "variable_type": self.vartype.name,
+            "offset": float(offset),
+            "info": self.info,
+            }
 
         if use_bytes:
             # these are vectors so don't need to specify byte-order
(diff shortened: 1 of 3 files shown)


oneklc (approved):
nice!

arcondello:
I removed some of the old fields and moved them around. Current scheme:
```
            {'bias_type': '<f4',
             'index_type': '<u2',
             'info': {},
             'linear_biases': b'\x00\x00\x80\xbf\x00\x00\x80?\x00\x00@@\x00\x00\x00\x00'
                              b'\x00\x00\x00\x00\x00\x00\x00\x00',
             'num_interactions': 3,
             'num_variables': 6,
             'offset': 3.0,
             'quadratic_biases': b'\x00\x00\x80?\x00\x00@@\x00\x00\x80\xbf',
             'quadratic_head': b'\x00\x00\x03\x00\x00\x00',
             'quadratic_tail': b'\x03\x00\x04\x00\x05\x00',
             'type': 'BinaryQuadraticModel',
             'use_bytes': True,
             'variable_labels': ['a', 4, ('a', 'complex key'), 'c', 'b', 3],
             'variable_type': 'SPIN',
             'version': {'bqm_schema': '3.0.0'}}
```

randomir (approved):
Looks good to me.

I would just be explicit (in the docs) about the fact we are actually using NumPy for serialization, and that out type definition is actually NumPy's `dtype` (the latter point being more important -- because it's similar, but quite different from Python's standard library format in `struct`/`array`).

arcondello:
Numpy uses the python buffer protocol syntax for types as defined in [struct](https://docs.python.org/3.7/library/struct.html#format-characters) and expanded in [PEP3118](https://www.python.org/dev/peps/pep-3118/#additions-to-the-struct-string-syntax). Agree it should be called out explicitly.

arcondello:
Hmm,there may be some differences, I may have misread the [numpy docs](https://docs.scipy.org/doc/numpy/reference/arrays.interface.html#the-array-interface). Anyway, I will look further into it and make is clear in the docs.

randomir:
Yes, that's the similar part, but for "quite different", check [`array`](https://docs.python.org/3.7/library/array.html#module-array)'s `'u'` -- `Py_UNICODE` -- not `unsigned int`.

arcondello:
Right, and of course array and struct don't match either (though they are at least they appear consistent in their overlap).

This leads to maybe we should use some other format for types that is not python-specific. Maybe use numpy's "nice" printing? e.g. 
```
>>> np.dtype(int).name
'int64'
```
it seems pretty common, but I cannot find a standard defined anywhere.

randomir:
Yes, I would prefer `dtype.name`. The output for types we'll use seems perfect: `int16`/`uint32`/`float64`...

Re: the standard -- this looks like the "N competing standards, let's create a new one to unite them all" xkcd situation. :smile:

PR #512, opened 2019-07-22
Recursive info field serialization

arcondello:
Created a more generic serialization for numpy arrays. This allows us to serialize the info field.

Slightly alters the serialization introduced in #511

### DIFF
--- a/dimod/sampleset.py
+++ b/dimod/sampleset.py
@@ -33,8 +33,12 @@ from numpy.lib import recfunctions
 from dimod.decorators import vartype_argument
 from dimod.exceptions import WriteableError
 from dimod.serialization.format import Formatter
-from dimod.serialization.utils import pack_samples, unpack_samples
-from dimod.serialization.utils import serialize_ndarray, deserialize_ndarray
+from dimod.serialization.utils import (pack_samples,
+                                       unpack_samples,
+                                       serialize_ndarray,
+                                       deserialize_ndarray,
+                                       dfs_serialize_ndarray,
+                                       dfs_deserialize_ndarray)
 from dimod.utilities import LockableDict
 from dimod.variables import Variables
 from dimod.vartypes import Vartype
@@ -1422,7 +1426,8 @@ class SampleSet(abc.Iterable, abc.Sized):
             # other
             "variable_labels": self.variables.to_serializable(),
             "variable_type": self.vartype.name,
-            "info": self.info,
+            "info": dfs_serialize_ndarray(self.info, use_bytes=use_bytes,
+                                          bytes_type=bytes_type),
             }
 
     def _asdict(self):
@@ -1513,7 +1518,7 @@ class SampleSet(abc.Iterable, abc.Sized):
         num_variables = obj['num_variables']
         variables = [tuple(v) if isinstance(v, list) else v
                      for v in obj["variable_labels"]]
-        info = obj['info']
+        info = dfs_deserialize_ndarray(obj['info'])
 
         # vectors
         vectors = {name: deserialize_ndarray(data)
--- a/dimod/sampleset.py
+++ b/dimod/sampleset.py
@@ -1392,35 +1392,37 @@ class SampleSet(abc.Iterable, abc.Sized):
         # bytes we could avoid a copy when undoing the serialization. However,
         # we want to pack the samples, so that means that we're storing the
         # arrays individually.
-        sample_data = {}
-        for name in self.record.dtype.names:
+        vectors = {name: serialize_ndarray(data, use_bytes=use_bytes,
+                                           bytes_type=bytes_type)
+                   for name, data in self.data_vectors.items()}
 
-            data = self.record[name]
+        if self.vartype is Vartype.SPIN:
+            packed = pack_samples(self.record.sample > 0)
+        else:
+            packed = pack_samples(self.record.sample)
 
-            if name == 'sample':
-                if self.vartype is Vartype.SPIN:
-                    data = data > 0
-
-                data = pack_samples(data)
-
-            sample_data[name] = serialize_ndarray(data,
-                                                  use_bytes=use_bytes,
-                                                  bytes_type=bytes_type)
+        sample_data = serialize_ndarray(packed,
+                                        use_bytes=use_bytes,
+                                        bytes_type=bytes_type)
 
         return {
             # metadata
             "type": type(self).__name__,
             "version": {"sampleset_schema": schema_version},
-            "use_bytes": bool(use_bytes),
-            "sample_type": self.record.sample.dtype.name,
 
-            # sampleset
+            # samples
             "num_variables": len(self.variables),
             "num_rows": len(self),
+            "sample_data": sample_data,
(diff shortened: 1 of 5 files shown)


dimod/sampleset.py
```diff
@@ -1391,39 +1396,38 @@ def to_serializable(self, use_bytes=False, bytes_type=bytes):
...
             "num_variables": len(self.variables),
             "num_rows": len(self),
+            "sample_data": sample_data,
+            "sample_type": self.record.sample.dtype.name,
+
+            # vectors
+            "vectors": vectors,
+
+            # other
             "variable_labels": self.variables.to_serializable(),
```
randomir: Do we need to handle NumPy scalars/primitives for variables here (e.g. `int32`)?

dimod/serialization/utils.py
```diff
@@ -13,9 +13,88 @@
...
+    arr = arr.reshape(obj['shape'])  # makes a view generally, but that's fine
+    return arr
+
+
+def dfs_serialize_ndarray(obj, use_bytes=False, bytes_type=bytes):
+    """Looks through the object, serializing numpy arrays."""
+    if isinstance(obj, np.ndarray):
+        return serialize_ndarray(obj, use_bytes=use_bytes, bytes_type=bytes_type)
+    elif isinstance(obj, abc.Mapping):
+        return {key: dfs_serialize_ndarray(val) for key, val in obj.items()}
```
randomir: Keys can be NumPy types too.

dimod/serialization/utils.py
```diff
@@ -13,9 +13,88 @@
...
+    """
+    if obj['use_bytes']:
+        arr = np.frombuffer(obj['data'], dtype=obj['data_type'])
+    else:
+        arr = np.asarray(obj['data'], dtype=obj['data_type'])
+    arr = arr.reshape(obj['shape'])  # makes a view generally, but that's fine
+    return arr
+
+
+def dfs_serialize_ndarray(obj, use_bytes=False, bytes_type=bytes):
```
randomir: Confusing name since also dicts and lists are serialized.

dimod/serialization/utils.py
```diff
@@ -13,9 +13,88 @@
...
+    """Looks through the object, serializing numpy arrays."""
+    if isinstance(obj, np.ndarray):
+        return serialize_ndarray(obj, use_bytes=use_bytes, bytes_type=bytes_type)
+    elif isinstance(obj, abc.Mapping):
+        return {key: dfs_serialize_ndarray(val) for key, val in obj.items()}
+    elif isinstance(obj, abc.Sequence) and not isinstance(obj, str):
+        return list(map(dfs_serialize_ndarray, obj))
```
randomir: List comprehensions are faster when you're not exploiting iter `map`.

dimod/serialization/utils.py
```diff
@@ -13,9 +13,88 @@
...
+    return arr
+
+
+def dfs_serialize_ndarray(obj, use_bytes=False, bytes_type=bytes):
+    """Looks through the object, serializing numpy arrays."""
+    if isinstance(obj, np.ndarray):
+        return serialize_ndarray(obj, use_bytes=use_bytes, bytes_type=bytes_type)
+    elif isinstance(obj, abc.Mapping):
+        return {key: dfs_serialize_ndarray(val) for key, val in obj.items()}
+    elif isinstance(obj, abc.Sequence) and not isinstance(obj, str):
```
randomir: Soon won't matter, but this won't handle unicode in Python 2. (you can use six.string_types, or something like that)

dimod/serialization/utils.py
```diff
@@ -13,9 +13,88 @@
...
+        return list(map(dfs_serialize_ndarray, obj))
+    return obj
+
+
+def dfs_deserialize_ndarray(obj):
+    """Inverse of dfs_serialize_ndarray"""
+    if isinstance(obj, abc.Mapping):
+        if obj.get('type', '') == 'array':
```
randomir: Out of curiosity (I've seen you used this pattern elsewhere) -- why use empty string for default (and not the default `None`)?

randomir (commented):
Looks good.

I would just consider adding ints and floats to the serialization mix.

dimod/serialization/utils.py
arcondello: Not if it's JSON serializable. Unless numpy chars?

dimod/serialization/utils.py
arcondello: no real reason, matching type but None would work just as well.

dimod/sampleset.py
arcondello: Good question, definitely an edge case we should test/handle.

arcondello: Handled by the Variables object https://github.com/dwavesystems/dimod/blob/368c8887948e62528a8369ea73744e2f584582d7/dimod/variables.py#L44

dimod/serialization/utils.py
randomir: Also numpy numbers.

dimod/serialization/utils.py
randomir: If not even better! :smiley_cat:

dimod/serialization/utils.py
arcondello: JSON needs string keys for objects. I don't think we should handle that case here.

arcondello: Also seems to work for the numeric types out of the box
```
In [4]: json.dumps({np.float(0): 1})                                                                                                   
Out[4]: '{"0.0": 1}'
```

oneklc (approved):

dimod/serialization/utils.py
randomir: Yes, I know JSON requires string keys. But as you see, built-in json package serializes python numbers as strings. It fails with numpy though:
```
>>> json.dumps({np.int64(0): 1})
...
TypeError: keys must be a string
```

randomir: I'm not saying I like casting arrays to strings, but this solution handles only ndarrays in values. I feel if we already are providing this service to users (serialization of numpy types), we should be little broader.

randomir (approved):

dimod/serialization/utils.py
arcondello: It's a simple enough change so will do. However, I don't think it is in scope for dimod to handle serialization of numpy objects in general. So far we are only providing it as an addition to the serialization of dimod objects. If we want to go beyond that this code will need to be a lot more sophisticated.

randomir: I agree. Actually, if you remember, I argued even more than that -- BQM/SampleSet serialization (for SAPI encoding) does not belong in dimod at all.

arcondello: I think serialization of dimod objects makes sense in dimod. The numpy objects that happen to be stored less so.

randomir: (But as long as we have it already, and we're now adding NumPy support too, I'm saying let's just cover simple edge cases to minimize gotcha moments.)

PR #515, opened 2019-08-02
C++ BQM

arcondello:
This is a partial implementation of a cythonized C++ binary quadratic model. I want to make a PR to a working branch so that I can start making feature pull requests in chunks, there will be a final review before it is merged to master.


arcondello:
It seems that some of the c++ features I am using (alias templates) are not supported in visual studio 10 which is the one that appveyor comes with. Not to mention that different python versions have different opinions about c++ versions. I will make a different PR to deal with the windows issues (see also https://github.com/dwavesystems/dwave-tabu/pull/15).

PR #516, opened 2019-08-02
Roof duality error

arcondello:
@wbernoudy reported a bug.When sampling mode is not on, it seems to fix a variables multiple times. I have added a failing test but have not yet fixed the bug. The following is the contents of `ret.fixedVars` before it fails:
```
2 1
4 1
5 1
7 1
10 1
12 1
14 1
15 1
17 1
19 1
20 1
21 1
25 1
27 1
0 0
1 0
3 0
6 0
8 0
9 0
10 0
11 0
13 0
14 0
15 0
16 0
18 0
22 0
23 0
24 0
25 0
26 0
27 0
```

The code under roof_duality is also due for a major refresh (it was imported from the legacy sapi-client).


arcondello:
Worth noting that scaling the problem by .0000000001 fixes the bug.

arcondello:
This would be a relatively large fix for a problem that only shows up with extremely large biases. Closing for now

PR #517, opened 2019-08-02
Refactor appveyor config

arcondello:
Using the scripts from https://github.com/dwavesystems/dwave-tabu/pull/15 and https://github.com/dwavesystems/dwave-tabu/pull/13, try to fix the build on windows.

Still need to change the actual c++ code to be compliant.


arcondello:
Python3.4 is still having issues so skipping - we are planning to deprecate support for 3.4 anyway.

PR #518, opened 2019-08-02
Minor changes to cython for roof duality

arcondello:
The exception changes from #516


randomir:
There’s also an option of subclassing std exceptions...

arcondello:
Indeed, which I would do for the other file `compressed_matrix.hpp` in this case I couldn't see any reason to have custom exceptions.

PR #523, opened 2019-08-26
Improve performance on AdjMapBQM

arcondello:


dimod/bqm/adjmapbqm.pyx
```diff
@@ -125,69 +141,175 @@ cdef class AdjMapBQM:
...
+
+        Returns:
+            hashable: The last variable added to the binary quadratic model.
+
+        Raises:
+            ValueError: If the binary quadratic model is empty.
+
+        """
         if num_variables(self.adj_) == 0:
             raise ValueError("pop from empty binary quadratic model")
```
wbernoudy: Might want to raise an `IndexError` here to be consistent with other builtin `.pop()` behavior

arcondello: In this case the variables are index-labelled, so I am not sure if `IndexError` makes sense? Though neither does `ValueError`...

wbernoudy: Agreed, I just noticed that's what happens with `list`:
```
[ins] In [6]: [].pop()                                                                                                                                                                                                                                                                                                                                                                         
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-6-47d711ddf6f1> in <module>
----> 1 [].pop()

IndexError: pop from empty list
```

arcondello: Yes, though in that case there is an optional parameter for `pop` that you could conceivably be raising an index error for. Anyway, can be fixed later

PR #526, opened 2019-08-30
Add AdjVectorBQM

arcondello:
Includes the commits from and therefore closes #523

### DIFF
--- a/dimod/bqm/adjvectorbqm.pxd
+++ b/dimod/bqm/adjvectorbqm.pxd
@@ -1,0 +1,42 @@
+# distutils: language = c++
+# cython: language_level=3
+#
+# Copyright 2019 D-Wave Systems Inc.
+#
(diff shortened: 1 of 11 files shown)


dimod/bqm/adjmapbqm.pyx
```diff
@@ -125,69 +141,175 @@ cdef class AdjMapBQM:
...
+            yield self._idx_to_label.get(v, v)
 
     def pop_variable(self):
+        """Remove a variable from the binary quadratic model.
+
+        Returns:
+            hashable: The last variable added to the binary quadratic model.
+
+        Raises:
+            ValueError: If the binary quadratic model is empty.
```
arcondello: As commented on in https://github.com/dwavesystems/dimod/pull/523#discussion_r319322082, raising a `ValueError` here might not make sense?

PR #532, opened 2019-09-18
Add ABC for cyBQMs and reference dict implementation

arcondello:
This is still obviously very much a work in progress - documentation and tests are incomplete.

### DIFF
--- a/dimod/bqm/adjdictbqm.py
+++ b/dimod/bqm/adjdictbqm.py
@@ -29,6 +29,8 @@ __all__ = ['AdjDictBQM']
 
 
 class AdjDictBQM(ShapeableBQM):
+    """
+    """
     def __init__(self, obj=0):
         # we could actually have these variable but to keep this consistent
         # with the other BQMs we fix them for now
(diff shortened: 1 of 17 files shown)


tests/test_bqm.py
```diff
@@ -23,18 +23,28 @@
...
             self.assertEqual(bqm.get_quadratic(u, v), 2)  # added
         for u in range(5):
             self.assertEqual(bqm.get_linear(u), 1)
 
+    # def test_construction_nans(self):
+    #     bqm = self.BQM(np.nan((5, 5)))
+    #     self.assertEqual(bqm.shape, (5, 0))
+    #     for u in range(5):
+    #         self.assertEqual(bqm.get_linear(u), 0)
+
```
wbernoudy: My opinion on how to treat nans is to raise an error. There isn't an obvious way to interpret them and no established convention.

dimod/bqm/adjdictbqm.py
```diff
@@ -0,0 +1,301 @@
...
+        """
+        try:
+            self._adj[u].pop(v)
+        except KeyError:
+            # nothing pop
+            return False
+
+        # want to raise an exception in the case that they got out of sync
+        # as a sanity check
+        self._adj[v].pop(u)
```
wbernoudy: How come you don't need to `self._adj[u].pop(v)`?

arcondello: Happens above in the try-catch. We only get to this line if that one has occured

PR #536, opened 2019-09-30
Add offset and vartype properties to cybqm

arcondello:
Construction now looks like
```
bqm = BQM(dimod.BINARY)  # empty
bqm = BQM(np.ones((5, 5)), dimod.BINARY)
bqm = BQM((h, J), dimod.SPIN)
bqm = BQM((h, J, offset), dimod.SPIN)
bqm = BQM(5, dimod.BINARY)  # 5 variables, 0 interactions
```

### DIFF
--- a/dimod/bqm/adjarraybqm.pyx
+++ b/dimod/bqm/adjarraybqm.pyx
@@ -88,25 +88,24 @@ cdef class cyAdjArrayBQM:
 
     This can be instantiated in several ways:
 
-        AdjVectorBQM()
+        AdjArrayBQM(vartype)
             Creates an empty binary quadratic model
 
-        AdjVectorBQM(n)
-            Where n is the number of nodes.
-
-        AdjVectorBQM((linear, [quadratic, [offset]]))
+        AdjArrayBQM((linear, [quadratic, [offset]]))
             Where linear, quadratic are:
                 dict[int, bias]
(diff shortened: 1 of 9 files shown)


arcondello:
In the future we will also support
```
bqm = BQM(another_bqm)
```

dimod/bqm/adjarraybqm.pyx
```diff
@@ -120,30 +120,40 @@ cdef class cyAdjArrayBQM:
...
 
-        if isinstance(arg1, Integral):
-            self.invars_.resize(arg1)
-        elif isinstance(arg1, tuple):
-            self.__init__(cyAdjMapBQM(arg1))  # via the map version
-        elif hasattr(arg1, "to_adjarray"):
+        if isinstance(obj, Integral):
+            self.invars_.resize(obj)
+        elif isinstance(obj, tuple):
+            self.__init__(cyAdjMapBQM(obj, vartype))  # via the map version
```
randomir: This might work in the current implementation, but I don't see why you would generally want to do this.

arcondello: Anecdotally it seems faster to cast it to the map/vector version of cyBQM then copy (in C) to the array version rather than try to figure out the sizes/degrees/labels needed in python space. Needs more testing though

PR #539, opened 2019-10-03
Add ConnectedComponentComposite

hsadeghidw:


dimod/reference/composites/connectedcomponent.py
```diff
@@ -0,0 +1,104 @@
...
+#
+# =============================================================================
+"""
+A composite that break the problem into sub-problems corresponding to the connected components of the binary
+quadratic model graph before sending to its child sampler.
+"""
+
+from dimod.core.composite import ComposedSampler
+from dimod.sampleset import SampleSet
+import networkx as nx
```
arcondello: One problem is that dimod actually does not have a dependency on networkx, it's optional. Because bfs/dfs is pretty easy to implement, it might be better to not go through networkx to find the connected components. I could add a dimod version of `nx.connected_components` pretty easily.

dimod/reference/composites/connectedcomponent.py
```diff
@@ -0,0 +1,104 @@
...
+        Returns:
+            :obj:`dimod.SampleSet`
+
+        """
+
+        # solve the problem on the child system
+        child = self.child
+        variables = bqm.variables
+        if components is None:
+            components = list(nx.connected_components(bqm.to_networkx_graph()))
```
arcondello: See above.

dimod/reference/composites/connectedcomponent.py
```diff
@@ -0,0 +1,104 @@
...
+
+    Fixes variables of a bqm and modifies linear and quadratic terms
+    accordingly. Returned samples include the fixed variable
+
+    Args:
+       sampler (:obj:`dimod.Sampler`):
+            A dimod sampler
+
+    Examples:
+       This example uses :class:`.FixedVariableComposite` to instantiate a
```
arcondello: This still references `FixedVariableComposite`

dimod/reference/composites/connectedcomponent.py
```diff
@@ -0,0 +1,104 @@
...
+from dimod.sampleset import SampleSet
+import networkx as nx
+
+__all__ = ['ConnectedComponentsComposite']
+
+
+class ConnectedComponentsComposite(ComposedSampler):
+    """Composite to fix variables of a problem to provided.
+
+    Fixes variables of a bqm and modifies linear and quadratic terms
```
arcondello: This still references `FixedVariableComposite`

dimod/reference/composites/connectedcomponent.py
hsadeghidw: Once I have `dimod.connected_components`, I can modify the code accordingly.

dimod/reference/composites/connectedcomponent.py
```diff
@@ -0,0 +1,106 @@
...
+            :obj:`dimod.SampleSet`
+
+        """
+
+        # solve the problem on the child system
+        child = self.child
+        variables = bqm.variables
+        if components is None:
+            components = list(connected_components(bqm))
+        if type(components) is set:
```
arcondello: better to use `isinstance(components, set)` to handle subclasses etc. You might just want to ditch this way of providing components alltogether and make the user provide `[components]` themselves, or update the docstring accordingly.

dimod/reference/composites/connectedcomponent.py
```diff
@@ -0,0 +1,106 @@
...
+                Binary quadratic model to be sampled from.
+
+            components (list(set)):
+                A list of disjoint set of variables that fully partition the variables
+
+            **parameters:
+                Parameters for the sampling method, specified by the child sampler.
+
+        Returns:
+            :obj:`dimod.SampleSet`
```
arcondello: This should document the fact that it's only returning a single state.

dimod/reference/composites/connectedcomponent.py
hsadeghidw: now using isinstance

dimod/reference/composites/connectedcomponent.py
hsadeghidw: It's now reflected in the document that only a single sample is returned.
Also added comment in the body of the code for future developers.

PR #540, opened 2019-10-04
Generator functions now accept edgelists as graphs

arcondello:


tests/test_decorators.py
```diff
@@ -147,6 +147,50 @@ def f(G):
...
+        @graph_argument('G')
+        def f(G):
+            return G
+
+        edgelist = [(0, 1), (2, 3), (0, 2)]
+
+        nodes, edges = f(edgelist)
+
+        for edge in edges:
+            self.assertIn(edge, edgelist)
```
m3ller: Shouldn't you also verify that when using simply the `edgelist` that the nodes are generated correctly as well?

tests/test_decorators.py
```diff
@@ -147,6 +147,50 @@ def f(G):
...
+        @graph_argument('G')
+        def f(G):
+            return G
+
+        edgelist = [(0, 1), (2, 3)]
+
+        nodes, edges = f(edgelist)
+
+        for edge in edges:
+            self.assertIn(edge, edgelist)
```
m3ller: ditto

tests/test_decorators.py
arcondello: Oops, good catch! should be `self.assertIn(edge, edges)`

arcondello: Yes probably, will add

PR #541, opened 2019-10-05
Add clip composite

hsadeghidw:


dimod/reference/composites/clipcomposite.py
```diff
@@ -0,0 +1,154 @@
...
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+# =============================================================================
+"""
+A composite that clips problem variables below and above threshold. if lower and upper bounds is not given
+it does nothing.
+
+"""
+try:
```
arcondello: Looks like you don't need this import

dimod/reference/composites/clipcomposite.py
```diff
@@ -0,0 +1,154 @@
...
+    import collections.abc as abc
+except ImportError:
+    import collections as abc
+
+
+from dimod.binary_quadratic_model import BinaryQuadraticModel
+from dimod.core.composite import ComposedSampler
+from dimod.sampleset import SampleSet
+
+__all__ = 'ClipComposite',
```
arcondello: @randomir has given me a hard time for this before :smile: preferred pattern is `__all__ = ['ClipComposite']`

dimod/reference/composites/clipcomposite.py
```diff
@@ -0,0 +1,154 @@
...
+    accordingly.
+
+    Args:
+       sampler (:obj:`dimod.Sampler`):
+            A dimod sampler
+
+    Examples:
+       This example uses :class:`.ClipComposite` to instantiate a
+       composed sampler that submits a simple Ising problem to a sampler.
+       The composed sampler scales linear, quadratic biases and offset as
```
arcondello: I don't think you mean 'scales' here?

dimod/reference/composites/clipcomposite.py
```diff
@@ -0,0 +1,154 @@
...
+        return param
+
+    @property
+    def properties(self):
+        return {'child_properties': self.child.properties.copy()}
+
+    def sample(self, bqm, lower_bound=None, upper_bound=None, **parameters):
+        """ Clip and sample from the provided binary quadratic model.
+
+        if lower_bound and upper_bound are given variables with value above upper_bound or below lower_bound are clipped.
```
arcondello: Trying to < 80 char width

dimod/reference/composites/clipcomposite.py
```diff
@@ -0,0 +1,154 @@
...
+        param = self.child.parameters.copy()
+        param.update({'lower_bound': [], 'upper_bound': []})
+        return param
+
+    @property
+    def properties(self):
+        return {'child_properties': self.child.properties.copy()}
+
+    def sample(self, bqm, lower_bound=None, upper_bound=None, **parameters):
+        """ Clip and sample from the provided binary quadratic model.
```
arcondello: Super picky but space between `"""` and `Clip`

dimod/reference/composites/clipcomposite.py
```diff
@@ -0,0 +1,154 @@
...
+        Args:
+            bqm (:obj:`dimod.BinaryQuadraticModel`):
+                Binary quadratic model to be sampled from.
+
+            lower_bound (number):
+                Value by which to clip the variables from below.
+
+            upper_bound (number):
+                Value by which to clip the variables from above.
+            **parameters:
```
arcondello: Missing a newline before `**parameters` which will cause issues for sphinx

dimod/reference/composites/clipcomposite.py
```diff
@@ -0,0 +1,154 @@
...
+            :obj:`dimod.SampleSet`
+
+        """
+        child = self.child
+        bqm_copy = _clip_bqm(bqm, lower_bound, upper_bound)
+        response = child.sample(bqm_copy, **parameters)
+
+        return SampleSet.from_samples_bqm(response.record.sample, bqm)
+
+    def sample_ising(self, h, J, offset=0.0, lower_bound=None, upper_bound=None, **parameters):
```
arcondello: This is handled by the abstract base class, think you can get rid of `sample_ising` alltogether.

dimod/reference/composites/clipcomposite.py
```diff
@@ -0,0 +1,154 @@
...
+
+        Returns:
+            :obj:`dimod.SampleSet`
+
+        """
+        child = self.child
+        bqm_copy = _clip_bqm(bqm, lower_bound, upper_bound)
+        response = child.sample(bqm_copy, **parameters)
+
+        return SampleSet.from_samples_bqm(response.record.sample, bqm)
```
arcondello: Should be able to just pass in the `response`, which will cover the variable labels case.

dimod/reference/composites/clipcomposite.py
hsadeghidw: side effects of copy-pasting another composite :laughing:

Issue #410, opened 2019-02-15
Refactor Exact Solver


m3ller:
**Application**
Refactor ExactSolver so that it can more quickly calculate the energy delta for the single bit flip.

PR #542, opened 2019-10-07
Refactor ExactSolver

hsadeghidw:


dimod/reference/samplers/exact_solver.py
```diff
@@ -86,46 +84,18 @@ def sample(self, bqm):
...
-
-                yield sample.copy(), float(energy) + off
-
-        samples, energies = zip(*iter_samples())
-
-        response = SampleSet.from_samples(np.array(samples, dtype='int8'), Vartype.BINARY, energies)
-
-        # make sure the response matches the given vartype, in-place.
+        bqm_copy = bqm.copy()
+        bqm_copy = bqm_copy.change_vartype(Vartype.BINARY)
```
randomir: Instead of manually copying, you can use `inplace=False` argument of `change_vartype`.

dimod/reference/samplers/exact_solver.py
```diff
@@ -86,46 +84,18 @@ def sample(self, bqm):
...
+        samples = list(itertools.product([0, 1], repeat=n))
+        response = SampleSet.from_samples_bqm(samples, bqm_copy)
```
randomir: Will bqms with non-sequential-integer variable names be handled?

hsadeghidw: Do we have test cases for the non-sequential-integer variable? If yes, it passes, otherwise, give me an example to test and add.

arcondello: It won't, unless you do
```
response = SampleSet.from_samples_bqm((samples, list(bqm_copy)), bqm_copy)
```

arcondello:
```
def test_arbitrary_labels(self):
    bqm = dimod.BQM.from_ising({}, {'ab': -1}
    sampleset = dimod.ExactSolver().sample(bqm)
    self.assertEqual(set(sampleset.variables), set(bqm))
```

dimod/reference/samplers/exact_solver.py
arcondello: You can probably get around the copy alltogether with
```
samples = itertools.product(bqm.vartype.value, repeat=n)
```

hsadeghidw: this one failed some tests

dimod/reference/samplers/exact_solver.py
hsadeghidw: This test is passed.

arcondello: Ah, because if you send in a spin-valued BQM and it returns a binary sampleset it will fail

dimod/reference/samplers/exact_solver.py
arcondello: Because you're returning a BINARY sampleset for a SPIN bqm I think

dimod/reference/samplers/exact_solver.py
```diff
@@ -86,46 +84,17 @@ def sample(self, bqm):
...
-        # make sure the response matches the given vartype, in-place.
+        bqm_copy = bqm.change_vartype(Vartype.BINARY, inplace=False)
+        samples = list(itertools.product([0, 1], repeat=n))
+        response = SampleSet.from_samples_bqm(samples, bqm_copy)
         response.change_vartype(bqm.vartype, inplace=True)
-
         return response
 
 
 def _ffs(x):
```
arcondello: You can remove this function now as well I think

dimod/reference/samplers/exact_solver.py
arcondello: Oh, I see why it's passing, because it's using the `bqm_index_labels` decorator. In the interests of readability and performance, I would remove the decorator and use the `response = SampleSet.from_samples_bqm((samples, list(bqm_copy)), bqm_copy)` syntax

arcondello:
Failures are now caused by other bugs, will fix

hsadeghidw:
Fixed #410

PR #544, opened 2019-10-07
Return numpy scalars from cybqm objects

arcondello:


dimod/bqm/utils.pyx
```diff
@@ -0,0 +1,30 @@
...
+    # https://github.com/numpy/numpy/blob/5ce770ae3de63861c768229573397cadd052f712/numpy/core/src/multiarray/scalarapi.c#L617
+    # seems to contradict that. A cursory read of the code seems to support the
+    # latter conclusion, so for now we'll assume this is safe to use.
+    object PyArray_Scalar(void* data, np.dtype dtype, object itemsize)
+
+np.import_array()
+
+cdef object as_numpy_scalar(double a, np.dtype dtype):
+    """Note that the memory is interpreted to match dtype, not a cast"""
+    return PyArray_Scalar(&a, dtype, None)
```
wbernoudy: What does the `None` mean here?

arcondello: I _think_ that's the size (in bytes) of the input object, but it seems to be ignored and instead read from the dtype.

PR #548, opened 2019-10-10
Feature/cut variables

### LINKED ISSUE #547: CutVariableComposite
Sometimes a variable is a bridge between two graphs. It would be good to have a composite that can apply to such problems. In such cases, it's possible to remove the cut variables and effectively cut the graph into pieces. Each piece can be solved and then combined with two possible values of the cut variables.

For each cut variable, the problem breaks into three components: [left_graph, cut_variable, right_graph]
The above list can be passed as an argument to ConnectedComponentComposite.

However, the composite combines the best of each. since cut_variable is not really a disjoint connected component, it needs to be handled differently. So, maybe a standalone composite is better.

aidanproy:
Adds CutVertexComposite, which decomposes a connected bqm into biconnected components, solves those components, and merges the solution.

Currently the composite returns a single solution, like the ConnectedComponentsComposite. Eventually it should have the option of sampling.

Closes #547


dimod/reference/composites/cutvertex.py
```diff
@@ -0,0 +1,281 @@
...
+"""
+A composite that breaks the problem into sub-problems corresponding to the
+biconnected components of the binary quadratic model graph before sending to its child sampler.
+"""
+
+from dimod.sampleset import as_samples
+from dimod.core.composite import ComposedSampler
+from dimod.sampleset import SampleSet
+import dimod
+import networkx as nx
```
arcondello: See discussion in https://github.com/dwavesystems/dimod/pull/539#discussion_r332116231. However, in this case the functions are potentially more involved, so rewriting might be more trouble than it's worth.

aidanproy: The nx.articulation_points and nx.biconnected_components aren't too difficult actually, and pure python. It's basically just DFS. So if you'd rather have dimod.articulation_points and dimod.biconnected_components we could do that (although it does seem like a lot of code replication).

arcondello: Yeah exactly, https://github.com/dwavesystems/dimod/commit/e5dd5f39c374209b1a7c843c5f42a3687fa563d2 added bfs so dfs as well is not such a stretch, but at some point we end up just re-implementing NetworkX and there's no point in that. On the other hand, casting BQMs to NetworkX graphs has a non-trivial copy cost so working natively on BQMs can be nice :man_shrugging:.

jberwald: IMHO, we should try to use nx's function until it's clear there is a drastic slowdown because of copy costs or pure Python. 

Glad to learn about `nx.articulation_points`. I somehow missed the `articulation_points` function last summer when I was working on this problem. That would have simplified things. 👍

arcondello: I tend to agree. It's just frustrating because the BQM and networkx graph are so close in API, for a lot of nx functions the only thing blocking them working on the BQM directly is a decorator checking whether it's a directed graph or not. We could potentially wrap the BQM with some attributes to get around that but I am sure there would be edge cases.

jberwald: BTW, I've had great results using Metis in the past to decompose graphs. IIRC, it has a Python interface, but would add another requirement (including C backend 👎 )

aidanproy: pymetis, I think. Yeah metis is great (although overkill if you just want cut vertices rather a partition of an arbitrary graph). In principle we could extend the functionality in CutVertexComposite to cut sets larger than a single vertex, but the run time grows exponentially with the size of the cut set.

dimod/reference/composites/cutvertex.py
```diff
@@ -0,0 +1,281 @@
...
+import itertools
+
+
+__all__ = ['CutVertexComposite']
+
+
+class CutVertexComposite(ComposedSampler):
```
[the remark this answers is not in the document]
jberwald: To make things easy to search for, would it be too weird to have an alias like `CutVertexComposite = CutVariableComposite`. Just because "cut vertex" is a more likely search term.

arcondello:
Closing in favour of https://github.com/dwavesystems/dwave-preprocessing/issues/36

PR #557, opened 2019-11-06
Change make_quadratic to use a loop rather than recursion

### LINKED ISSUE #556: Exceeded maximum recursion depth when converting large BinaryPolynomial to QUBO
When a large `BinaryPolynomial` instance is converted to a QUBO using `make_quadratic`, the recursive procedure exceeds maximum recursion depth.

Here is an example that reproduces the problem.
```
import dimod
from itertools import combinations
terms = combinations(range(20), 15)
poly = {k: 1.0 for k in terms}
bqm = dimod.make_quadratic(poly, 1.0, 'SPIN')
```

This should successfully return a bqm object instance. However, an error is raised instead,
`RecursionError: maximum recursion depth exceeded while calling a Python object`
and,
`TypeError: unsupported callable`

**Environment**
 - OS: [e.g., Ubuntu 18.04.4 LTS]
 - Python version: [e.g., 3.7.3]
 - Default maximum recursion depth limit

**Related issues**
This may be related to #550

arcondello:
Closes #556

### DIFF
--- a/dimod/higherorder/utils.py
+++ b/dimod/higherorder/utils.py
@@ -150,7 +150,7 @@ def make_quadratic(poly, strength, vartype=None, bqm=None):
     bqm.info['reduction'] = {}
 
     # we want to be able to mutate the polynomial so copy. We treat this as a
-    # dict but by using BinaryPolynomail we also get automatic handling of
+    # dict but by using BinaryPolynomial we also get automatic handling of
     # square terms
     poly = BinaryPolynomial(poly, vartype=bqm.vartype)
     variables = set().union(*poly)
@@ -195,9 +195,9 @@ def make_quadratic(poly, strength, vartype=None, bqm=None):
         constraint.scale(strength)
         for v, bias in constraint.linear.items():
             try:
-                poly[v,] += bias
+                poly[v, ] += bias
             except KeyError:
-                poly[v,] = bias
+                poly[v, ] = bias
         for uv, bias in constraint.quadratic.items():
             try:
                 poly[uv] += bias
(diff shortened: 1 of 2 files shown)


dimod/higherorder/utils.py
```diff
@@ -134,87 +152,77 @@ def make_quadratic(poly, strength, vartype=None, bqm=None):
     # we want to be able to mutate the polynomial so copy. We treat this as a
     # dict but by using BinaryPolynomail we also get automatic handling of
```
conta877: typo: BinaryPolynomail -> BinaryPolynomial

dimod/higherorder/utils.py
```diff
@@ -134,87 +152,77 @@ def make_quadratic(poly, strength, vartype=None, bqm=None):
...
+            u, v = term
+            bqm.add_interaction(u, v, bias)
+        elif len(term) == 1:
+            v, = term
+            bqm.add_variable(v, bias)
+        elif len(term) == 0:
+            bqm.add_offset(bias)
+        else:
+            # still has higher order terms, this shouldn't happen
+            raise RuntimeError
```
conta877: can we add a better error message here.

arcondello: Hmm, this is an "internal error", but suppose I could say that the user should file a bug report.

conta877: yeah it is internal - but if it happens its a legit bug and will help with bug reporting.

conta877: Also will save frustration for users.

PR #560, opened 2019-11-08
Add windows support in circle-ci

arcondello:
Also support python 3.8

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,10 @@
-version: 2
+version: 2.1
+
+orbs:
+  win: circleci/windows@2.2.0
+
 jobs:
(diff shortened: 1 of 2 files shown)


arcondello:
Current issues with circle-ci windows:
* it's very slow because one has to download/install python and boost. This might be able to be sped up with caching
* I hate that 2.1 does not allow for `.` in job names

arcondello:
This is semi-blocked until we remove the dependency on boost because `build_ext` does not support the `--include-dirs` - we could probably screw around with env variables like we did in https://github.com/dwavesystems/dimod/blob/4bfd23c28917d160aa2bd06fac234eeeb472bbd0/.appveyor.yml#L7
but for now I'll leave it be.

PR #561, opened 2019-11-12
Store an adjarray bqm in c++ as a pair (invars, outvars)

arcondello:
This makes function overloading a lot nicer e.g.
```
num_variables(bqm) // for all
```
rather than
```
num_variables(invars, outvars)  // for array
num_variables(bqm)  // for map, vector
```


arcondello:
The build errors are caused by using old pyenv installs through caching, not by this PR

PR #563, opened 2019-11-16
Add fixed variable composite for BinaryPolynomial

hsadeghidw:


arcondello:
The test failures are due to a stale cache on circle-ci and not because of this PR

arcondello:
Would it be possible to move the `FixedPolyVariableComposite` into `dimod/reference/composites/higherordercomposites.py` or `dimod/reference/composites/fixedvariables.py`? All of the other composites are in `higherordercomposites.py` but I am starting to feel like it would be nicer to have them in the same file as their quadratic counterpart (which was @conta877's original suggestion and I should have listened).

Also, I think to keep the pattern, this should be `PolyFixedVariableComposite`.

dimod/reference/composites/fixedpolyvariable.py
```diff
@@ -0,0 +1,122 @@
...
+import dimod
+
+
+__all__ = ['FixedPolyVariableComposite']
+
+
+class FixedPolyVariableComposite(ComposedPolySampler):
+    """Composite to fix variables of a problem to provided.
+
+    Fixes variables of a BinaryPolynomial and modifies linear and k-local terms
```
arcondello: in this sentence it should be `binary polynomial` (the noun) rather than `BinaryPolynomial` (the object).

dimod/reference/composites/fixedpolyvariable.py
```diff
@@ -0,0 +1,122 @@
...
+            sampleset = child.sample_poly(poly_copy, **parameters)
+            if len(sampleset):
+                return sampleset.append_variables(fixed_variables)
+            elif fixed_variables:
+                return type(sampleset).from_samples_bqm(fixed_variables, bqm=poly)
+            else:
+                return sampleset
+
+
+def fix_variables(poly, fixed_variables):
```
arcondello: I think this should just be a method on the `BinaryPolynomial`, to mirror [BQM.fix_variables](https://github.com/dwavesystems/dimod/blob/4bfd23c28917d160aa2bd06fac234eeeb472bbd0/dimod/binary_quadratic_model.py#L1045).

hsadeghidw: `fix_variables` creates a copy of the `BinaryPolynomial`, is that okay to be added to the class?

arcondello: Ah, good point. Ok, keep it separate for now then

dimod/reference/composites/higherordercomposites.py
```diff
@@ -471,3 +474,99 @@ def sample_poly(self, poly, **kwargs):
...
+
+
+class PolyFixedVariableComposite(ComposedPolySampler):
+    """Composite to fix variables of a problem to provided.
+
+    Fixes variables of a binary polynomial and modifies linear and k-local terms
+    accordingly. Returned samples include the fixed variable
+
+    Args:
+       sampler (:obj:`dimod.Sampler`):
```
arcondello: this should be `:obj:`dimod.PolySampler`.

randomir: Actually, it should be:
```
:class:`dimod.PolySampler`
```

arcondello: In this case the input it another object, not the type

arcondello: FWIW, I am having a hard time finding a reference/guide for `obj` vs `class`, except of course https://www.sphinx-doc.org/en/1.5/domains.html#python-roles. However, in dimod the convention has been to use `class` when referencing types or classes in the abstract and `obj` when referencing instantiated versions of those classes.

randomir: The point I'm trying to make here is you should always list the **type** in parenthesis when describing an argument. For example, `count (int)` looks correct, right? Well, `int` is a type/class.

I believe the documentation is clear about obj/class distinction -- `:obj:` is used to reference an object of unspecified type, and `:class:` is used to reference a class (which represents a type).

hsadeghidw: I went with :obj:`dimod.PolySampler` similar to other composites

randomir: That was a good call, @hsadeghidw! Local consistency trumps global one. According to PEP-8, at least. :smiley:

PR #564, opened 2019-11-26
File-like access to experimental BQMs

arcondello:
Closes #562 by duplication


arcondello:
Example usage:
```
arr = np.ones((1000, 1000), dtype=np.float64)
bqm = AdjArrayBQM(arr, 'BINARY')
with FileView(bqm) as fp:
    buff = fp.read()
new = load(buff)

assert bqm == new
```

PR #565, opened 2019-12-02
Install correct python version in windows circle-ci builds

arcondello:


arcondello:
Fixes the deploys and correctly runs the tests in python 3.5+. This does mean that we will need to continue to rely on appveyor for 2.7 and 3.4, though those are being deprecated anyway, see https://github.com/dwavesystems/dwave-ocean-sdk/issues/28.

PR #566, opened 2019-12-04
Fix variables to one of the allowed values

hsadeghidw:
When fixing the independent variables, the value of the fixed variable has no effect. However, if a solver checks for valid values only, we need to pass 0 or 1 for 'BINARY' and -1 or 1 for 'SPIN'. This pull request fixes this issue.

### DIFF
--- a/dimod/reference/composites/connectedcomponent.py
+++ b/dimod/reference/composites/connectedcomponent.py
@@ -94,9 +94,10 @@ class ConnectedComponentsComposite(ComposedSampler):
         if isinstance(components, set):
             components = [components]
         sampleset = None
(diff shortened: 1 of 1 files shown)


dimod/reference/composites/connectedcomponent.py
```diff
@@ -94,9 +94,10 @@ def sample(self, bqm, components=None, **parameters):
         if isinstance(components, set):
             components = [components]
         sampleset = None
+        fixed_value = list(bqm.vartype.value)[0]
```
arcondello: `bqm.vartype.value` is actually a FrozenSet which is unordered, this should be `fixed_value = min(bqm.vartype.value)`

PR #570, opened 2019-12-13
Avoid making a copy in the case of memoryviews and bytearrays

arcondello:


dimod/serialization/fileview.py
```diff
@@ -208,8 +255,8 @@ def load(fp, cls=None):
 
     """
 
-    if isinstance(fp, (bytes, bytearray)):
-        fp = io.BytesIO(fp)
+    if isinstance(fp, (bytes, bytearray, memoryview)):
```
wbernoudy: Still going to use the custom class with `bytes`?

arcondello: No harm and consistency is good I think

wbernoudy: :+1:

PR #571, opened 2019-12-17
Restructure cyBQM tests

arcondello:


arcondello:
I have run into some issues with this approach, marking it as WIP for now while I fix (it seems you cannot change an open PR to a draft PR).

arcondello:
These commits have been folded into other branches

PR #584, opened 2020-01-21
cleared signed/unsigned comparison warnings from roof_duality

boothby:
It's really hard to wade through compiler warnings... so I changed a bunch of `int` to `size_t`.  I did this in a rather ham-fisted manner and I think it's appropriate to update `CompressedMatrix` to use `size_t` for indices... but this is probably okay for now.


arcondello:
This is currently pointing at dev branch so closing for now.

PR #600, opened 2020-02-10
Make the old-style bqm a subclass of the cybqms

arcondello:


dimod/binary_quadratic_model.py
```diff
@@ -205,2361 +193,244 @@ class for the K4 complete graph, where the nodes have biases
...
     def __contains__(self, v):
+        msg = ("treating binary quadratic models as containers is deprecated, "
+               "please use `v in bqm.variables` or `bqm.has_variable(v)` "
+               "instead")
+        warnings.warn(msg, DeprecationWarning, stacklevel=2)
         return v in self.adj
 
     def __iter__(self):
+        msg = ("iterating over binary quadratic models is deprecated, "
+               "please use `for in bqm.variables` or `bqm.iter_variables()` "
```
JoelPasvolsky: --> `for v in bqm.variables`

dimod/binary_quadratic_model.py
```diff
@@ -205,2361 +193,244 @@ class for the K4 complete graph, where the nodes have biases
...
+                to represent the bias values in the serialized format.
 
-    @property
-    def binary(self):
-        """:class:`.BinaryQuadraticModel`: An instance of the QUBO model subclass of
-        the :class:`.BinaryQuadraticModel` superclass (a BQM with binary variables).
+            bytes_type (class, optional, default=bytes):
+                This class will be used to wrap the bytes objects in the
+                serialization if `use_bytes` is true. Useful for when using
+                Python 2 and using BSON encoding, which will not accept the raw
```
JoelPasvolsky: --> "If `use_bytes` is true, this class is used to wrap the bytes objects in the serialization" to save users time. I mention it because of the "Useful for when using" typo. 
BTW, "when using Python 2 and ", why are we still support Py2?

dimod/binary_quadratic_model.py
```diff
@@ -205,2361 +193,244 @@ class for the K4 complete graph, where the nodes have biases
...
-                :class:`.Vartype.BINARY`.
+            >>> bqm = dimod.BinaryQuadraticModel({'a': -1.0, 'b': 1.0}, {('a', 'b'): -1.0}, 0.0, dimod.SPIN)
+            >>> doc = bqm.to_serializable(use_bytes=True)
+            >>> b = bson.BSON.encode(doc)  # doctest: +SKIP
 
-        Examples:
-            This example creates an Ising model with two variables, adds a third,
-            adds to the bias of the initial interaction, and creates
-            a new interaction.
+            Encode using BSON in python 2.7. Because :class:`bytes` is an alias for :class:`str`,
```
JoelPasvolsky: Can't we just get rid of this?

PR #603, opened 2020-02-10
Add as_bqm function

arcondello:


dimod/bqm/construction.py
```diff
@@ -0,0 +1,107 @@
...
+    if cls is None:
+        if isinstance(args[0], BQM):
+            cls = type(args[0])
+        else:
+            cls = AdjVectorBQM
+    elif isinstance(cls, (Sequence, Set)):  # want Collection, but not in 3.5
+        classes = tuple(cls)
+        if not classes:
+            raise ValueError("cls kwarg should be a type or a list of types")
+        if type(args[0]) in classes:
```
JoelPasvolsky: I have to run but will this be okay for ``as_bqm(n, vartype, cls=`` where vartype is legit and cls not and n an integer, or even a float?

Issue #610, opened 2020-02-20
Update generators to use the new BQM classes


arcondello:
Update the BQM [generation code](https://github.com/dwavesystems/dimod/tree/dd33fe988534bd7043faec223cecfc764a282026/dimod/generators) to use the new bqms added by #533. Probably they should use a similar class specification as added in #603.

arcondello:
Closed by #993 (and in many cases implicitly by #828 )

PR #621, opened 2020-02-27
WIP: Problem generators

conta877:
adding two new problem generators

anti_crossing these are dickson instances with growing size
and doped - creates a doped FM or AFM problem.

### DIFF
--- a/dimod/generators/anti_crossing.py
+++ b/dimod/generators/anti_crossing.py
@@ -14,11 +14,26 @@
 #
 # =============================================================================
 
+from dimod import BinaryQuadraticModel
+__all__ = ['anti_crossing_clique', 'anti_crossing_loops']
 
-def anti_crossing(num_qubits):
-    """ given number of qubits, the code will generate instances with perturbative anticrossing. These
-        instances are known to be hard for Quantum annealing. For more information please refer to
-        [Phys. Rev. A 85, 032303] and [Phys. Rev. A 96, 042322] """
+
+
+def anti_crossing_clique(num_qubits):
+    """Anti crossing problems with a single clique.
+
+    Given the number of qubits, the code will generate a clique of
+    size num_qubits/2 each qubit ferromagnetically coupled to a partner qubit with opposite bias. Single qubit in
+    the cluster will have no bias applied
+
+    Args:
+        num_qubits (int):
+            number of qubits to use to generate the problem
+
+    Returns:
+        :obj:`.BinaryQuadraticModel`.
+
+    """
 
     if num_qubits % 2 != 0 or num_qubits < 6:
         raise ValueError('num_qubits  must be an even number > 6')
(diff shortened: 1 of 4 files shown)


dimod/generators/anti_crossing.py
```diff
@@ -0,0 +1,90 @@
...
+from dimod import BinaryQuadraticModel
+__all__ = ['anti_crossing_clique', 'anti_crossing_loops']
+
+
+
+def anti_crossing_clique(num_qubits):
+    """Anti crossing problems with a single clique.
+
+    Given the number of qubits, the code will generate a clique of
+    size num_qubits/2 each qubit ferromagnetically coupled to a partner qubit with opposite bias. Single qubit in
```
arcondello: 80 char width

dimod/generators/anti_crossing.py
```diff
@@ -0,0 +1,90 @@
...
+
+from dimod import BinaryQuadraticModel
+__all__ = ['anti_crossing_clique', 'anti_crossing_loops']
+
+
+
+def anti_crossing_clique(num_qubits):
+    """Anti crossing problems with a single clique.
+
+    Given the number of qubits, the code will generate a clique of
```
arcondello: Maybe we should use "variables/interactions" rather than "qubit/couplers". Though I suppose this is mostly a useful in the context of QPUs. Could go either way.

dimod/generators/anti_crossing.py
```diff
@@ -0,0 +1,90 @@
...
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+# =============================================================================
+
+from dimod import BinaryQuadraticModel
+__all__ = ['anti_crossing_clique', 'anti_crossing_loops']
```
arcondello: Line between import and `__all__`.

dimod/generators/anti_crossing.py
```diff
@@ -0,0 +1,90 @@
...
+        J[(n, n + hf)] = -1
+        h[n] = 1
+        h[n + hf] = -1
+    h[1] = 0
+    return BinaryQuadraticModel.from_ising(h, J)
+
+
+def anti_crossing_loops(num_qubits):
+    """ Anti crossing problems with two loops. These instances are copies of the instance studied in
+    [Nature Comms. 4, 1903 (2013)]
```
arcondello: Should make this a full sphinx reference, see for instance https://github.com/dwavesystems/dimod/blob/fa7ee3d6ab63e0fbe799090d86952ef1255944d5/dimod/roof_duality/fix_variables.py#L67

Also no space between `""" A`

dimod/generators/random.py
```diff
@@ -205,3 +205,58 @@ def ran_r(r, graph, cls=BinaryQuadraticModel, seed=None):
 
     return cls.from_numpy_vectors(ldata, (irow, icol, qdata), offset, vartype='SPIN',
                                   variable_order=variables)
+
+
+@graph_argument('graph')
+def doped(p, graph, cls=BinaryQuadraticModel, seed=None, fm=True):
+    """Generate bqm with weighted couplings parametrized by the doping variable p
+
```
arcondello: Need a paragraph here providing more explanation

arcondello:
One note overall, we probably want to generate the BQMs directly rather than making the `h, J` dictionaries then making a BQM. This will matter a lot more when we want to scale up the size, especially when we get around to #610 .

Issue #466, opened 2019-04-23
FCL.py creates FLs not clusters


conta877:
badly named file.
fcl stands for frustrated cluster loops. however the code inside generates loops given a graph without clustering.

arcondello:
I think a good solution would be to add frustrated_cluster_loop generators as well. Is there a paper that defines them somewhere that we can use to add them?

conta877:
for chimera, clusters are defined as unit tiles. can do the same for pegasus - use chimera unit tiles.

PR #622, opened 2020-02-27
rename FCL.py to FL.py

conta877:
I am aware that this potentially breaks backwards compatibility for certain people but there is nothing cluster in that file. Adding clusters requires a cluster definition so I'd suggest adding FCLs under chimera.py and creating a pegasus.py if one wants chimera shaped clusters

Feel free the reject


arcondello:
Closes #466

arcondello:
If we're renaming, probably it should be to `frustrated_loop.py`.

PR #627, opened 2020-03-03
Add a simple sampler-tests generating decorator

arcondello:
I also fixed some bugs that this testing framework uncovered. There are two bugs remaining to fix but they are significant enough to be beyond the scope of this PR.


arcondello:
This covers the labelling and the large problems can be filtered using the `max_num_variables` parameter. Or by line up do you mean correct answers?

tests/test_scalecomposite.py
```diff
@@ -118,6 +119,8 @@ def properties(self):
         return self.child.properties()
 
 
+# @dimod.testing.load_sampler_bqm_tests(ScaleComposite(ExactSolver()))
+# @dimod.testing.load_sampler_bqm_tests(ScaleComposite(dimod.NullSampler))
```
m3ller: Out of curiosity, why aren't these samplers tested here as well?

dimod/testing/sampler.py
```diff
@@ -0,0 +1,240 @@
...
+        method.__name__ += '_' + '_'.join(specifiers)
+
+    return method
+
+
+def create_bqm_tests(sampler, max_num_variables=None):
+    """Yields test methods that can be added to a TestCase."""
+
+    if max_num_variables is None:
+        max_num_variables = float('inf')
```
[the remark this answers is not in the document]
arcondello: Definitely could, though I do say in the description that it should be an `int`, then immediately violate that :smile: . Also `None` makes it nicer when/if this function is later wrapped.

tests/test_scalecomposite.py
arcondello: They fail! There are two issues #629 #628 linked above.

wbernoudy:
I mean "correct" answers. If I submit h={0: -100, 1: 100}, and get magnetizations of {0: -0.98, 1: 0.99}, the labeling is probably reversed.

PR #638, opened 2020-03-26
Expand doctest coverage & update examples (part 1)

JoelPasvolsky:
@arcondello, we can see just before SDK release whether to merge or leave till next time. I'll still be working on this tomorrow morning.

### DIFF
--- a/dimod/core/sampler.py
+++ b/dimod/core/sampler.py
@@ -27,14 +27,14 @@ For example, the following steps show how to easily create a dimod sampler. It i
 sufficient to implement a single method (in this example the :meth:`sample_ising` method)
 to create a dimod sampler with the :class:`.Sampler` class.
 
-.. code-block:: python
+.. testcode::
 
     class LinearIsingSampler(dimod.Sampler):
 
         def sample_ising(self, h, J):
             sample = linear_ising(h, J)
             energy = dimod.ising_energy(sample, h, J)
-            return dimod.SampleSet.from_samples([sample], energy=[energy]})
+            return dimod.SampleSet.from_samples([sample], vartype='SPIN', energy=[energy])
 
         @property
         def properties(self):
@@ -47,7 +47,7 @@ to create a dimod sampler with the :class:`.Sampler` class.
 For this example, the implemented sampler :meth:`~.Sampler.sample_ising` can be based on
 a simple placeholder function, which returns a sample that minimizes the linear terms:
 
-.. code-block:: python
+.. testcode::
 
     def linear_ising(h, J):
         sample = {}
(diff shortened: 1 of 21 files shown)


dimod/core/sampler.py
```diff
@@ -27,14 +27,14 @@
 sufficient to implement a single method (in this example the :meth:`sample_ising` method)
 to create a dimod sampler with the :class:`.Sampler` class.
 
-.. code-block:: python
+.. testcode::
```
arcondello: This one is new to me and the docs for it are lousy. I gather that it just is a regular `code-block` that also allows doctests?

JoelPasvolsky: Yes. I use it here, for example: https://docs.ocean.dwavesys.com/en/latest/examples/multi_gate.html

dimod/reference/composites/higherordercomposites.py
```diff
@@ -409,18 +411,18 @@ def sample_poly(self, poly, scalar=None, bias_range=1, poly_range=None,
...
+    treat the lowest-energy samples. This composite layer allows one to
+    pre-select the samples within a multi-composite pipeline.
 
     Args:
         child_sampler (:obj:`dimod.PolySampler`):
             A dimod binary polynomial sampler.
 
         n (int):
-            Maximum number of rows in the returned sample set.
+            Maximum number of rows in the returned SampleSet.
```
arcondello: "sample set" is to "binary quadratic model" as "SampleSet" is to "BinaryQuadraticModel"

dimod/reference/composites/structure.py
```diff
@@ -104,7 +104,7 @@ def sample(self, bqm, **sample_kwargs):
             >>> edges = [(0, 1)]
             >>> composed_sampler = dimod.StructureComposite(dimod.ExactSolver(),
             ...                          nodes, edges)
-            response = composed_sampler.sample_ising({0: 1, 1: 1}, {})
+            >>> response = composed_sampler.sample_ising({0: 1, 1: 1}, {})
```
arcondello: this aught to be "sampleset"

JoelPasvolsky: For dimod, because it is not sampling, I have been mostly changing to "solutions" or, as here when I was rushing, leaving unchanged for now.

dimod/reference/composites/higherordercomposites.py
JoelPasvolsky: Fair enough. I was on a rampage through these because I was finding "sampleSet".

dimod/reference/composites/structure.py
arcondello: Calling it "solutions" is really only valid in the case of the `ExactSolver` and I think it's better to be consistent with `sampleset`. But agree that leaving it as `response` is fine.

JoelPasvolsky: My typo: I meant "For dimod.ExactSolver, because"

PR #639, opened 2020-03-30
Add BQM.relabel_variables_as_integers

arcondello:


dimod/core/bqm.py
```diff
@@ -727,6 +727,30 @@ def min_and_max(iterable):
...
+                A binary quadratic model with the variables relabeled. If
+                `inplace` is set to True, returns itself.
+
+                dict: The mapping that will restore the original labels.
+
+        """
+        if not inplace:
+            return self.copy().relabel_variables(inplace=True)
+
+        mapping = dict((v, i) for i, v in enumerate(self.variables) if i != v)
```
wbernoudy: Is there a cost to relabeling variables to the same label?

arcondello: Not really, but cleaner this way

PR #645, opened 2020-04-24
Add examples for serialization.coo

arcondello:


dimod/serialization/coo.py
```diff
@@ -12,9 +12,69 @@
...
+
+    >>> Q = {(0, 0): -1, (0, 1): 1, (1, 2): -4.5}
+    >>> bqm = dimod.BinaryQuadraticModel.from_qubo(Q)
+    >>> print(coo.dumps(bqm, vartype_header=True))
+    # vartype=BINARY
+    0 0 -1.000000
+    0 1 1.000000
+    1 2 -4.500000
+
+    Loading the same, note that we must specify the vartype
```
JoelPasvolsky, line 51: Ungrammatical. I think you want something like "When loading, you must specify the vartype."

dimod/serialization/coo.py
```diff
@@ -12,13 +12,73 @@
...
 
 import re
+
+from numbers import Integral
+
 from dimod import Vartype
 from dimod.binary_quadratic_model import BinaryQuadraticModel
 
-_LINE_REGEX = r'^\s*(\d+)\s+(\d+)\s+([+-]?(?:[0-9]*[.]):?[0-9]+)\s*$'
+_LINE_REGEX = r'^\s*(\d+)\s+(\d+)\s+([+-]?(?:[0-9]*[.]?):?[0-9]*)\s*$'
```
randomir, line 81: This will parse lines like:
```
1 2 3:4
1 2 3:
1 2 :4
1 2 :
```

I'm not sure what `:` is supposed to stand for. Looks like the bias parsing code later will fail on it.

arcondello: Huh, I am also not sure. It does work (because it's modified by `?`) for the stuff we want, but obviously we don't want the cases you raised. Wonder how that got in there in the first place. Anyway, will remove

randomir: Might've been a typo when [adding the preceding non-capturing group](https://github.com/dwavesystems/dimod/commit/eecf05c3af62f004a4179869c7a6670f269f289e#diff-01222c8607c470b70355f012b961e0b6R21).

dimod/serialization/coo.py
```diff
@@ -12,13 +12,73 @@
...
 
 import re
+
+from numbers import Integral
+
 from dimod import Vartype
 from dimod.binary_quadratic_model import BinaryQuadraticModel
 
-_LINE_REGEX = r'^\s*(\d+)\s+(\d+)\s+([+-]?(?:[0-9]*[.]):?[0-9]+)\s*$'
+_LINE_REGEX = r'^\s*(\d+)\s+(\d+)\s+([+-]?(?:[0-9]*[.]?)[0-9]*)\s*$'
```
randomir, line 81: This will now accept:
```
1 2 .
```

randomir: I'd go with:
```
r'^\s*(\d+)\s+(\d+)\s+([+-]?\d*(\.\d+)?)\s*$'
```
This accepts:
```
1 2 3
1 2 3.4
1 2 .4
```
but not:
```
1 2 3.
```

randomir, line 81:
```suggestion
_LINE_REGEX = r'^\s*(\d+)\s+(\d+)\s+([+-]?\d*(\.\d+)?)\s*$'
```

arcondello: Sigh, I hate regex. Thanks for the check and the suggested change!

dimod/serialization/coo.py
```diff
@@ -78,7 +78,7 @@
 from dimod import Vartype
 from dimod.binary_quadratic_model import BinaryQuadraticModel
 
-_LINE_REGEX = r'^\s*(\d+)\s+(\d+)\s+([+-]?(?:[0-9]*[.]?)[0-9]*)\s*$'
+_LINE_REGEX = r'^\s*(\d+)\s+(\d+)\s+([+-]?\d*(?:\.\d+)?)\s*$'
```
randomir, line 81: Oh, yes, you need "triplets" below, that's why the non-capturing group. I'd recommend named groups in general, but let's leave that for another day :slightly_smiling_face:

PR #654, opened 2020-05-13
Sort different types supported by Python 2 and 3

joseppinilla:
Something to consider, since the expected behaviour is that all variables are sorted deterministically and only Python 2.7 has the expected behaviour right now.

This change could be done to all other instances of `sort` not being done deterministically for Python 3.


arcondello:
Unfortunately sorting this way is unstable. E.g.
```
>>> sorted([1, '1'], key=lamba x: str(x))
[1, '1']
>>> sorted(['1', 1], key=lamba x: str(x))
['1', 1]
```

There are ways to get python 2 sorting behaviour but they are complicated. See [stack overflow](https://stackoverflow.com/questions/26575183/how-can-i-get-2-x-like-sorting-behaviour-in-python-3-x).

PR #662, opened 2020-06-08
Unify variables object between cyBQM and SampleSet

arcondello:


arcondello:
The `BiDict` is the first part of the implementation, but I wanted to put it up now for feedback (especially about the docs).

JoelPasvolsky:
I'll take a look, but am curious what are the reasons not to use an existing one like https://bidict.readthedocs.io/en/master/ instead? Is the MIT license a restriction or will this one be faster?

arcondello:
Avoids bringing in another dependency for a pretty simple object. Also their license is weak copy-left which we'd like to avoid.

dimod/bidict.py
```diff
@@ -0,0 +1,140 @@
...
+#    limitations under the License.
+
+from collections.abc import Mapping, MutableMapping
+
+
+__all__ = ['BiDict']
+
+
+class BiDict(MutableMapping):
+    """A BiDict is a dictionary with fast reverse lookup.
```
JoelPasvolsky, line 22: A dict-like class with fast reverse lookup.

dimod/bidict.py
```diff
@@ -0,0 +1,140 @@
...
+
+    All dictionary methods are supported. In addition there is an `inverse`
+    attribute, which it itself a `BiDict`, storing the values as keys
+    and vice versa.
```
JoelPasvolsky, line 26: Provides an ``inverse`` attribute, which stores values as keys in a BiDict, to enable 
fast retrieval of a key for a given value of the original BiDict.

Supports all dictionary methods.  

All values in a BiDict must be unique.

dimod/bidict.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Attributes:
+
+        inverse:
+            Another `BiDict`, equivalent to one constructed by
+            `{value: key for key, value in d}`.
```
JoelPasvolsky, line 32: A BiDict with keys and values swapped compared to the instantiated BiDict. 
Equivalent to constructing dict `{value: key for key, value in d}`, where ``d`` is
the instantiated BiDict.

dimod/bidict.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Examples:
+
+        >>> from dimod.bidict import BiDict
+
+        `BiDict` can be constructed and interacted with just like a normal
+        dictionary, but it also has an `inverse` attribute, which is itself
+        another (inversed) `BiDict`.
```
JoelPasvolsky, line 40: I'd delete `BiDict` can be constructed and ... itself  another (inversed) `BiDict`.

dimod/bidict.py
```diff
@@ -0,0 +1,140 @@
...
+    and vice versa.
+
+    Attributes:
+
+        inverse:
+            Another `BiDict`, equivalent to one constructed by
+            `{value: key for key, value in d}`.
+
+    Examples:
+
```
JoelPasvolsky, line 35: This example creates a simple BiDict with key-values ``a=1`` and ``b=2`` and looks up key ``a`` 
by its value.

dimod/bidict.py
```diff
@@ -0,0 +1,140 @@
...
+        >>> bidict = BiDict(a=1, b=2)
+        >>> bidict['a']
+        1
+        >>> bidict.inverse[1]
+        'a'
+        >>> bidict.inverse.inverse['a']
+        1
+
+        Note that this means that the values must also be unique.
```
JoelPasvolsky, line 50: Delete this note from here

dimod/bidict.py
```diff
@@ -0,0 +1,140 @@
...
+
```
JoelPasvolsky, line 51: This example reuses a value for a new key, and by doing so deletes an existing entry with that value (values in 
a BiDict must be unique).

Issue #660, opened 2020-06-02
Create local headers in FileView serialization


arcondello:
Inspired by [zip's format](https://en.wikipedia.org/wiki/Zip_(file_format)), it might make sense for the offset, linear, quadratic, and variable labels (see #659) to have their own local headers. The offsets could also be stored in the header for easier reading, rather than being implicitly derived ([example](https://github.com/dwavesystems/dimod/blob/7bec06d6c868efef012b817ffafb5043440ffd8d/dimod/serialization/fileview.py#L168), [example](https://github.com/dwavesystems/dimod/blob/7bec06d6c868efef012b817ffafb5043440ffd8d/dimod/bqm/adjarraybqm.pyx#L481)).

PR #667, opened 2020-06-19
Fileview 2.0

### LINKED ISSUE #659: Move variable labels outside of the data field in FileView serialization
Currently the variable labels are [encoded in the header](https://github.com/dwavesystems/dimod/blob/7bec06d6c868efef012b817ffafb5043440ffd8d/dimod/serialization/fileview.py#L56). However, this means that the header length can be (arbitrarily) long. A better approach would be to move the labels outside of the header.

*Additional Context*
If the labels were stored _after_ the quadratic data, it would make https://github.com/dwavesystems/dimod/issues/652  cleaner

arcondello:
Closes #659


arcondello:
Error is caused by different in neighborhood dtype in windows 32 (`uint32`) vs 64 bit systems (`uint64`). Will fix.

arcondello:
Current failure is on circle-ci's side. This code is not the larger refactor that I'd like but I can make that as a subsequent PR.

dimod/serialization/fileview.py
```diff
@@ -12,15 +12,109 @@
...
+
+.. code-block:: python
+
+    dict(shape=bqm.shape,
+         dtype=bqm.dtype.name,
+         itype=bqm.itype.name,
+         ntype=bqm.ntype.name,
+         vartype=bqm.vartype.name,
+         type=type(bqm).__name__,
+         variables=all(v == i for i, v in enumerate(bqm.variables)),
```
randomir, line 84: This seems counter-intuitive: if `variables==False`, then variables section follows?

dimod/serialization/fileview.py
```diff
@@ -12,15 +12,109 @@
...
+         )
+
+If the BQM is index-labeled, then no additional data is added. Otherwise,
+a new section is appended after the bias data.
+
+The first 11 bytes are exactly "[VARIABLES]".
```
randomir, line 90: Not sure why the "section magic"?

dimod/serialization/fileview.py
randomir: From code, I see it's actually `variables=not all(...)`, so as expected (just update the docs).

dimod/serialization/fileview.py
```diff
@@ -12,15 +12,109 @@
...
+.. code-block:: python
+
+    dict(shape=bqm.shape,
+         dtype=bqm.dtype.name,
+         itype=bqm.itype.name,
+         ntype=bqm.ntype.name,
+         vartype=bqm.vartype.name,
+         type=type(bqm).__name__,
+         variables=all(v == i for i, v in enumerate(bqm.variables)),
+         )
```
randomir, line 85: In data serialization, some redundancy is good, IMO.

Instead of requiring manual calculation of bqm data section length, I would include that in the header. Same for the variables section.

Alternatively, I would include section length as the first `uint4` of each section, as long as the section order is well defined. (One option is to also have offsets of important sections in header, but we're at only two sections, so that's not yet needed.)

arcondello: Hmm, I had tried to avoid implementing https://github.com/dwavesystems/dimod/issues/660 in the interests of not changing the format too much. But I think you're right and we should have each local header (`[OFFSET]`, `[LINEAR]`, `[QUADRATIC]`, `[VARIABLES]`) with its length. The order is well defined.

arcondello: If we're going to 2.0, might as well. Also changing the encoding from `ascii` to `utf8` a la https://numpy.org/doc/stable/reference/generated/numpy.lib.format.html#format-version-3-0

randomir: Section headers yes, section magic no, IMO.

You don't need UTF8 because you encode user text (labels) with JSON.

arcondello: What would the header be then? Just 4 bytes of length?

randomir: On second thought, there might be many sections, so it makes sense to identify the section type in the header. One byte would be sufficient for this, but if we want to make the file human-semi-readable, a text label works better. I'd prefer one byte for section type field, though :slightly_smiling_face:

arcondello: I think I am leaning towards having the section offsets in the header, but will try some things.

arcondello: I ended up going with 4 bytes. Seemed like a reasonable compromise between short and human readable.

randomir:
Looks good.

PR #672, opened 2020-07-06
Add AdjArrayBQM.data and AdjArrayBQM.from_data

arcondello:
We need to use cdef classes for the reference counting to work correctly. I will see if there are ways to reduce the boilerplate but the typing creates a lot of repetition.


arcondello:
Looks like I need to go recheck my understanding of garbage collection and Cython :smile:

PR #674, opened 2020-07-09
Don't deploy for python 3.9 (yet)

arcondello:
Numpy does not have 3.9 wheels yet to manylinux deploy fails for 3.9.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -348,7 +348,7 @@ jobs:
           name: build wheels
           command: |
             for PYBIN in /opt/python/*/bin; do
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -348,7 +348,7 @@ jobs:
           name: build wheels
           command: |
             for PYBIN in /opt/python/*/bin; do
-              if "${PYBIN}/python" -c "import sys; sys.exit(sys.version_info>=(3, 5))"; then continue; fi;
+              if "${PYBIN}/python" -c "import sys; sys.exit(sys.version_info>=(3, 5) and sys.version_info<(3, 9))"; then continue; fi;
```
randomir, line 351:
```suggestion
              if "${PYBIN}/python" -c "import sys; sys.exit((3, 5) <= sys.version_info < (3, 9))"; then continue; fi;
```

PR #682, opened 2020-07-23
Added to_networkx_graph method to Structured Samplers

hhtong:


arcondello:
Can you also add it to [the table in the docs](https://github.com/dwavesystems/dimod/blame/master/docs/reference/sampler_composites/api.rst#L31)?

tests/test_structured.py
```diff
@@ -62,3 +62,36 @@ def edgelist(self):
         self.assertEqual(sampler.structure.nodelist, [0, 1])
         self.assertEqual(sampler.structure.edgelist, [(0, 1)])
         self.assertEqual(sampler.structure.adjacency, {0: {1}, 1: {0}})
+
+    def test_networkx_graph(self):
+
+        # Create simple NetworkX graph to compare to
+        import networkx as nx
```
arcondello, line 69: Might as well make this a soft test
```
try:
    import networkx as nx
except ImportError:
    raise unittest.SkipTest("no networkx installed")
```

arcondello: We might not do that elsewhere, but we should :smile:

Issue #402, opened 2019-02-08
Minimum and maximum method for biases


arcondello:
**Application**
It is sometimes useful to know the range of biases used in a bqm.

**Proposed Solution**
Could make an attribute on LinearView and QuadraticView, so
```
bqm.linear.min()
bqm.linear.max()

bqm.quadratic.min()
bqm.quadratic.max()
```

**Alternatives Considered**
This does not save significant code relative to the current method `min(bqm.linear.values())` and might be considered confusing since `min(bqm.linear)` would return the minimum label rather than the bias.

We could also make it an attribute rather than a method, but a method allows things like
```
bqm.linear.max(default=0)
```

jberwald:
That looks helpful. Have you considered `linear_min, linear_max, quad_min, quad_max = bqm.ranges()`?  Maybe `ranges` is quite right, `bounding_box` seems more appropriate.

arcondello:
I like bounding box! It could also similarly do `(lmin, lmax), (qmin, qmax) = lrange, qrange = bqm.bounding_box()`

arcondello:
It would also be nice to be able to calculate things like mean, median, standard deviation etc.

randomir:
Since all these (min/max/mean/quartiles/etc) are [one-liners](https://docs.python.org/3/library/statistics.html), I'd vote to add only the most frequently used ones.

That is, unless you planned to distribute calculation over bqm modify ops (which would be easy to do for some, like mean, but would require O(n) space for min/max; again worth doing only if used very frequently).

arcondello:
Nice! I missed that package, joys of dropping 2.7...

One reason to implement some of these directly is that, although this package handles iterators, it does so by [converting to a list](https://github.com/python/cpython/blob/f56c75ed53dcad4d59dff4377ae463d6b96acd3e/Lib/statistics.py#L311). That's fine for `BQM.linear` but for `BQM.quadratic`, especially on the large cyBQMs, the performance difference will be very noticeable.

arcondello:
Example:
```
import time

import numpy as np

import dimod
cimport dimod

from cython.operator cimport dereference as deref, preincrement as inc


def pymax(bqm):
    return max(bqm.quadratic.values(), default=0)

def cymax(dimod.cyBQM bqm):
    cdef double max_ = 0

    cdef Py_ssize_t vi
    for vi in range(bqm.bqm_.num_variables()):
        span = bqm.bqm_.neighborhood(vi)
[...]

PR #683, opened 2020-07-23
Added min/max methods for biases 

hhtong:


dimod/bqm/utils.pyx
```diff
@@ -40,6 +42,53 @@ cdef object as_numpy_scalar(double a, np.dtype dtype):
     """Note that the memory is interpreted to match dtype, not a cast"""
     return PyArray_Scalar(&a, dtype, None)
 
+def cylmin(cyBQM bqm):
+    cdef double min_ = sys.maxsize
```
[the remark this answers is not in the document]
randomir: Same for `cyqmin()`.

dimod/bqm/utils.pyx
```diff
@@ -40,6 +42,53 @@ cdef object as_numpy_scalar(double a, np.dtype dtype):
...
+    cdef Py_ssize_t vi
+    for vi in range(bqm.bqm_.num_variables()):
+        val = bqm.bqm_.get_linear(vi)
+        if (val < min_):
+            min_ = val
+    
+    return min_
+
+def cylmax(cyBQM bqm):
+    cdef double max_ = 0
```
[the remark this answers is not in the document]
randomir: Same for `cyqmax()`.

dimod/core/bqm.py
```diff
@@ -141,6 +143,22 @@ def __setitem__(self, v, bias):
         # inherits its ability to reshape the bqm from the `.set_linear` method
         self._bqm.set_linear(v, bias)
 
+    def min(self):
+        try:
+            return cylmin(self._bqm)
+        except TypeError:
+            pass
```
randomir, line 150: Why would this fail?

tests/test_minmax.py
```diff
@@ -0,0 +1,65 @@
...
+#
+# ================================================================================================
+
+import unittest
+import dimod
+import numpy as np
+
+num_vars = 1000
+
+class TestMinMax(unittest.TestCase):
```
randomir, line 23: I would probably add these tests next to BQM methods tests in [`test_bqm.py`](https://github.com/dwavesystems/dimod/blob/master/tests/test_bqm.py) or [`test_binary_quadratic_model.py`](https://github.com/dwavesystems/dimod/blob/master/tests/test_binary_quadratic_model.py).

tests/test_minmax.py
```diff
@@ -0,0 +1,65 @@
...
+
+        self.linear_minmax(bqm) 
+        self.quad_minmax(bqm)
+
+    def test_adjmap_minmax(self):
+        D = np.arange(num_vars*num_vars).reshape((num_vars, num_vars))
+        bqm = dimod.AdjMapBQM(D, 'SPIN') 
+
+        self.linear_minmax(bqm) 
+        self.quad_minmax(bqm)
```
randomir, line 65: These can be parametrized, see for example [`test_bqm.py#L34-L47`](https://github.com/dwavesystems/dimod/blob/d697266e41f2e073c801221b7da46455a0ef1116/tests/test_bqm.py#L34-L47).

dimod/core/bqm.py
arcondello: `cylin` and others are defined only for the cyBQMs. So this will fail for AdjDict for instance.

dimod/bqm/utils.pyx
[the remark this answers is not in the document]
arcondello: This will currently return `sys.maxsize` for an empty BQM, which I think is very confusing behavior. I would rather follow python's behavior and fail when `bqm.num_interactions == 0`, but allow the user to provide a default.

NumPy has a [similar notion](https://numpy.org/doc/stable/reference/generated/numpy.amin.html#numpy.amin).

tests/test_minmax.py
arcondello: Also should test with empty BQMs, see note above.

tests/test_minmax.py
arcondello: `test_bqm.py` please :smile:

dimod/bqm/utils.pyx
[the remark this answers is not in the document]
randomir: I like NumPy's `initial`.

dimod/bqm/utils.pyx
```diff
@@ -40,6 +42,78 @@ cdef object as_numpy_scalar(double a, np.dtype dtype):
     """Note that the memory is interpreted to match dtype, not a cast"""
     return PyArray_Scalar(&a, dtype, None)
 
+def cylmin(cyBQM bqm, **kwargs):
```
arcondello, line 45: Probably better to name the keyword argument explicitly, e.g `cylmin(cyBQM bqm, object default=None)`. The `object` type specifier is a cython thing, but better explicit than implicit...

One reason to do it this way it that it helps catch hard-to-find errors like
```
cylmin(bqm, dfault=1)
```

dimod/bqm/utils.pyx
```diff
@@ -40,6 +42,78 @@ cdef object as_numpy_scalar(double a, np.dtype dtype):
...
+        else:
+            raise ValueError("Argument is an empty sequence")
+
+    cdef double max_ = -sys.float_info.max
+
+    cdef Py_ssize_t vi
+    for vi in range(bqm.bqm_.num_variables()):
+        span = bqm.bqm_.neighborhood(vi)
+        
+        while span.first != span.second:
```
arcondello, line 110: As a detail, the biases are kept symmetrically, so you could do 
```
while span.first != span.second and deref(span.first).first < vi:
```
which would speed this up a bit.

tests/test_bqm.py
```diff
@@ -2312,3 +2312,49 @@ def test_quadratic_setitem(self, name, BQM):
...
+        bqm = BQM(D, 'SPIN') 
+
+        qmin = min(bqm.quadratic.values())
+        self.assertEqual(qmin, bqm.quadratic.min())
+       
+        qmax = max(bqm.quadratic.values())
+        self.assertEqual(qmax, bqm.quadratic.max())
+
+    @parameterized.expand([(cls.__name__, cls) for cls in BQM_SUBCLASSES])
+    def test_lin_minmax_empty(self, name, BQM):
```
arcondello, line 2342: I think this shouldn't be named `test_lin_minmax_empty` since the BQM has 500 variables? I would add some empty tests and make sure that they throw the correct error, something like
```
    BQM('SPIN').linear.min()
self.assertEqual(BQM('SPIN').linear.min(10), 10)
```

dimod/bqm/utils.pyx
arcondello: It would also let people use positional arguments, e.g.
```
cylmin(bqm, 10)
```

arcondello:
Also, these need docstrings. I took a look at how to add them to the documentation, but since they are on the `.linear` and `.quadratic` attributes, it's not so straightforward. If you add the docstrings, I can add them to the docs in a subsequent PR.

dimod/bqm/utils.pyx
randomir: For posterity, the only way to handle the (edge) case of `None` (or any other sentinel) as the actual default value is not to depend on argument value, but existence of argument; therefore the need for `**kwargs`.

But since this is such a rare edge case, I'd agree explicitness trumps it here.

randomir: FWIW, both unknown kwarg case and `default` as positional argument can be handled "manually."

arcondello:
Closes #402

PR #687, opened 2020-08-21
Quadratic iterator

shpface:


dimod/include/dimod/quadraticiterator.h
```diff
@@ -0,0 +1,218 @@
...
+                operator++();
+            }
+        }
+
+        reference operator*() const { return {u_, out_it_->first, out_it_->second}; }
+        reference operator->() const { return operator*(); }
+
+        QuadraticIterator& operator++() {
+            ++out_it_;
+            while (out_it_ == out_end() || u_ < operator*().v) {
```
amahmudwave, line 62: When there are two conditions, its better to put the more probable one first, so that the compiler skips evaluating two conditions. in this case u_ < operator*().v is more probable of evaluating to true than out_it_ == out_end() , but then again how these two checks will be ordered in the end by the compiler is not too predictable, but its better to be on the safe side for frequently used functions like this.

dimod/include/dimod/quadraticiterator.h
```diff
@@ -0,0 +1,218 @@
...
+                operator++();
+            }
+        }
+
+        reference operator*() const { return {u_, out_it_->first, out_it_->second}; }
+        reference operator->() const { return operator*(); }
+
+        QuadraticArrayIterator& operator++() {
+            ++out_it_;
+            while (out_it_ == out_end() || u_ < operator*().v) {
```
amahmudwave, line 150: Same comment as above, related to ordering of condition checks.

amahmudwave (commented):
Apart from minor nits, the PR looks okay. But do we need to add test cases along with the PR ? Some comments will help the readability a lot.

dimod/include/dimod/quadraticiterator.h
```diff
@@ -0,0 +1,218 @@
...
+        using difference_type = typename out_iterator::difference_type;
+        using value_type = QuadraticTerm<variable_type, bias_type>;
+        using pointer = value_type const*;
+        using reference = value_type;
+        using iterator_category = std::input_iterator_tag;
+
+        QuadraticArrayIterator() = default;
+        QuadraticArrayIterator(in_iterator in_it, in_iterator in_end, out_iterator out_begin1, out_iterator out_end1,
+                               variable_type u = 0)
+            : u_(u), out_begin_(out_begin1), out_end_(out_end1), in_it_(in_it), in_end_(in_end) {
```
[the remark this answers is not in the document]
shpface: Without the '1's, they will collide with the member functions. I can give those functions better names though.

dimod/include/dimod/quadraticiterator.h
shpface: The two conditions are checked in this order to prevent dereferencing the iterator if it is past the end.

arcondello:
Closing for now

PR #690, opened 2020-08-24
VartypeView

arcondello:
This also improves the performance of the views significantly by incorporating  #686  and #688.

### DIFF
--- a/dimod/core/bqm.py
+++ b/dimod/core/bqm.py
@@ -401,7 +401,10 @@ class BQM(metaclass=abc.ABCMeta):
             return self
 
         try:
-            return self._binary
+            # it may have changed vartype in the mean time
+            if self._binary.vartype is Vartype.BINARY:
+                return self._binary
+            view = self._binary
         except AttributeError:
             pass
 
@@ -436,7 +439,9 @@ class BQM(metaclass=abc.ABCMeta):
             return self
 
         try:
-            return self._spin
+            # it may have changed vartype in the mean time
+            if self._spin.vartype is Vartype.SPIN:
+                return self._spin
         except AttributeError:
             pass
 
@@ -1385,11 +1390,13 @@ class VartypeView(BQM):
         else:
             return self._bqm.spin
 
-    def change_vartype(self, *args, **kwargs):
-        # todo
-        msg = '{} can only be {}-valued'.format(type(self).__name__,
-                                                self.vartype.name)
-        raise NotImplementedError(msg)
+    def change_vartype(self, vartype, inplace=True):
+        if not inplace:
+            return self.copy().change_vartype(vartype, inplace=True)
+
+        self.vartype = as_vartype(vartype)
+
+        return self
 
     def copy(self):
         return self._bqm.change_vartype(self.vartype, inplace=False)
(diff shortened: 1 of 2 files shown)


dimod/core/bqm.py
```diff
@@ -1330,170 +1335,161 @@ def update(self, other):
...
+            return bqm.get_quadratic(u, v, default=default)
+        elif self.vartype is Vartype.SPIN and bqm.vartype is Vartype.BINARY:
+            return bqm.get_quadratic(u, v) / 4
+        elif self.vartype is Vartype.BINARY and bqm.vartype is Vartype.SPIN:
+            return 4 * self._bqm.get_quadratic(u, v)
+        else:
+            raise RuntimeError("unknown vartype combination")
 
-        bqm.offset += delta / 4
+    def iter_linear(self):
```
boothby, line 1429: I'd argue that this takes DRY too far, and results in a significant performance hit because of the repeated vartype tests and duplicated linear lookups.  To maintain the DRY and recover some performance, I'd let `iter_linear` take a `variables` argument, and replace `get_linear` with `next(iter_linear(v))[1]` or similar.  But personally, I'd just suffer the duplication.

dimod/core/bqm.py
```diff
@@ -1330,170 +1335,161 @@ def update(self, other):
...
+    def copy(self):
+        return self._bqm.change_vartype(self.vartype, inplace=False)
 
-        bqm.set_linear(v, bqm.get_linear(v) + delta / 2)
-        bqm.offset += delta / 2
+    def degree(self, *args, **kwargs):
+        return self._bqm.degree(*args, **kwargs)
 
-    def set_quadratic(self, u, v, bias):
+    def get_linear(self, v):
```
boothby, line 1407: should this function take a `default`?

dimod/core/bqm.py
```diff
@@ -1330,170 +1335,161 @@ def update(self, other):
...
+            raise RuntimeError("unknown vartype combination")
 
-        bqm.offset += delta / 4
+    def iter_linear(self):
+        for v, _ in self._bqm.iter_linear():
+            yield v, self.get_linear(v)
 
+    def iter_quadratic(self, variables=None):
+        for u, v, _ in self._bqm.iter_quadratic(variables):
+            yield u, v, self.get_quadratic(u, v)
```
boothby, line 1435: As above, I'm not stoked with this application of DRY

dimod/core/bqm.py
```diff
@@ -1330,170 +1335,161 @@ def update(self, other):
...
 
-class SpinView(VartypeView):
-    @property
-    def binary(self):
-        return self._bqm.binary
+    def relabel_variables(self, *args, **kwargs):
```
boothby, line 1437: This seems wrong because it can modify the bqm beneath the view.

```python
def relabel_variables(self, mapping, inplace=True):
    relabeled = self._bqm.relabel_variables(mapping, inplace=False)
    if inplace:
        self._bqm = relabeled
    return relabeled
```

dimod/core/bqm.py
```diff
@@ -1330,231 +1330,159 @@ def update(self, other):
...
 
     @property
     def binary(self):
-        return EitherView(self._base_bqm, BINARY)
-
-    @property
-    def spin(self):
-        return EitherView(self._base_bqm, SPIN)
+        if self.vartype is Vartype.BINARY:
+            return self
```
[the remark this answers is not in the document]
arcondello: Hmm, I was about to change this, but then I realized that it is consistent with the existing BQM implementation (which is not so costly). We could always make a view for the two vartype attributes. Need to think about it a bit more.

Issue #643, opened 2020-04-20
Add tracking to `dimod.sampleset.concatenate`


arcondello:
See https://github.com/dwavesystems/dwave-system/pull/290/files#diff-d4c234141a935a5b96a975cbc6474e08R362

Something like
```
sampleset = dimod.sampleset.concatenate(samplesets, track_parent=True, field_name='parent', labels=['sampleset1', 'sampleset2'])
```
with reasonable defaults etc.

PR #691, opened 2020-08-25
Concatenate and merge info, preserving conflicts as lists.

joseppinilla:
If there is interest in merging info when concatenating samplesets since right now all info is ignored.
This preserves conflicts by listing them, but squeezes unique values.

### DIFF
--- a/dimod/sampleset.py
+++ b/dimod/sampleset.py
@@ -237,7 +237,16 @@ def concatenate(samplesets, defaults=None):
     record = recfunctions.stack_arrays(records, defaults=defaults,
                                        asrecarray=True, usemask=False)
 
-    return SampleSet(record, variables, {}, vartype)
+    # Merge info, preserving conflicts as lists
+    info = {}
+    for k in set().union(*[s.info for s in samplesets]):
+        info[k] = []
+        for s in samplesets:
+             if k in s.info and s.info[k] not in info[k]:
+                 info[k] += [s.info[k]]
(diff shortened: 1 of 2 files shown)


arcondello:
Hmm, I am worried that the conflict resolution would lead to inconsistent/confusing results. I wonder if we could combine it with https://github.com/dwavesystems/dimod/issues/643 by always making a list, even when the values are identical?

joseppinilla:
I was considering one option flag, e.g. `squeeze_info=False` which would either behave like "list" by default or "squeeze" if `True`.

I see now how "drop" would allow compatibility with the current implementation, but I'm not sure if adding more flexibility is necessary.

randomir:
I prefer categorical option to boolean because it allows future expansion. Accepting a callable (in addition) is trivial and provides an absolute flexibility.

joseppinilla:
One reason I didn't originally implement "list" is because I had to decide whether lists are always `len(samplesets)`, e.g filled with `None`, or if they only list values of existing keys.

I think I'd go for the latter, even though it wouldn't allow tracking.

randomir:
IMO, "list" should include fields from all samplesets. We might splitting "list" into "list-all" and "list-existing"? @arcondello, thoughts?

joseppinilla:
`squeeze` may not be the right name since it's not the same behaviour as numpy squeeze, it only lists conflicts... so... "list-conflicts"?

randomir:
I don't see how `numpy.squeeze` is relevant here, but I agree it's too general. `unique` sounds much better.

Issue #519, opened 2019-08-02
Refactor roof duality code


arcondello:
Goals:
* make it compliant with the google style guide
* complete the exception refactor begun with #518 
* ideally remove the dependency on boost

PR #692, opened 2020-08-31
Add Catch2 tests

hhtong:


arcondello:
Ooph, adding 17k lines...

Maybe we can included it as a submodule?

.circleci/config.yml
```diff
@@ -54,6 +54,12 @@ jobs:
             python --version
             coverage run -m unittest discover
 
+      - run: &run-cpp-tests
```
arcondello, line 57: I think this should probably be a separate job, rather than added to the existing python jobs.

dimod/include/dimod/adjmapbqm.h
```diff
@@ -19,7 +19,7 @@
 #include <utility>
 #include <vector>
 
-#include "dimod/utils.h"
+#include "utils.h"
```
arcondello, line 22: I think we should include the full path, by google style guide https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes. For dimod's c++ code the "source directory" is `dimod/include/`

tests/cpp/tests/test_bqm.cpp
```diff
@@ -0,0 +1,69 @@
...
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+//
+// =============================================================================
+
+#include "../Catch2/single_include/catch2/catch.hpp"
+#include "../../../dimod/include/dimod/adjvectorbqm.h"
```
arcondello, line 18: I think this should just be `#include "dimod/adjvectorbqm.h"` and we should add the c++ source directory to the include path when compiling.

tests/cpp/tests/test_bqm.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include "../../../dimod/include/dimod/adjmapbqm.h"
+#include "../../../dimod/include/dimod/adjarraybqm.h"
+
+#include <iostream>
+
+using namespace dimod;
```
arcondello, line 24: No inline namespaces https://google.github.io/styleguide/cppguide.html#Namespaces

tests/cpp/tests/test_roofduality.cpp
```diff
@@ -0,0 +1,47 @@
...
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+//
+// =============================================================================
+
+#include "../Catch2/single_include/catch2/catch.hpp"
+#include "../../../dimod/roof_duality/src/fix_variables.hpp"
```
arcondello, line 18: `fix_variables_` needs to be refactored anyway, see https://github.com/dwavesystems/dimod/issues/519. Part of that will be moving the c++ code under `dimod/include/`. For now, it doesn't much matter how we handle it.

.circleci/config.yml
hhtong: I've added one linux job. Will switch to CMake and add additional jobs (for Windows, OSX, and additional compilers) in a subsequent PR.

PR #696, opened 2020-09-04
Add Adj*BQM::linear

arcondello:
Gives slightly nicer access to the linear biases in c++ BQMs. E.g.
```
bqm.linear(v) += 1;
```
rather than
```
bqm.set_linear(v, bqm_get_linear(v) + 1);
```

I would love to have `bqm.quadratic(u, v)` as well, but because we keep two copies of the bias around for quadratic, we can't just return a reference like we do with `.linear`. Open to suggestions.


shpface:
LGTM

shpface (approved):
LGTM

PR #698, opened 2020-09-09
fixes #697, fixed relabel variables as integers with no inplace

### LINKED ISSUE #697: bqm relabel variables as integers does not work when inplace=False
Wrong method is called when `inplace=False`

alexzucca90:
Closes #697

### DIFF
--- a/dimod/core/bqm.py
+++ b/dimod/core/bqm.py
@@ -860,7 +860,7 @@ class BQM(metaclass=abc.ABCMeta):
 
         """
         if not inplace:
-            return self.copy().relabel_variables(inplace=True)
+            return self.copy().relabel_variables_as_integers(inplace=True)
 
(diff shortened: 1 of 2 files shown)


arcondello (approved):

arcondello:
Actually, can we add a test that would have caught it?

alexzucca90:
fixed the test, as it was awkwardly using `inplace=True` for an `inplace=False` operation. Should work now

arcondello:
Now I am worried that we don't test `inplace=True`

alexzucca90:
well, `inplace=True` is called inside `inplace=False` so it should actually raise the error

arcondello:
That is true, although an implementation detail. We should have tests that are still valuable even if the implementation changes.

arcondello (approved):
Good catch, thanks

PR #702, opened 2020-09-15
Add DiscreteQuadraticModel class

arcondello:


dimod/discrete/discrete_quadratic_model.py
```diff
@@ -0,0 +1,367 @@
...
+    def index(self, v):
+        # todo: support start and end like list.index
+        if v not in self:
+            raise ValueError('unknown variable {!r}'.format(v))
+        return self._label_to_idx.get(v, v)
+
+
+class DiscreteQuadraticModel:
+    """A discrete quadratic model is a polynomial over discrete variables with
+    terms all of degree two or less.
```
JoelPasvolsky, line 109: Encodes a discrete quadratic model.

Discrete quadratic model is a polynomial over...

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -0,0 +1,367 @@
...
+
+    Examples:
+
+        This example constructs a map coloring with Canadian provinces. To
+        solve the problem we penalize adjacent provinces having the same color.
```
JoelPasvolsky, line 114: Maybe instead of glossing an explanation, provide a link to https://docs.ocean.dwavesys.com/en/stable/examples/map_coloring.html?

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -0,0 +1,367 @@
...
+
+    def get_quadratic(self, u, v, array=False):
+        """The biases associated with the interaction between `u` and `v`.
+
+        Args:
+            u: A variable in the discrete quadratic model.
+
+            v: A variable in the discrete quadratic model.
+
+            array (bool, optional, default=False): If True a dense array is
```
JoelPasvolsky, line 252: If True,

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -0,0 +1,367 @@
...
+            returned rather than a dict.
+
+        Returns:
+            The quadratic biases. If `array=False`, then a dictionary of the
+            form `{case_u, case_v: bias, ...}` is returned. If `array=True`,
+            then a :meth:`~DiscreteQuadraticModel.num_cases(u)` by
+            :meth:`~DiscreteQuadraticModel.num_cases(v)` numpy array is
+            returned.
+
```
JoelPasvolsky, line 261: "If `array=False`, returns a dictionary..."
"If `array=True`, returns a :meth:`~DiscreteQuadraticModel.num_cases(u)`..."

docs/reference/dqm.rst
```diff
@@ -0,0 +1,41 @@
+.. _dqm:
+
+=========================
+Discrete Quadratic Models
+=========================
+
+:term:`Discrete quadratic model`\ s (DQMs) are described under
+:std:doc:`Discrete Quadratic Models <oceandocs:concepts/dqm>`.
```
JoelPasvolsky, line 8: This should be a short description with the embedded link.

JoelPasvolsky (commented):
Why do we use "case" for what looks to me from scanning this PR to be the values of the variables?

docs/reference/dqm.rst
arcondello: I copied this exactly from https://github.com/dwavesystems/dimod/blob/master/docs/reference/bqm.rst

JoelPasvolsky: Yes, sorry, we agreed that the repos would point to the SDK for concepts.

arcondello:
We use "case" rather than "values" because in general the cases of a variable are just labels, whereas "values" imply that they contribute to the energy. When we're representing an integer quadratic model using a discrete quadratic model, they become equivalent.

dimod/discrete/discrete_quadratic_model.py
arcondello: I am a bit worried that it might be confusing to link the two... Maybe a wikipedia link? We'll definitely want a map coloring example using DQM as well, possibly in the same repo

JoelPasvolsky:
@arcondello, approaching this without yet looking at the implementation details (trying to experience it as a new users to see where difficulties might arise), I think of a DQM as y = axy with x in ["a", "b"] then for a solution to select x="a" should mean E(y|x="a") < E(y|x="b"). So what do you mean by the cases not contributing to energy? Do you mean that case is translated to a numerical value that represents its contribution to energy?

arcondello:
Imagine that I wanted to encode the one-variable integer quadratic model `E(i) = 3i, -2 <= i <= 2, i != 0` as a discrete quadratic model. To do that I would do
```
dqm = dimod.DiscreteQuadraticModel()
ivar = dqm.add_variable(4, label='i')
dqm.set_linear(ivar, [-6, -3, 3, 6])
```
the point being that the indices of `[-6, -3, 3, 6]` are the cases. They are just indices/labels but they are not the values themselves.

JoelPasvolsky:
Thanks, last question I hope: so is this not an artifact of our implementation, that we have users set biases on the low-level binary nodes? Could we not equivalently let users map numerical values to each variable label and label-label interaction, then set only a single bias per variable and per interaction between variables? Would that not make the labels be the values of a discrete quadratic model?

arcondello:
That would work for an integer quadratic model, which is only a special case of discrete quadratic model. And it is not generally a good idea to use DQMs to represent IQMs, since solver that natively solver IQMs can exploit the specific properties of that solver.

PR #706, opened 2020-09-19
Parallelized AdjVectorBQM constructor using dense matrix

amahmudwave:
Added a parallelized AdjVectorBQM constructor from dense matrix.
This constructor requires the source matrix to be not read-only.

### DIFF
--- a/dimod/include/dimod/adjvectorbqm.h
+++ b/dimod/include/dimod/adjvectorbqm.h
@@ -16,6 +16,7 @@
 #define DIMOD_ADJVECTORBQM_H_
 
 #include <algorithm>
+#include <stdio.h>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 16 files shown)


dimod/include/dimod/adjvectorbqm.h
```diff
@@ -83,6 +84,96 @@ class AdjVectorBQM {
...
+     * Construct a BQM from a dense array. This constructor is parallelized
+     * and temporarily zeroes out the diagonal of the dense array but restores
+     * it back.
+     *
+     * @param dense An array containing the biases. Assumed to contain
+     *     `num_variables`^2 elements. The upper and lower triangle are summed.
+     * @param num_variables The number of variables.
+     */
+    template <class B2>
+    AdjVectorBQM(B2 dense[], size_type num_variables,
```
amahmudwave, line 97: To avoid check in the innermost loop I need to zero out the diagonal entries of the source matrix temporarily and restore it back. I restore it back to maintain consistency with the existing constructor that uses a read-only dense matrix. Making it read-write can only lead to an issue when multiple threads are accessing the same matrix in different functions. In this function I access it using multiple threads but I handle the synchronization myself so it is okay.

arcondello:
This may be a silly question, but is there a way to make the number of threads used by openmp configurable at runtime? I ask because we ran into issues before running openmp in docker images (on circle-ci specifically).

amahmudwave:
> This may be a silly question, but is there a way to make the number of threads used by openmp configurable at runtime? I ask because we ran into issues before running openmp in docker images (on circle-ci specifically).

Yes it is configurable but I intentionally tried not rely on number of threads because making the execution path dependent on the number of threads would make it prone to errors, lets say I query maximum number of threads, openmp will generally return number of cores * SMT (hyperthreading) value  or may be less , I use that number to allocate something and then in my loop or parallel region I get a different number, it could be that I may end up accessing wrong indices.

I tried to make sure each parallel region independent of the number of threads that executes another parallel region. that is ideally one could comment out all the openmp pragmas and the code should still give the same result.

Issue #599, opened 2020-02-10
Transparent BQM serialization to binary stream


randomir:
For cleaner serialization of BQMs we should have `as_file()` on all BQM objects that support (or are supported by) the `FileView`.

Dimod's FileView implementation should be considered an implementation detail, not something users should know about (just like Cloud Client's implementation of FileView is an implementation detail).

Alternatively, `to_serializable(.., binary=True)` could return a file-like object.

Both variants can come as context managers as well.

PR #707, opened 2020-09-19
DQM Serialization

arcondello:
Also adds intermediate representation in `DQM.to_numpy_vectors` and `.from_numpy_vectors`.

Also uses the API suggested in #599.

### DIFF
--- a/dimod/discrete/cydiscrete_quadratic_model.pyx
+++ b/dimod/discrete/cydiscrete_quadratic_model.pyx
@@ -17,14 +17,10 @@ import collections.abc as abc
 cimport cython
 
 from cython.operator cimport preincrement as inc, dereference as deref
-from libcpp.algorithm cimport lower_bound, upper_bound
-from libcpp.pair cimport pair
+from libcpp.algorithm cimport lower_bound, sort
+from libcpp.unordered_set cimport unordered_set
 
 import numpy as np
-
-# this is the same as in dimod/utils.h but cython cannot recognize that one
-cdef bint comp_v(pair[CaseIndex, Bias] ub, Bias v):
-    return ub.first < v
 
 
 cdef class cyDiscreteQuadraticModel:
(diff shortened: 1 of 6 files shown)


tests/test_discrete_quadratic_model.py
```diff
@@ -13,12 +13,44 @@
...
+import tempfile
 import unittest
 
 import dimod
 import numpy as np
 
 
+# will want to migrate this to generators at some point
+def gnp_random_dqm(num_variables, num_cases, p, p_case, seed=None):
+    np.random.seed(seed)
```
randomir, line 27: To limit function side-effects, you can use [`numpy.random.default_rng`](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.default_rng).

tests/test_discrete_quadratic_model.py
```diff
@@ -107,6 +139,96 @@ def test_two_variable_labelled(self):
...
+            new = dimod.DQM.from_file(tf)
+
+        self.assertEqual(new.num_variables(), dqm.num_variables())
+        self.assertEqual(new.num_cases(), dqm.num_cases())
+        self.assertEqual(new.variables, dqm.variables)
+        for v in dqm.variables:
+            self.assertEqual(new.num_cases(v), dqm.num_cases(v))
+            np.testing.assert_array_equal(new.get_linear(v),
+                                          dqm.get_linear(v))
+        self.assertEqual(new.adj, dqm.adj)
```
randomir, line 175: `DQM.__eq__`?

dimod/discrete/cydiscrete_quadratic_model.pyx
```diff
@@ -109,6 +105,142 @@ cdef class cyDiscreteQuadraticModel:
...
+        # if necessary
+        index_dtype = np.result_type(starts, irow, icol, np.uint16)
+
+        starts = np.asarray(starts, dtype=index_dtype)
+        ldata = np.asarray(ldata, dtype=obj.dtype)
+        irow = np.asarray(irow, dtype=index_dtype)
+        icol = np.asarray(icol, dtype=index_dtype)
+        qdata = np.asarray(qdata, dtype=obj.dtype)
+
+        if index_dtype == np.uint16:
```
randomir, line 227: No `(u)int8`?

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -383,5 +521,210 @@ def set_quadratic_case(self, u, u_case, v, v_case, bias):
...
+
+        if not index_labeled:
+            file.write(VariablesSection(self.variables).dumps())
+
+        file.seek(0)
+
+        return file
+
+    def to_numpy_vectors(self, return_labels=False):
+        """Convert the DQM to five numpy vectors.
```
randomir, line 683: Just a thought: maybe it's time to start returning `namedtuple`.

randomir (approved):
LGTM

dimod/discrete/cydiscrete_quadratic_model.pyx
arcondello: A few lines above I coerce it to be at least `np.uint16`

tests/test_discrete_quadratic_model.py
arcondello: Totally, I was in a hurry and this one lives only in testing. But good point, might as well to it correctly :smile:

tests/test_discrete_quadratic_model.py
arcondello: The problem comes with handling differently ordered DQMs. Right now `bqm == other_bqm` returns `True` even when they have a different variable order. I would probably want to be consistent with BQM, but I wanted to think about it a bit more first. In either case this (mostly) tests the stronger condition that they are equal and have identical underlying values.

dimod/discrete/discrete_quadratic_model.py
arcondello: Yeah, I was trying to maintain similarity to `BQM.to_numpy_vectors`, but probably practicality trumps similarity. `namedtuple` seems cleaner. Maybe I'll add an option to return a "flat" `namedtuple` from the BQM ones as well as an argument and put up a deprecation notice.

dimod/discrete/cydiscrete_quadratic_model.pyx
randomir: Missed that. :+1:

dimod/discrete/discrete_quadratic_model.py
randomir: No need, `namedtuple` is `tuple`'s subclass, so existing code would not break. At least destructuring will work. Doubt anyone checks exact rtype.

arcondello: I was more referring to the fact that I returned a nested tuple `(case_starts, linear_biases, (irow, icol, qdata))`. In either case, I can just return a nested namedtuple as well, which is what I will do. I think if I was writing it from scratch I would just return a flat one but :shrug:

randomir: Oh, I see. Well, we'll fix it in Ocean 5. :smile:

PR #709, opened 2020-09-21
Add BQM.to_file and BQM.from_file for supported BQM types

### LINKED ISSUE #599: Transparent BQM serialization to binary stream

arcondello:
Closes #599


randomir (approved):
LGTM

PR #713, opened 2020-09-23
DISCRETE vartype in SampleSet

arcondello:


tests/test_sampleset.py
```diff
@@ -257,6 +258,48 @@ def test_from_bqm_with_sorting(self):
         np.testing.assert_array_equal(np.flip(raw, 1), samples.record.sample)
 
 
+class TestDiscreteSampleSet(unittest.TestCase):
```
randomir, line 261: I would add tests for `from_samples_bqm` on a DQM.

randomir (approved):
Elegant under current constraints.

tests/test_sampleset.py
arcondello: I'll add some tests in a followup PR with when I address https://github.com/dwavesystems/dimod/blob/3402a03a337b64f4618fc2fbc44a3c5e58f60e5d/dimod/discrete/discrete_quadratic_model.py#L252 and https://github.com/dwavesystems/dimod/blob/3402a03a337b64f4618fc2fbc44a3c5e58f60e5d/dimod/discrete/discrete_quadratic_model.py#L104

PR #715, opened 2020-09-24
Parallelized AdjMapBQM constructor from dense matrix

amahmudwave:
I could only get 5X speedup with 10 cores. Any suggestion is welcome. 

Note : 

I tried not using a buffer since the size of the buffer does not help us preallocate memory in the map like in case of AdjVectorBQM, but reading the dense matrix and handling the reallocation of the map inside the same for loop nullifies the benefits of cache blocking and it ends up being slower than the current implementation in the PR. Cache blocking provides around 25% improvement. And thought of keeping the feature since it is being used in AdjVectorBQM.

Using emplace_hint when inserting into the map instead of the way I used gives a tiny benefit of 8% or so, but did not opt for that. 

I have left the makefiles unchanged, to test with parallelism we should include -fopenmp , should that be changed ? In that case are there other places where I need to insert a similar flag ( I mean for the python and cython test cases).

@arcondello

### DIFF
--- a/dimod/include/dimod/adjmapbqm.h
+++ b/dimod/include/dimod/adjmapbqm.h
@@ -83,6 +83,102 @@ class AdjMapBQM {
         }
     }
 
+    /**
+     * Construct a BQM from a dense array. This constructor is parallelized
+     * and temporarily zeroes out the diagonal of the dense array but restores
+     * it back.
+     *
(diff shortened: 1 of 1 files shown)


arcondello:
Sorry, can you remake this against master?

PR #718, opened 2020-09-24
Parallelized AdjMapBQM constructor from dense array

amahmudwave:
Parallelized AdjMapBQM constructor from dense array. 
Added testing of cpp files with openmp.

PR #722, opened 2020-09-25
Parallelized AdjArrayBQM constructor.

amahmudwave:
Parallelized AdjArrayBQM constructor. There is a bottleneck in the parallel code,
which is the resizing of the vector representing the adjacency matrix of the bqm.
As long as a resize function is not available that does not do initialization during
resizing, we can only improve it through custom allocators, which we likely want
to avoid.

### DIFF
--- a/dimod/include/dimod/adjarraybqm.h
+++ b/dimod/include/dimod/adjarraybqm.h
@@ -89,6 +89,131 @@ class AdjArrayBQM {
         }
     }
 
+    /**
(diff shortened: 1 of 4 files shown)


amahmudwave:
@arcondello This is ready for review. Only https://github.com/dwavesystems/dimod/pull/722/commits/67346e9a0eb322090caf290bed1d0cc3c8545893 needs to be reviewed as the other commit is under review in a separate PR.

arcondello:
I merged #718 , does this need to be rebased?

PR #725, opened 2020-09-30
Parallelized AdjArrayBQM constructor from dense array

amahmudwave:
Parallelized AdjArrayBQM constructor from dense array. Threads in parallel first calculate the size
of the resulting array and indices where they should start writing to. Then in the second phase they
write into the BQM. The bottleneck in this method is calling the resize function on the vector of the
BQM as resize() also initializes the data, this cannot be circumvented without changing data structure
or using custom vectors/allocators. It is mentioned as a TODO if improvements to standard 
vector::resize function is made.

### DIFF
--- a/dimod/include/dimod/adjarraybqm.h
+++ b/dimod/include/dimod/adjarraybqm.h
@@ -89,6 +89,131 @@ class AdjArrayBQM {
         }
     }
 
+    /**
+     * Construct a BQM from a dense array. This constructor is parallelized
+     * and temporarily zeroes out the diagonal of the dense array but restores
+     * it back.
+     *
+     * @param dense An array containing the biases. Assumed to contain
+     *     `num_variables`^2 elements. The upper and lower triangle are summed.
+     * @param num_variables The number of variables.
+     */
+    template <class B2>
+    AdjArrayBQM(B2 dense[], size_type num_variables, bool ignore_diagonal = false) {
+        // we know how big our linear is going to be.
+        invars.resize(num_variables);
+
+        // Aligned memory is to avoid false sharing between threads.
+        size_type* counters_cumsum = (size_type*)utils::aligned_calloc(num_variables, sizeof(size_type));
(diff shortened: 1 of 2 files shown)


dimod/include/dimod/adjarraybqm.h
```diff
@@ -89,6 +89,131 @@ class AdjArrayBQM {
...
+        // indices in the bqm for inserting elements for each row of the dense matrix. This
+        // could be parallelized but has diminishing returns.
+        size_type sum_counters = 0;
+        for (size_type u = 0; u < num_variables; ++u) {
+            size_type prev_counter = counters_cumsum[u];
+            counters_cumsum[u] = sum_counters;
+            sum_counters += prev_counter;
+        }
+
+        // TODO : This is the bottleneck for moderately dense input arrays.
```
amahmudwave, line 161: @arcondello  @shpface , I could not find a straightforward way of removing this bottleneck. Tried some approaches which work for vectors of primitive elements but not pairs, do you have any idea ? Note I cannot reserve and pushback in parallel in the code below. This is a problem for very large dense matrices. Otherwise fine. With this bottleneck we get 10X speedup on 10 cores for dense matrices. Without would give us (20-40X) speedup though.

amahmudwave:
@shpface  @arcondello This is ready for review.

dimod/include/dimod/utils.h
```diff
@@ -27,6 +28,58 @@ bool comp_v(std::pair<V, B> ub, V v) {
     return ub.first < v;
 }
 
+// The aligned_malloc and aligned_free functions were written with the help of this link:
+// https://stackoverflow.com/questions/38088732/explanation-to-aligned-malloc-implementation
```
amahmudwave, line 32: @arcondello

arcondello:
No longer relevant, see https://github.com/dwavesystems/dimod/pull/788

PR #727, opened 2020-10-05
Deprecate ``compressed`` in DQM

JoelPasvolsky:
Part of https://github.com/dwavesystems/dwave-system/pull/356

### DIFF
--- a/dimod/discrete/discrete_quadratic_model.py
+++ b/dimod/discrete/discrete_quadratic_model.py
@@ -632,7 +632,7 @@ class DiscreteQuadraticModel:
             self.variables.index(v), v_case,
             bias)
 
-    def _to_file_numpy(self, file, compressed):
+    def _to_file_numpy(self, file, compress):
         # the biases etc, saved using numpy
 
         # we'd like to just let numpy handle the header etc, but it doesn't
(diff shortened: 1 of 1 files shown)


JoelPasvolsky:
Failed `test-osx-py35` looks unrelated to this PR, it is using `dqm.to_file()` but without parameters:
`buff = f.read() != buff2 = bytearray(len(buff))`  for `dqm.to_file()`

arcondello (approved):

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -663,14 +663,17 @@ def _to_file_numpy(self, file, compressed):
         file.write(np.dtype('<u4').type(end - start).tobytes())
         file.seek(end)
 
-    def to_file(self, compressed=False, ignore_labels=False,
+    def to_file(self, compress=False, compressed=None, ignore_labels=False,
```
randomir, line 666: To preserve function signature for positional args, it's better to just append the new argument. In theory, this change breaks the public API, and would require a major version increment :grimacing:

randomir (commented):
LGTM

dimod/discrete/discrete_quadratic_model.py
JoelPasvolsky: Sorry, I figured it's unlikely anyone was already using this, so I preferred the least confusing signature and to ensure a deprecation notice if there is someone using it already.

randomir: No, absolutely, this was just an FYI comment. I'm pretty sure this change won't break anyone's code.

PR #733, opened 2020-10-15
Support python 3.9 in CI

arcondello:
I don't think this will work since numpy does not yet have a 3.9 wheel. But might as well do the work in advance


arcondello:
Going to wait to merge this until numpy deploys 3.9

PR #734, opened 2020-10-15
Use c++11 with MSVC

### LINKED ISSUE #732: Installing dimond fails due to boost errors
**Description**
I tried to install via pip but I got the following error even though boost is installed and BOOST_ROOT environment variable is set.
`dimod/roof_duality/src/fix_variables.cpp(30): fatal error C1083: Cannot open include file: 'boost/graph/graph_traits.hpp': No such file or directory`
Then I tried to install via source, but I got the same error.
This time I copied boost\boost_1_62_0\boost to dimod\dimod\roof_duality\src and it passed that stage, and after that it encounters a huge amount of the same syntax error, a sample of which you can see below:
image

How can I get past all these errors? 

**Environment**
 - OS: Windows 10 x64
 - Python version: 3.9.0

arcondello:
Closes #732

Occurs because the roof duality code is not supported in c++17

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -69,7 +69,7 @@ else:
 ext = '.pyx' if USE_CYTHON else '.cpp'
 
 extra_compile_args = {
-    'msvc': ['/std:c++latest', '/EHsc'],
+    'msvc': ['/std:c11', '/EHsc'],
     'unix': ['-std=c++11'],
 }
 


setup.py
```diff
@@ -69,7 +69,7 @@
 ext = '.pyx' if USE_CYTHON else '.cpp'
 
 extra_compile_args = {
-    'msvc': ['/std:c++latest', '/EHsc'],
+    'msvc': ['/std:c11', '/EHsc'],
```
randomir, line 72: I believe this is ignored when compiling C++ code (`c11` != `c++11`). Modern MSVC will [default to `/std:c++14`](https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=vs-2019) for C++ code.

arcondello: Got it, so the fix came from removing `c++latest` rather than adding `c11`. Thanks for the sanity check! I will fix

PR #736, opened 2020-10-20
Use default c++ with MSVC

arcondello:
See https://github.com/dwavesystems/dimod/pull/734#discussion_r506533860

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -69,7 +69,7 @@ else:
 ext = '.pyx' if USE_CYTHON else '.cpp'
 
 extra_compile_args = {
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -69,7 +69,7 @@
 ext = '.pyx' if USE_CYTHON else '.cpp'
 
 extra_compile_args = {
-    'msvc': ['/std:c11', '/EHsc'],
+    'msvc': ['/EHsc'],
```
randomir, line 72: Just to make sure a recent enough version of MSVC is used, I would keep `/std:c++14`.

PR #741, opened 2020-10-26
Remove release version from docs

arcondello:
See https://github.com/dwavesystems/dwave-ocean-sdk/issues/56


JoelPasvolsky (approved):
Thanks! I had made the list in https://github.com/dwavesystems/dwave-ocean-sdk/issues/56 to track my doing the work but tracking your doing it is much nicer.

PR #742, opened 2020-10-27
Add .clang-format and apply to dimod/include/*

arcondello:
See https://github.com/dwavesystems/dwave-ocean-sdk/pull/90 and https://github.com/dwavesystems/dwave-ocean-sdk/pull/103

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,11 @@
+---
+Language: Cpp
+BasedOnStyle: Google
+
+ColumnLimit: 80
+NamespaceIndentation: Inner
+
+# Scaled by a factor of 2 such that the base indent is 4
+AccessModifierOffset: -3
+IndentWidth: 4
+...
--- a/.clang-format
+++ b/.clang-format
@@ -8,4 +8,6 @@ NamespaceIndentation: Inner
 # Scaled by a factor of 2 such that the base indent is 4
 AccessModifierOffset: -3
 IndentWidth: 4
+ContinuationIndentWidth: 8
+ConstructorInitializerIndentWidth: 8
 ...
--- a/dimod/include/dimod/adjmapbqm.h
+++ b/dimod/include/dimod/adjmapbqm.h
@@ -15,6 +15,7 @@
 #ifndef DIMOD_ADJMAPBQM_H_
 #define DIMOD_ADJMAPBQM_H_
 
+#include <algorithm>
 #include <map>
 #include <utility>
 #include <vector>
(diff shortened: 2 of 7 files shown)


.clang-format
```diff
@@ -0,0 +1,11 @@
+---
+Language: Cpp
+BasedOnStyle: Google
+
+ColumnLimit: 80
+NamespaceIndentation: Inner
+
+# Scaled by a factor of 2 such that the base indent is 4
+AccessModifierOffset: -3
```
randomir, line 9: Why not -2?

dimod/include/dimod/adjvectorbqm.h
```diff
@@ -96,159 +98,156 @@ class AdjVectorBQM {
...
+    std::pair<outvars_iterator, outvars_iterator> neighborhood(
+        variable_type u) {
         assert(u >= 0 && u < adj.size());
         return std::make_pair(adj[u].first.begin(), adj[u].first.end());
     }
 
-    std::pair<const_outvars_iterator, const_outvars_iterator>
-    neighborhood(variable_type u) const {
+    std::pair<const_outvars_iterator, const_outvars_iterator> neighborhood(
+        variable_type u) const {
```
randomir, line 227: This feels like it should be indented 8 spaces. Not sure is that due to `ContinuationIndentWidth` or `ConstructorInitializerIndentWidth` not scaled with `IndentWidth` in dwave-ocean-sdk#90.

.clang-format
arcondello: Discussion: https://github.com/dwavesystems/dwave-ocean-sdk/pull/103#discussion_r513567538

arcondello:
Ok, scaled the indent. @randomir , can you take another look? Figure it's better to iterate here than on dwavesystems/dwave-ocean-sdk#103

dimod/include/dimod/adjmapbqm.h
```diff
@@ -95,108 +97,110 @@ class AdjMapBQM {
...
-      {
-        // Zero out the diagonal to avoid expensive checks inside innermost
-        // loop in the code for reading the matrix. The diagonal will be
-        // restored so a backup copy is saved.
-        #pragma omp for schedule(static)
-        for (size_type v = 0; v < num_variables; ++v) {
-          dense_diagonal[v] = dense[v * (num_variables + 1)];
-          dense[v * (num_variables + 1)] = 0;
+        if (!ignore_diagonal) {
+#pragma omp parallel for
```
arcondello, line 107: Annoys me that it cannot handle `#pragma omp parallel for` etc. There [are solutions](https://stackoverflow.com/questions/31353748/how-could-i-indent-c-pragma-using-clang-format) but probably not common enough to require.

PR #743, opened 2020-10-27
Add append_data_vectors() to SampleSet

### LINKED ISSUE #738: SampleSet.append_data_vectors or similar
Would wrap NumPy's [append_fields](https://numpy.org/devdocs/user/basics.rec.html#numpy.lib.recfunctions.append_fields) with an API similar to [append_variables](https://docs.ocean.dwavesys.com/en/stable/docs_dimod/reference/generated/dimod.SampleSet.append_variables.html#dimod.SampleSet.append_variables) (returns a new sample set).

hhtong:
- Helps to improve [this PR](https://github.com/dwavesystems/dwave-system/pull/359)

Closes #738

### DIFF
--- a/dimod/sampleset.py
+++ b/dimod/sampleset.py
@@ -39,8 +39,69 @@ from dimod.variables import Variables, iter_deserialize_variables
 from dimod.vartypes import as_vartype, Vartype, DISCRETE
 from dimod.views.samples import SampleView, SamplesArray
 
-__all__ = 'as_samples', 'concatenate', 'SampleSet'
+__all__ = ['append_data_vectors', 'as_samples', 'concatenate', 'SampleSet']
 
+def append_data_vectors(sampleset, **vectors):
+    """Create a new :obj:`.SampleSet` with additional fields in
+    :attr:`SampleSet.record`.
+
+    Args:
+        sampleset (:obj:`.SampleSet`):
+            :obj:`.SampleSet` to build from.
+
+        **vectors (list):
+            Per-sample data to be appended to :attr:`SampleSet.record`. Each
+            keyword is a new field name and each keyword parameter should be a
+            list of scalar values or numpy arrays (lists and tuples will be
+            converted to numpy arrays).
+
+    Returns:
+        :obj:`.SampleSet`: SampleSet
+
+    Examples:
+        The following example appends a field of lists to :attr:`SampleSet.record`.
+
+        >>> sampleset = dimod.SampleSet.from_samples([[-1,  1], [-1,  1]], energy=[-1.4, -1.4], vartype='SPIN')
+        >>> print(sampleset)
+           0  1 energy num_oc.
+        0 -1 +1   -1.4       1
+        1 -1 +1   -1.4       1
+        ['SPIN', 2 rows, 2 samples, 2 variables]
+
+        >>> sampleset = dimod.append_data_vectors(sampleset, new=[[0, 1], [1, 2]])
+        >>> print(sampleset)
+           0  1 energy num_oc. new
+        0 -1 +1   -1.4       1 [0 1]
+        1 -1 +1   -1.4       1 [1 2]
+        ['SPIN', 2 rows, 2 samples, 2 variables]
+
+        >>> print(sampleset.record.dtype)
+        (numpy.record, [('sample', 'i1', (2,)), ('energy', '<f8'), ('num_occurrences', '<i8'), ('new', '<i8', (2,))])
+    
+    """
+    record = sampleset.record
+
+    for name, vector in vectors.items():
+        if len(vector) != len(record.energy):
+            raise ValueError("Length of vector {} must be equal to number of samples.".format(name))
+
+        try:
+            vector = np.asarray(vector)
+
+            if vector.ndim == 1:
+                record = recfunctions.append_fields(record, name, vector, usemask=False, asrecarray=True)
+            else:
+                # np's append_fields cannot append a vector with a shape that
+                # doesn't match the base array's, so appending non-scalar data
+                # requires a workaround
+                dtype = np.dtype([(name, vector[0].dtype, vector[0].shape)]) 
+                new_arr = recfunctions.unstructured_to_structured(vector, dtype=dtype)
+                record = recfunctions.merge_arrays((record, new_arr), flatten=True, asrecarray=True)              
+        
+        except (TypeError, AttributeError):
+            raise ValueError("Field value type not supported.")
+
+    return SampleSet(record, sampleset.variables, sampleset.info, sampleset.vartype)
 
 def as_samples(samples_like, dtype=None, copy=False, order='C'):
     """Convert a samples_like object to a NumPy array and list of labels.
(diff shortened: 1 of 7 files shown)


JoelPasvolsky:
Are you going to be adding this method to https://docs.ocean.dwavesys.com/en/stable/docs_dimod/reference/sampleset.html#methods? I don't see an update to https://github.com/dwavesystems/dimod/blob/master/docs/reference/sampleset.rst.

hhtong:
> Are you going to be adding this method to https://docs.ocean.dwavesys.com/en/stable/docs_dimod/reference/sampleset.html#methods? I don't see an update to https://github.com/dwavesystems/dimod/blob/master/docs/reference/sampleset.rst.

Yes, thank you, I forgot. Will do that now!

dimod/sampleset.py
```diff
@@ -1391,6 +1391,90 @@ def slice(self, *slice_args, **kwargs):
...
+        
+        """
+        self.resolve()
+
+        record = self._record
+
+        for name, vector in vectors.items():
+            if name in record.dtype.names:
+                raise ValueError("Field '{}' occurs more than once".format(name))
+
```
arcondello, line 1448: I think this can be simplified with `vector = np.asarray(vector)`, which would give you access to `.shape` and would check that it can be cast to a numpy array etc etc.

hhtong: Hm, I actually do end up doing that on line 1459. The checks above just take other possible errors into account.

dimod/sampleset.py
```diff
@@ -1391,6 +1391,90 @@ def slice(self, *slice_args, **kwargs):
...
+            if len(vector) is not len(record.energy):
```
arcondello: What is the error raised by `recfunctions.append_fields`? If it is sufficiently informative it might make sense to just propagate it rather than having our own extra fail case.

dimod/sampleset.py
arcondello: Right, I guess I find `arr.ndim == 1` more satisfying than `np.isscalar(vector[0])`

arcondello: But I cannot think of an error case where `np.isscalar(vector[0])` would fail off hand so maybe it's fine.

dimod/sampleset.py
```diff
@@ -1391,6 +1391,90 @@ def slice(self, *slice_args, **kwargs):
...
+            else:
+                # np's append_fields cannot append a vector with a shape that
+                # doesn't match the base array's, so appending non-scalar data
+                # requires a workaround
+                try:
+                    narr = np.asarray(vector)
+                    dtype = np.dtype([(name, narr[0].dtype, narr[0].shape)])
+
+                    # creating a new np array to merge
+                    new_arr = np.empty(len(vector), dtype=dtype)
```
arcondello, line 1463: I suspect that `recfunctions.unstructured_to_structured(vector, dtype=dtype)` will be more performant

dimod/sampleset.py
hhtong: Actually, `recfunctions.append_fields` does not raise an error when the length of the vector is incorrect. It has a parameter `usemask` and `fill_value`, which will provide default values for any incomplete data. I had assumed that we wouldn't need this functionality, so I raised an error instead.

arcondello: makes sense. Ok, works for me!

dimod/sampleset.py
hhtong: On second thought, I agree and have made the change 👌

dimod/sampleset.py
```diff
@@ -1391,6 +1391,81 @@ def slice(self, *slice_args, **kwargs):
...
+        :attr:`SampleSet.record`.
+
+        Args:
+            vectors (dict):
+                Each key is a new field name and each value is a list of 
+                per-sample data to be appended to :attr:`SampleSet.record`. 
+                Per-sample data should be scalar or numpy arrays (lists 
+                and tuples will be converted to numpy arrays).
+
+            inplace (bool, optional, default=True):
```
arcondello, line 1405: I would _not_ support this parameter, like in [.append_variables](https://github.com/dwavesystems/dimod/blob/4941d2301aca0dc2c61b586605d0d3c69e2a7320/dimod/sampleset.py#L1123). The sampleset is a fairly lightweight wrapper around the record array and IMO remaking the record should also just remake the entire object.

arcondello: This would also allow the user to pass in the vectors as kwargs. Which might be a nicer api. E.g. `new = sampleset.append_vectors(test=[1,2,3], another=[3, 4, 5])`.

I personally don't have much of an opinion between passing in dict or separate kwargs, but thought I would raise it as an option. Keeping the dict format is more future proof (easy to add unrelated kwargs later).

hhtong: Personally I like the idea of separate kwargs and agree that it's a nicer api. It doesn't seem very likely that we would need to add any unrelated kwargs in the future

randomir: On that note (or notes), I wanted to suggest we rename this and/or add an update method, e.g. `update_vectors`, with a `dict.update`-like semantics.

`update_vectors` would always be inplace, and it would accept vectors as kwargs.

We can have `append_vectors` in parallel, and here I would leave the `inplace`, to be consistent with other `SampleSet` methods. Performance argument doesn't hold (if it's roughly the same perf), immutability by default if conceptually lovely, but that's not how SampleSet/BQM are currently designed. OTOH, `append_variables` doesn't have inplace either. :thinking:

dimod/sampleset.py
```diff
@@ -1391,6 +1391,81 @@ def slice(self, *slice_args, **kwargs):
...
+            >>> print(sampleset.record.dtype)
+            (numpy.record, [('sample', 'i1', (2,)), ('energy', '<f8'), ('num_occurrences', '<i8'), ('new_col', '<i8', (2,))])
+        
+        """
+        self.resolve()
+
+        record = self._record
+
+        for name, vector in vectors.items():
+            if len(vector) is not len(record.energy):
```
randomir, line 1443: **Never** compare integers with `is`:
```
>>> a = 257
>>> b = 257
>>> a is b
False
```
Use `is` only for identity check (pointer address compare in cpython).

(*) Reason why it works for integers in [-5, 256] range (IIRC), is because that narrow range of ints is interned.

randomir (changes requested):
LGTM in the first pass, except the `is` bug.

dimod/sampleset.py
arcondello: I think that because we cannot update the structure of the record in-place, we need to follow that behavior with the sampleset. Another approach would be to make `append_variables` and `append_vectors` functions rather than methods. 

Further, in the current implementation, it is not really safe to replace the `.record` object altogether (I can work up a failing example if you like).

hhtong: I think maybe having `append_variables` and `append_vectors` as functions might be more intuitive than having them as methods that don’t update in-place. 

On another note, I’m not sure if we should change `append_data_vectors` to `append_vectors`. We also have another method `SampleSet.data_vectors` which returns the data in the record.

arcondello: I am totally fine with making them functions (deprecation warning for the existing method of course).

randomir: I guess it boils down to whether we want SampleSet to be a thin wrapper around numpy's recarray, or a (feature) rich container. I could easily argue for both, e.g. (1) savvy users with performance in mind would probably prefer thin wrapper over numpy; (2) users caring about API simplicity would probably appreciate rich feature set more than performance. The art, I guess, is in re-conciliating those two worlds.

Anyway, @arcondello, you're closer to dimod/SampleSet, so take my suggestions lightly. :)

dimod/sampleset.py
```diff
@@ -1391,6 +1391,81 @@ def slice(self, *slice_args, **kwargs):
...
+            :obj:`.SampleSet`: SampleSet. If `inplace` is True, returns itself.
+
+        Examples:
+            The following example appends a field of lists to :attr:`SampleSet.record`.
+
+            >>> import numpy as np
+            ...
+            >>> sampleset = dimod.SampleSet(np.rec.array([([-1,  1], -1.4, 1), ([-1,  1], -1.4, 1)],
+            ...                             dtype=[('sample', 'i1', (2,)), ('energy', '<f8'), 
+            ...                             ('num_occurrences', '<i8')]), [0, 1], {}, 'SPIN')
```
randomir, line 1419: Why not something lighter, like:
```suggestion
            >>> sampleset = dimod.SampleSet.from_samples([[-1,  1], [-1,  1]], energy=[1, 1], vartype='SPIN')
```
Then you also don't need numpy import.

But ignore if you were trying to visualize the record structure this way.

dimod/sampleset.py
hhtong: It seems to me that everyone here is fine with having `append_data_vectors` as a thin wrapper around recarray, so I’m going to go ahead and make it (and `append_variables`) a function that will return a new SampleSet. 

@radomir I would also be ok with having an additional update method that is always in place.

arcondello (approved):

PR #748, opened 2020-11-09
Added header files for new roof duality implementation

amahmudwave:
The roof duality is being re-implemented, these header files are going to be
used in the final version. This PR is to merge these header files early to make
the final PR smaller and also so that they can be reviewed early.

### DIFF
--- a/dimod/roof_duality/src/helper_data_structures.hpp
+++ b/dimod/roof_duality/src/helper_data_structures.hpp
@@ -1,0 +1,126 @@
+/**
+# MIT License
+#
+# Copyright (c) 2019 Jan Groschaft
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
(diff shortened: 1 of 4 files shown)


amahmudwave:
@arcondello these can be reviewed early, since a single PR for roof duality would be too big to review at once. 
Thanks !

arcondello:
This looks good to me, but I am increasingly thinking that this should live in a new package, `dwave-optimization` or similar, rather than dimod. @randomir , what do you think about us making that package for @amahmudwave to make PRs against? We can figure out structure etc as we go.

randomir:
@arcondello, absolutely, that's exactly what I was proposing for #567.

We should just figure out the "low-levelness" of the package. It would be nice to be on the `dimod` level, with practically no Ocean dependencies. Then again, it'll probably be hard to stay like that as we keep adding hybrid support utils.

Re: name, I like it, but we should think about it in the light of "optimization vs. sampling" dichotomy.

arcondello:
For dependencies, totally agree that low level would be ideal. I think it is likely that it will end up depending on dimod, specifically the C++ stuff, which brings numpy. #567 comes with [mip](https://pypi.org/project/mip/) and SciPy... I think we would want to prune that if possible. But yes, we'll have to work it as we go.

My inclination would be to make the repo, starting with `dwave-optimization`, and rename if needed?

JoelPasvolsky:
Can you say more about what functionality would be in such a package. Do you mean a package for generic code optimization?

arcondello:
For things like roof duality, the lower bounds in #567. Possibly would even include some of the stuff in dwave-binarycsp.

arcondello:
Closing because this has moved to a new package

PR #755, opened 2020-11-13
BQM functions docstring updates 

JoelPasvolsky:


dimod/bqm/construction.py
```diff
@@ -66,7 +66,7 @@ def as_bqm(*args, cls=None, copy=False):
         *args:
             See above.
 
-        cls (type/list, optional, default=:class:`.AdjVectorBQM`):
+        cls (type/list, optional, default=\ :class:`.AdjVectorBQM`):
```
arcondello, line 69: What is this for?

JoelPasvolsky: Breaks sphinx rendition. Here's what it looks like currently on RTD:
image

arcondello: I see, it works for the first argument, but not for `default=`. That's annoying.

JoelPasvolsky: They fixed the problem of adjacent periods but most adjacent characters need backslashing

arcondello: I hate it. I know it's silly, but how about
```
        cls (type/list, optional):
            Default  :class:`.AdjVectorBQM`. Class of the returned BQM. If...
```

The problem is my linter (and I imagine others) flags the backslash as an error.

JoelPasvolsky: If we go with stating the default in the description, then probably at the end. The disadvantage of moving it is that because we often state the default in the parentheses, people might scan those and not see it. Another option is `cls (type/list, optional, default is :class:`.AdjVectorBQM`):` although I'd want to check that the parenthesis doesn't mess that up :-)

arcondello: I think anyone looking to understand the default will know that if it's not stated in the first line, it'll be in the description. We do it in several other places in Ocean as well.

I don't hate the `default is`, but prefer in the description.

JoelPasvolsky: I can go with 
```
cls (type/list, optional)  
     Class of the returned BQM. If given as a list, the returned BQM is of one of the types in the list.  Default is :class:`AdjVectorBQM`.
```

arcondello: Works for me, thanks.

dimod/generators/random.py
```diff
@@ -27,7 +27,7 @@
 @graph_argument('graph')
 def uniform(graph, vartype, low=0.0, high=1.0, cls=BinaryQuadraticModel,
             seed=None):
-    """Generate a binary quadratic model with random biases and offset.
+    """Generate a binary quadratic model with random integer biases and offset.
```
arcondello, line 30: Not integer, we use `np.random.uniform`

arcondello (approved):

Issue #761, opened 2020-11-25
Disallow self-interactions among cases in DQM variables


arcondello:
Right now it is possible to specify self-interactions within DQM variables, e.g.
```
dqm.set_quadratic(i, i, {(c, c): 1})
```
These should instead raise an error.

Note: de-serializaing DQMs constructed in this way raises and exception (as of https://github.com/dwavesystems/dimod/pull/759, before it would segfault).

PR #766, opened 2020-12-03
Refactor DQM.from_numpy_vectors

### LINKED ISSUE #762: Use promote_types rather than result_types in DQM.from_numpy_vectors
https://github.com/dwavesystems/dimod/blob/aee92735437ff9290f306c2b90d45bd18ecdc1e6/dimod/discrete/cydiscrete_quadratic_model.pyx#L257

`result_types` is for multiplication, this use case needs `promote_types`.

arcondello:
And create some functions to abstract away some of the type coercion.

Closes #762 
See also #761, this PR disallows self loops when constructing from NumPy arrays

### DIFF
--- a/dimod/include/dimod/adjvectorbqm.h
+++ b/dimod/include/dimod/adjvectorbqm.h
@@ -209,9 +209,19 @@ class AdjVectorBQM {
             adj.resize(max_label + 1);
         }
 
-        // we can get 5-10% speedup on dense problems by counting the degrees
-        // and using that to reserve the neighborhood vectors. However, since
-        // it uses more memory and slows down sparse problems, we don't.
+        // Count the degrees and use that to reserve the neighborhood vectors
+        std::vector<size_type> degrees(adj.size());
+        ItRow rit(row_iterator);
+        ItRow cit(col_iterator);
+        for (size_type i = 0; i < length; ++i, ++rit, ++cit) {
+            if (*rit != *cit) {
+                degrees[*rit] += 1;
+                degrees[*cit] += 1;
+            }
+        }
+        for (size_type i = 0; i < degrees.size(); ++i) {
+            adj[i].first.reserve(degrees[i]);
+        }
 
         // add the values to the adjacency, not worrying about order or
         // duplicates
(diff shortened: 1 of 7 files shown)


dimod/include/dimod/adjvectorbqm.h
```diff
@@ -209,9 +209,19 @@ class AdjVectorBQM {
             adj.resize(max_label + 1);
         }
 
-        // we can get 5-10% speedup on dense problems by counting the degrees
-        // and using that to reserve the neighborhood vectors. However, since
-        // it uses more memory and slows down sparse problems, we don't.
+        // Count the degrees and use that to reserve the neighborhood vectors
```
arcondello, line 212: Turns out the memory saved is valuable as well (memory savings come from not over-allocating the vectors).

dimod/utilities.py
```diff
@@ -489,3 +493,121 @@ def update(self, *args, **kwargs):
...
+            raise RuntimeError("unexpected dtype")
+        dtype = np.promote_types(dtype, kind+size)
+
+    arrays = tuple(np.require(arr, dtype=dtype, requirements=requirements)
+                   for arr in arrays)
+
+    if len(arrays) > 1:
+        return arrays
+    else:
+        return arrays[0]
```
randomir, line 553: Convenient, but confusing. Input is always a sequence, but the output might be a single array. And it's not internal, since callers just propagate this.

randomir (approved):
LGTM

dimod/utilities.py
arcondello: Sure, but leads to the much nicer (IMO) syntax
```
arr0, arr1, arr2 = asintegerarrays(arr0, arr1, arr2)
arr0, arr1 = asintegerarrays(arr0, arr1)
arr0 = asintegerarrays(arr0)
```
alternative would be something like
```
arr0, arr1, arr2 = asintegerarrays([arr0, arr1, arr2])
arr0, arr1 = asintegerarrays([arr0, arr1])
arr0, = asintegerarrays([arr0])
```

I didn't go with that because numpy's `array-like` gets pretty fiddly when given lists of arrays.

PR #768, opened 2020-12-05
fixed for dot-parsed networkx graphs

filipceglik:
Hi, while using the dimod library with networkx to solve the minimum vertex cover problem, I noticed that the dimod library is having issues with networkx graph objects generated from dot files using read_dot function. 
The dimod function _init_components was using a (u,v) tuple in a for loop. It worked well for all other graphs except those generated from dot files. The quadratic parameter was a collection of dicts which in my case did not always have 2-element keys. When the bias was different than -1.0, the dict contained a 3-element key with a zero at the end (always). 
To solve this issue i changed how the function iterates over a quadratic collection to avoid the "too many values to unpack" error.

### DIFF
--- a/dimod/bqm/adjdictbqm.py
+++ b/dimod/bqm/adjdictbqm.py
@@ -195,16 +195,18 @@ class AdjDictBQM(ShapeableBQM):
         adj = self._adj
 
         if isinstance(quadratic, abc.Mapping):
-            for (u, v), bias in quadratic.items():
-                self.add_variable(u)
-                self.add_variable(v)
+            for quadraticitem in quadratic.items():
+                key = quadraticitem[0]
+                bias = quadraticitem[1]
+                self.add_variable(key[0])
+                self.add_variable(key[1])
 
-                if u == v and vartype is Vartype.SPIN:
+                if key[0] == key[1] and vartype is Vartype.SPIN:
                     offset = offset + bias  # not += on off-chance it's mutable
-                elif u in adj[v]:
-                    adj[u][v] = adj[v][u] = adj[u][v] + bias
+                elif key[0] in adj[key[1]]:
+                    adj[key[0]][key[1]] = adj[key[1]][key[0]] = adj[key[0]][key[1]] + bias
                 else:
-                    adj[u][v] = adj[v][u] = bias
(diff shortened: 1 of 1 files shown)


arcondello:
Hi @filipceglik , thanks for the PR. I think we definitely want to support networkx graphs created this way, but I am worried that the current solution will have unintended side-effects. Are you referring to graphs constructed with [read_dot](https://networkx.org/documentation/stable/reference/generated/networkx.drawing.nx_agraph.read_dot.html)?

Could you post a short example of the code that's not working for you?

filipceglik:
Hi @arcondello, absolutely. I have a function in my code which returns the networkx graph using the read_dot function. Perhaps this is a bug in the read_dot function but I wasn't able to track where this might be going wrong. I am pasting the code below. At one point the quadratic variable has keys with 3 items inside of them for the bias different that -1.0. 

image

Please bear in mind that I am not a professional developer, just someone who noticed this behavior for this particular case so please keep me informed on any issues this solution might create and I am sure we can work some correct solution if it indeed turns out to be a bug in the dimod.

arcondello:
Hi @filipceglik , thanks! I actually meant an example of a dotfile, I am not familiar with that format. Sorry for the confusion!

Agree that we should get this working.

filipceglik:
Sure thing @arcondello, attaching the file. 
I changed the extension from dot to txt since github does not support uploading dot files. 
[graph.txt](https://github.com/dwavesystems/dimod/files/5688404/graph.txt)

arcondello:
Hi @filipceglik , does your program look something like
```
G = nx.drawing.nx_agraph.read_dot(dotfile)
bqm = dimod.BQM.from_networkx_graph(G)
```
or are you constructing the BQM some other way?

filipceglik:
Hi @arcondello, 

I am really sorry, I haven't seen your reply and I just looked at the PR. 

This is the most logic of my program, it's parsing the dotfile and uses the dwave quantum computing API to determin the minimum vertex cover of a graph. 
I am creating the sampler using 
`EmbeddingComposite(DWaveSampler(endpoint=ENDPOINT, token=TOKEN, solver=SOLVER))`

image

Issue #758, opened 2020-11-20
num_oc. is misaligned when using dimod.serialization.format.Formatter 


charleswhchan:
**Description**
When using `dimod.serialization.format.Formatter`, sometimes the num_oc. column may be misaligned when the number of significant figures various with energy  

**Steps To Reproduce**
```
...
sample_pprint = dimod.serialization.format.Formatter(width=50, depth=50).format(sample_set)
print(sample_pprint)
```

Output:
```
   0  1  2  3  4  5  6  7 ... 127  energy num_oc.
0 -1 +1 +1 -1 -1 +1 +1 +1 ...  +1 -131.1        1
1 -1 +1 +1 -1 -1 +1 +1 +1 ...  +1 -131.02        1
2 -1 +1 +1 +1 +1 -1 +1 +1 ...  +1 -131.98        1
3 -1 +1 +1 +1 +1 -1 +1 +1 ...  +1 -131.9        1
4 -1 +1 +1 -1 -1 +1 +1 +1 ...  +1 -131.9        1
5 -1 +1 +1 -1 -1 +1 +1 +1 ...  +1 -131.82        1
6 -1 +1 +1 -1 -1 +1 +1 +1 ...  +1 -131.78        1
7 -1 +1 +1 -1 -1 +1 +1 +1 ...  -1 -131.78        1
8 -1 +1 +1 -1 -1 +1 +1 +1 ...  +1 -131.74        1
9 -1 +1 +1 +1 +1 -1 +1 +1 ...  +1 -131.7       1
['SPIN', 10 rows, 10 samples, 128 variables]
```

**Expected Behavior**
num_oc to be aligned vertical in a single column

**Environment**
 - OS: CentOS 6
 - Python version: 3.6

**Additional Context**

arcondello:
Do we have the original sample set that's failing? Any serialization format is fine

charleswhchan:
I couldn't find the exact issue from yesterday, but I did find another problem with similar issue: 14fe9e7a-a604-4661-a5bb-1b1bb2eb3a91

answer:
```
{
    "format": "qp",
    "num_variables": 128,
[...]

PR #770, opened 2020-12-10
Fix misaligned energy values in printed sampleset

arcondello:
Closes https://github.com/dwavesystems/dimod/issues/758

I really want to just chuck and rewrite that entire file, but just fixing the bug for now.

### DIFF
--- a/tests/test_serialization_format.py
+++ b/tests/test_serialization_format.py
@@ -206,3 +206,17 @@ class TestSampleSet(unittest.TestCase):
                             "['DISCRETE', 4 rows, 4 samples, 3 variables]"])
 
         self.assertEqual(target, s)
+
+    def test_misalignment(self):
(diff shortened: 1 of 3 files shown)


tests/test_serialization_format.py
```diff
@@ -206,3 +206,19 @@ def test_depth(self):
...
 
         self.assertEqual(target, s)
+
+    def test_misalignment(self):
+        ss = dimod.SampleSet.from_samples([[0, 1], [0, 1], [1, 0]], 'BINARY',
+                                          energy=[-1, 1.55, 2])
+
+        s = Formatter().format(ss)
+
+        print(s)
```
hhtong, line 216: print statement can be removed

hhtong (approved):
LGTM!

tests/test_serialization_format.py
arcondello: Good catch, thank you

PR #774, opened 2021-01-08
add add_constraint_as_quadratic method to DQM and cyDQM

hsadeghidw:
Add a method to `dimod.discrete_quadratic_model` to convert linear constraints to a quadratic objective.

### DIFF
--- a/dimod/discrete/cydiscrete_quadratic_model.pyx
+++ b/dimod/discrete/cydiscrete_quadratic_model.pyx
@@ -274,21 +274,37 @@ cdef class cyDiscreteQuadraticModel:
         cdef VarIndex v, u
         cdef CaseIndex cv, cu
         cdef Bias bv, bu, bias
+        cdef vector[CaseIndex] global_index
         for i in range(num_terms):
             v = variables[i]
-            cv = cases[i]
+            case = cases[i]
+            if case < 0 or case >= self.num_cases(v):
+                raise ValueError("case {} is invalid, variable only supports {} "
+                                 "cases".format(case, self.num_cases(v)))
+            global_index.push_back(self.case_starts_[v] + case)
+
+        for i in range(num_terms):
+            v = variables[i]
+            cv = global_index[i]
             bv = biases[i]
-            bias = self.get_linear_case(v, cv)
+            bias = self.bqm_.get_linear(cv)
             bias += lagrange_multiplier * (2 * bv * constant + bv * bv)
-            self.set_linear_case(v, cv, bias)
+            self.bqm_.set_linear(cv, bias)
             for j in range(i + 1, num_terms):
                 u = variables[j]
-                cu = cases[j]
+                cu = global_index[j]
                 bu = biases[j]
                 if v != u:
-                    bias = self.get_quadratic_case(v, cv, u, cu)
+                    bias = self.bqm_.get_quadratic(cv, cu).first
                     bias += 2 * lagrange_multiplier * bv * bu
-                    self.set_quadratic_case(v, cv, u, cu, bias)
+                    self.bqm_.set_quadratic(cv, cu, bias)
+
+                    low = lower_bound(self.adj_[u].begin(), self.adj_[u].end(), v)
+                    if low == self.adj_[u].end() or deref(low) != v:
+                        self.adj_[u].insert(low, v)
+                        self.adj_[v].insert(
+                            lower_bound(self.adj_[v].begin(), self.adj_[v].end(), u),
+                            u)
 
     @cython.boundscheck(False)
     @cython.wraparound(False)
--- a/dimod/discrete/cydiscrete_quadratic_model.pyx
+++ b/dimod/discrete/cydiscrete_quadratic_model.pyx
@@ -279,7 +279,7 @@ cdef class cyDiscreteQuadraticModel:
             cv = cases[i]
             bv = biases[i]
             bias = self.get_linear_case(v, cv)
-            bias += 2 * lagrange_multiplier * (bv * constant + bv * bv)
+            bias += lagrange_multiplier * (2 * bv * constant + bv * bv)
             self.set_linear_case(v, cv, bias)
             for j in range(i + 1, num_terms):
                 u = variables[j]
--- a/dimod/discrete/cydiscrete_quadratic_model.pyx
+++ b/dimod/discrete/cydiscrete_quadratic_model.pyx
@@ -265,20 +265,31 @@ cdef class cyDiscreteQuadraticModel:
     @cython.wraparound(False)
     cpdef void add_constraint_as_quadratic(self, vector[VarIndex] variables, vector[CaseIndex] cases,
                                            vector[Bias] biases, Bias lagrange_multiplier, Bias constant):
-        num_terms = len(variables)
+        if len(variables) != len(cases) or len(variables) != len(biases):
+            raise ValueError("There should be equal number of variables, "
+                             "cases, and biases. Each term should be a "
+                             "3-tuple")
+
+        cdef unsigned long int num_terms = len(variables)
+        cdef unsigned long int i, j
+        cdef VarIndex v, u
+        cdef CaseIndex cv, cu
+        cdef Bias bv, bu, bias
         for i in range(num_terms):
             v = variables[i]
             cv = cases[i]
             bv = biases[i]
             bias = self.get_linear_case(v, cv)
-            self.set_linear_case(v, cv, bias + 2 * lagrange_multiplier * bv * constant + lagrange_multiplier * bv * bv)
+            bias += 2 * lagrange_multiplier * (bv * constant + bv * bv)
+            self.set_linear_case(v, cv, bias)
             for j in range(i + 1, num_terms):
                 u = variables[j]
                 cu = cases[j]
                 bu = biases[j]
                 if v != u:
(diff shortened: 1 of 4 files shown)


dimod/discrete/discrete_quadratic_model.py
```diff
@@ -416,6 +419,24 @@ def get_quadratic_case(self, u, u_case, v, v_case):
         return self._cydqm.get_quadratic_case(
             self.variables.index(u), u_case, self.variables.index(v), v_case)
 
+    def add_constraint_as_quadratic(self, terms: Linear, lagrange_multiplier: float,
+                                    constant: float):
+        """
```
arcondello, line 424: Should be on the first line, also no more than 80 char wide

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -416,6 +419,24 @@ def get_quadratic_case(self, u, u_case, v, v_case):
...
+        Add a linear constraint of the form \\sum_{ik} a_{ik} x_{ik} + C = 0 to dqm object as a quadratic objective.
+        Args:
```
arcondello, line 426: Space between description and `Args`

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -416,6 +419,24 @@ def get_quadratic_case(self, u, u_case, v, v_case):
...
+            dqm: DiscreteQuadraticModel
```
arcondello, line 427: This shouldn't be here

dimod/discrete/cydiscrete_quadratic_model.pyx
```diff
@@ -261,6 +261,25 @@ cdef class cyDiscreteQuadraticModel:
 
         return biases
 
+    @cython.boundscheck(False)
```
arcondello, line 264: If we're not doing these checks here, we need to do some sanity checks in the code body. Otherwise this function can segfault

dimod/discrete/cydiscrete_quadratic_model.pyx
```diff
@@ -261,6 +261,25 @@ cdef class cyDiscreteQuadraticModel:
...
+    @cython.wraparound(False)
+    cpdef void add_constraint_as_quadratic(self, vector[VarIndex] variables, vector[CaseIndex] cases,
+                                           vector[Bias] biases, Bias lagrange_multiplier, Bias constant):
+        num_terms = len(variables)
+        for i in range(num_terms):
+            v = variables[i]
+            cv = cases[i]
+            bv = biases[i]
+            bias = self.get_linear_case(v, cv)
+            self.set_linear_case(v, cv, bias + 2 * lagrange_multiplier * bv * constant + lagrange_multiplier * bv * bv)
```
arcondello, line 274: max width of 80 characters

arcondello (changes requested):
Need to be careful with our inputs to not get segfaults. Also needs tests

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -416,6 +419,24 @@ def get_quadratic_case(self, u, u_case, v, v_case):
...
+            terms: A list or a generator of tuples of the type (variable, case, bias).
+                Each tuple is evaluated to the term bias * variable_case. All terms in the list/generator are summed.
+            lagrange_multiplier: The coefficient or the penalty strength
+            const: The constant value of the constraint.
+
+        Returns: None. Updates the dqm object
+
+        """
+        variables, cases, biases = zip(*terms)
```
arcondello, line 436: IMO we should not accept a generator if we're just going to dump it all into memory in the first line.

arcondello: I am wrong, since even in the case that you're giving it as a list, it's making a copy into the three different lists. I suppose a generator saves on half the memory in this case.

arcondello:
```
for v, c, b in terms:
    variables.append(self.variables.index(v))
    cases.append(c)
    biases.append(b)
```
will save a decent chunk of memory without a performance hit I don't think.

dimod/discrete/cydiscrete_quadratic_model.pyx
hsadeghidw: will do a check on the length of all three vectors

dimod/discrete/cydiscrete_quadratic_model.pyx
hsadeghidw: will do

dimod/discrete/discrete_quadratic_model.py
hsadeghidw: ok

dimod/discrete/discrete_quadratic_model.py
hsadeghidw: ok

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -421,20 +421,25 @@ def get_quadratic_case(self, u, u_case, v, v_case):
 
     def add_constraint_as_quadratic(self, terms: Linear, lagrange_multiplier: float,
                                     constant: float):
-        """
-        Add a linear constraint of the form \\sum_{ik} a_{ik} x_{ik} + C = 0 to dqm object as a quadratic objective.
+        """Add a linear constraint of the form below
```
arcondello, line 424: How about
```
    """Add a linear constraint as a quadratic objective

    Adds a linear constraint of the form :math:`\sum_{i,k} a_{i,k} x_{i,k} + C = 0`
    to the discrete quadratic model as a quadratic objective.

    Args:
        ...
```

The reason is that was sphinx can pick it up for the summary. You'll also want to add this method to the docs for the DQM, see https://github.com/dwavesystems/dimod/blob/f4d2744e9514e6d0ab41fd9637bdfe021919029b/docs/reference/dqm.rst

dimod/discrete/cydiscrete_quadratic_model.pyx
```diff
@@ -261,6 +261,36 @@ cdef class cyDiscreteQuadraticModel:
 
         return biases
 
+    @cython.boundscheck(False)
+    @cython.wraparound(False)
+    cpdef void add_constraint_as_quadratic(self, vector[VarIndex] variables, vector[CaseIndex] cases,
+                                           vector[Bias] biases, Bias lagrange_multiplier, Bias constant):
+        if len(variables) != len(cases) or len(variables) != len(biases):
+            raise ValueError("There should be equal number of variables, "
```
arcondello, line 269: I know it's pedantic, but I think this exception should just be "There should be an equal number of variables, cases and biases". The 3-tuple part is not applicable for those who are accessing this method directly - it only applies to the DQM method, _not_ the cyDQM one. Also, since the user is giving the input as three-tuples in the other method, this error should never be seen when calling from the DQM method.

tests/test_discrete_quadratic_model.py
```diff
@@ -409,6 +409,29 @@ def test_set_quadratic_offdtype(self, name, dtype):
...
 
 
+class TestConstraint(unittest.TestCase):
+    def test_simple_constraint(self):
+        dqm = dimod.DQM()
+        num_variables = 2
+        num_cases = 3
+        x = {}
+        for i in range(num_variables):
+            x[i] = dqm.add_variable(num_cases, label=f'x_{i}')
```
hbarovertwo, line 419: this line will cause issues in python 3.5 because f strings were not supported at the time. I think instead you could do `label='x_{i}'.format(i=i)`

hbarovertwo (changes requested):

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -27,6 +27,9 @@
 from dimod.sampleset import as_samples
 from dimod.serialization.fileview import VariablesSection, _BytesIO
 from dimod.variables import Variables
+from typing import List, Tuple, Union, Generator
+
+Linear = Union[List[Tuple], Generator[Tuple, None, None]]
```
arcondello, line 32: One last thing, I think `Linear` is too broad of a name, there are a lot of other things we'd like to call Linear. Maybe `LinearConstraint` or `LinearTriplets` or something

arcondello (approved):

arcondello:
LGTM, subject to fixing the logic so the test works.

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -416,6 +419,29 @@ def get_quadratic_case(self, u, u_case, v, v_case):
...
+        to the discrete quadratic model as a quadratic objective.
+
+        Args:
+            terms: A list of tuples of the type (variable, case, bias).
+                Each tuple is evaluated to the term (bias * variable_case).
+                All terms in the list are summed.
+            lagrange_multiplier: The coefficient or the penalty strength
+            constant: The constant value of the constraint.
+
+        Returns: None
```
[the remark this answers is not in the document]
arcondello: If `None`, just drop the `Returns` section altogether.

arcondello (approved):
I believe this is ready to merge now?

hsadeghidw:
Yes, I believe it's ready to go!

arcondello:
@hbarovertwo do you still want to do more review?

hbarovertwo (approved):

PR #776, opened 2021-01-20
[WIP] DQM refactoring for Q/A and design discussion.

amahmudwave:


dimod/discrete/cppdqm.pxd
```diff
@@ -0,0 +1,97 @@
...
+# =============================================================================
+
+from libcpp cimport bool
+from libcpp.pair cimport pair
+from libcpp.vector cimport vector
+
+cimport numpy as np
+
+from dimod.bqm.common cimport VarIndex, Bias
+from dimod.bqm.cppbqm cimport AdjVectorBQM as cppAdjVectorBQM
```
amahmudwave, line 27: Included the bqm as is done in other places.

dimod/discrete/cppdqm.pxd
```diff
@@ -0,0 +1,97 @@
...
+
+cdef extern from "dimod/adjvectordqm.h" namespace "dimod" nogil:
+
+    cdef cppclass AdjVectorDQM[V, B]:
+        ctypedef V variable_type
+        ctypedef B bias_type
+        ctypedef size_t size_type
+
+	cdef cppAdjVectorBQM[variable_type, bias_type] bqm_ 
+        vector[variable_type] case_starts_
```
amahmudwave, line 37: What do you use to format the pxd files ? 
I have been using the clang-format file you gave me for c++ files though.

dimod/discrete/cppdqm.pxd
```diff
@@ -0,0 +1,97 @@
...
+        # methods
+
+        bool is_self_loop_present() except +
+        size_type num_variables() except +
+	size_type num_variable_interactions() except +
+	size_type num_cases() except +
+	size_type num_case_interactions() except +
+        bias_type get_linear_case(variable_type, variable_type) except +
+        void set_linear_case(variable_type, variable_type, bias_type) except +
+        void get_linear[io_bias_type](variable_type, io_bias_type*) except +
```
amahmudwave, line 86: io_bias_type / io_variable_type. are used for passing templated pointers, io standing for input output here.

dimod/discrete/cydiscrete_quadratic_model.pxd
```diff
@@ -20,7 +20,7 @@ from libcpp.vector cimport vector
 
 cimport numpy as np
 
-from dimod.bqm.cppbqm cimport AdjVectorBQM as cppAdjVectorBQM
+from dimod.discrete.cppdqm cimport AdjVectorDQM as cppAdjVectorDQM
```
amahmudwave, line 23: As we discussed the bqm should be accessible as we included it in the pxd file.

dimod/discrete/cydiscrete_quadratic_model.pyx
```diff
@@ -22,14 +22,10 @@ from libcpp.unordered_set cimport unordered_set
 
 import numpy as np
 
-from dimod.utilities import asintegerarrays, asnumericarrays
-
 
 cdef class cyDiscreteQuadraticModel:
 
     def __init__(self):
-        self.case_starts_.push_back(0)
```
amahmudwave, line 31: The default constructor of the dqm pushes back 0, I think we do not need it here anymore.

dimod/discrete/cydiscrete_quadratic_model.pyx
```diff
@@ -51,24 +47,12 @@ cdef class cyDiscreteQuadraticModel:
...
-        self.adj_.resize(v+1)
-
-        cdef Py_ssize_t i
-        for i in range(num_cases):
-            self.bqm_.add_variable()
-
-        self.case_starts_.push_back(self.bqm_.num_variables())
-
-        return v
+        return self.dqm_.add_variable(num_cases)
```
amahmudwave, line 50: I am returning the value returned by the C function directly, is there any type conversion from c value to python value needed here ?

dimod/discrete/cydiscrete_quadratic_model.pyx
```diff
@@ -83,38 +67,19 @@ cdef class cyDiscreteQuadraticModel:
         cdef Py_ssize_t num_samples = samples.shape[0]
         cdef VarIndex num_variables = samples.shape[1]
 
-        cdef Bias[:] energies = np.zeros(num_samples, dtype=self.dtype)
+        cdef Bias[:] energies = np.empty(num_samples, dtype=self.dtype)
```
amahmudwave, line 70: I do not need the array to be filled with zeroes, as I overwrite the values in the way I calculated energy.

dimod/discrete/cydiscrete_quadratic_model.pyx
```diff
@@ -83,38 +67,19 @@ cdef class cyDiscreteQuadraticModel:
...
-                    case_v = samples[si, v]
-
-                    cv = self.case_starts_[v] + case_v
-
-                    out = self.bqm_.get_quadratic(cu, cv)
-
-                    if out.second:
-                        energies[si] += out.first
-
+        self.dqm_.get_energies(&samples[0,0], num_samples, num_variables, &energies[0])
```
amahmudwave, line 82: This is a templated function, will this work ? I saw some examples which indicate it should work. 
Secondly is it okay to pass the address of the starting point of samples the way I did. I could also pass in samples.data but I was wondering in that case templating may not work since samples.data may be just a void pointer.

dimod/discrete/cydiscrete_quadratic_model.pyx
```diff
@@ -160,68 +125,14 @@ cdef class cyDiscreteQuadraticModel:
...
-            span = dqm.bqm_.neighborhood(ci)
-
-            v = 0
-            while span.first != span.second:
-                cj = deref(span.first).first
-
-                # see above note
-                while cj >= dqm.case_starts_[v+1]:
-                    v += 1
+        cdef cyDiscreteQuadraticModel cyDQM = cls()
```
amahmudwave, line 128: Quite confused regarding the next 5 lines of code. I will look up more into this or talk to you in person.

dimod/discrete/cppdqm.pxd
```diff
@@ -0,0 +1,97 @@
...
+	cdef cppAdjVectorBQM[variable_type, bias_type] bqm_ 
+        vector[variable_type] case_starts_
+        vector[vector[variable_type]] adj_
+
+        # constructors
+        # cython cannot handle templated constructors, so we call out the types
+        # explicitly
+
+        AdjVectorDQM() except +
+
```
amahmudwave, line 45: Do I need copy constructors or anything like that ? 
The dqm object does not contain any pointers so shouldnt the default copy constructor work fine ? Since no deep copying is not needed.

dimod/discrete/cydiscrete_quadratic_model.pyx
```diff
@@ -51,24 +47,12 @@ cdef class cyDiscreteQuadraticModel:
...
-        for i in range(num_cases):
-            self.bqm_.add_variable()
-
-        self.case_starts_.push_back(self.bqm_.num_variables())
-
-        return v
+        return self.dqm_.add_variable(num_cases) 
 
     def copy(self):
         cdef cyDiscreteQuadraticModel dqm = type(self)()
```
amahmudwave, line 53: I am assuming no copy constructor is needed since the dqm object does not contain any pointer, so deep copy is not needed I think.

dimod/discrete/cydiscrete_quadratic_model.pyx
```diff
@@ -291,32 +197,21 @@ cdef class cyDiscreteQuadraticModel:
 
         if array:
             # build a numpy array
-            quadratic = np.zeros((self.num_cases(u), self.num_cases(v)),
+            quadratic = np.empty((self.num_cases(u), self.num_cases(v)),
```
amahmudwave, line 200: Filling up the array with zeroes has extra overhead, may be nothing compared to python's own overhead, but still its not needed.

dimod/discrete/cydiscrete_quadratic_model.pyx
```diff
@@ -291,32 +197,21 @@ cdef class cyDiscreteQuadraticModel:
...
-
-                    inc(span.first)
+	    self.dqm_.get_quadratic(u, v, &quadratic[0,0])
 
         else:
             # store in a dict
             quadratic = {}
 
-            for ci in range(self.case_starts_[u], self.case_starts_[u+1]):
+            for ci in range(self.dqm_.case_starts_[u], self.dqm_.case_starts_[u+1]):
```
amahmudwave, line 208: Leaving this part in cython, seems like it is more python specific and will not be needed in the c++ version of the dqm. I assume self.dqm_.case_starts_ .... these are all accessible as usual.

dimod/include/dimod/adjvectordqm.h
```diff
@@ -0,0 +1,312 @@
...
+
+        // Set the case starts.
+        case_starts_.resize(num_variables + 1);
+        for (auto v = 0; v < num_variables; v++) {
+            case_starts_[v] = case_starts[v];
+        }
+        case_starts[num_variables] = num_cases;
+
+        // Fill the adjacency list for variables.
+        std::vector<std::unordered_set<variable_type>> adjset;
```
amahmudwave, line 70: We probably could parallelize it later.

dimod/include/dimod/adjvectordqm.h
```diff
@@ -0,0 +1,312 @@
...
+                      io_variable_type *icol, io_bias_type *quadratic_biases) {
+        auto num_variables = this->num_variables();
+        auto num_total_cases = bqm_.num_variables();
+
+        for (auto v = 0; v < num_variables; v++) {
+            case_starts[v] = case_starts_[v];
+        }
+
+        size_type qi = 0;
+        for (auto ci = 0; ci < num_total_cases; ci++) {
```
amahmudwave, line 108: This could be parallelized.

This could be moved to adjvectorbqm too.

dimod/include/dimod/adjvectordqm.h
```diff
@@ -0,0 +1,312 @@
...
+                span.first++;
+            }
+        }
+
+        adj_.resize(num_variables);
+#pragma omp parallel for
+        for (auto v = 0; v < num_variables; v++) {
+            adj_[v].insert(adj_[v].begin(), adjset[v].begin(), adjset[v].end());
+            std::sort(adj_[v].begin(), adj_[v].end());
+        }
```
amahmudwave, line 94: Will check the self loop using a separate function since the constructor cannot return any value. The other function will. And will be called from cython.

amahmudwave:
@arcondello  your input will be helpful here :) thanks.

dimod/discrete/cppdqm.pxd
arcondello: No, I use python's formatting. pep8 doesn't directly apply but it's pretty close.

dimod/discrete/cydiscrete_quadratic_model.pyx
arcondello: If it's a numeric type then cython will handle it

dimod/discrete/cydiscrete_quadratic_model.pyx
arcondello: Honestly I am not sure if it will work. You'll have to try and see!

I am not 100% sure how address of a python object like that works. Cython might be able to handle it, might not.

dimod/discrete/cppdqm.pxd
arcondello: Probably better to be explicit

dimod/discrete/cydiscrete_quadratic_model.pyx
```diff
@@ -435,32 +326,14 @@ cdef class cyDiscreteQuadraticModel:
                     raise ValueError("case {} is invalid, variable only supports {} "
                                      "cases".format(case_v, self.num_cases(v)))
 
-                cu = self.case_starts_[u] + case_u
-                cv = self.case_starts_[v] + case_v
+                self.dqm_.set_quadratic_case(u, case_u, v, case_v, bias)
 
-                self.bqm_.set_quadratic(cu, cv, bias)
+            self.dqm_.connect_variables(u,v)
```
amahmudwave, line 331: @arcondello  I put this line of code only for a test case to pass, I think I should remove it and modify the test case, instead
'''
350         dqm.set_quadratic('a', 'b', {})
351
352         self.assertEqual(dqm.get_quadratic('a', 'b'), {})
'''

since if we serialize this dqm, and de-serialize differences will arise, as connections among variables are recreated based on actual connections among cases.  I can fix it in this PR or in a different PR (since this changes the original behaviour of the code, so far it just imitates)

PR #777, opened 2021-01-21
Fix DQM.copy

arcondello:


dimod/discrete/cydiscrete_quadratic_model.pyx
```diff
@@ -73,6 +73,8 @@ cdef class cyDiscreteQuadraticModel:
         dqm.dtype = self.dtype
         dqm.case_dtype = self.dtype
 
+        return dqm
```
randomir, line 76: Best fix ever. :rofl:

arcondello: Certainly an important line of code!

rahmani15 (approved):

PR #779, opened 2021-01-22
Fix handling of integer-valued floats in Variables

### LINKED ISSUE #778: Accessing DQM variables with np.int64, etc. array integer elements throws ValueError
**Description**
When I use a numpy array with `np.int64`,  `np.int32`, or `np.uint16`  integer elements to label DQM variables, accessing the variables through DQM method `num_cases(...)` fails with integer elements from the numpy array used to apply labels.

**Steps To Reproduce**
```
import numpy as np
from dimod import DQM

dqm = DQM()

variables = np.array([0,1,2], dtype=np.int64)

for variable in variables:
    dqm.add_variable(num_cases=5, label=variable)

dqm.num_cases(variables[0])
>>>ValueError: unknown variable 0
```

However this works fine to access:
`dqm.num_cases(0)`

**Expected Behavior**
To be able to access the variable by the (numpy) integer label used to create it.

**Environment**
 - OS: Windows 10 Pro 64-bit, Ubuntu 20.04 LTS
 - Python version: 3.8.5

**Additional Context**
Observed in dimod versions `0.9.10`, `0.9.13`

arcondello:
Closes #778

### DIFF
--- a/dimod/variables.py
+++ b/dimod/variables.py
@@ -116,13 +116,18 @@ class Variables(abc.Sequence, abc.Set):
 
     def __contains__(self, v):
         try:
-            return (isinstance(v, int)
-                    and 0 <= v < self._stop
-                    and v not in self._idx_to_label
+            in_range = (isinstance(v, Number)
+                        and (isinstance(v, Integral) or v.is_integer())
+                        and 0 <= v < self._stop)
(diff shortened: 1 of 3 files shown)


hbarovertwo (approved):
LGTM!

PR #781, opened 2021-01-24
Ported DQM code to C++

amahmudwave:
Ported all the dqm code to C++. Not making any attempt to rebase with upstream/main since there is a lot of divergence now.


dimod/discrete/cydiscrete_quadratic_model.pyx
```diff
@@ -435,32 +326,14 @@ cdef class cyDiscreteQuadraticModel:
                     raise ValueError("case {} is invalid, variable only supports {} "
                                      "cases".format(case_v, self.num_cases(v)))
 
-                cu = self.case_starts_[u] + case_u
-                cv = self.case_starts_[v] + case_v
+                self.dqm_.set_quadratic_case(u, case_u, v, case_v, bias)
 
-                self.bqm_.set_quadratic(cu, cv, bias)
+            self.dqm_.connect_variables(u,v)
```
amahmudwave, line 331: @arcondello  I put this line of code only for a test case to pass, I think I should remove it and modify the test case, instead

```
350 dqm.set_quadratic('a', 'b', {})
351
352 self.assertEqual(dqm.get_quadratic('a', 'b'), {})
```

since if we serialize this dqm, and de-serialize differences will arise, as connections among variables are recreated based on actual connections among cases. I can fix it in this PR or in a different PR (since this changes the original behaviour of the code, so far it just imitates)

amahmudwave:
@arcondello  This is ready for review, I guess we need some co-ordination to get it merged. I think merging will be better than rebasing in this case.

dimod/include/dimod/adjvectordqm.h
```diff
@@ -0,0 +1,356 @@
...
+        variable_type v = adj_.size();
+        adj_.resize(v + 1);
+        for (variable_type n = 0; n < num_cases; n++) {
+            bqm_.add_variable();
+        }
+        case_starts_.push_back(bqm_.num_variables());
+        return v;
+    }
+
+    bias_type get_linear_case(variable_type v, variable_type case_v) {
```
arcondello, line 200: In the bqm I am switching to `bias_type& linear(v)` and I think `bias_type& linear_case(u, v)` would also be preferable to get/set.

Note that for setting to work with cython, you need some trickery. E.g.
```
        cdef bias_type *b = &(self.cppbqm.linear(v))
        b[0] = bias
```
In python there isn't really a notion of a reference/proxy object in the same way. So we'll want to keep get/set.

dimod/include/dimod/adjvectordqm.h
```diff
@@ -0,0 +1,356 @@
...
+        if (inserted) {
+            connect_variables(u, v);
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    template <class io_variable_type, class io_bias_type>
+    void get_energies(io_variable_type *samples, int num_samples,
```
arcondello, line 323: IMO it makes sense to just have a single method, `bias_type energy(Iter sample_start)`. The user can then parallelize it themselves if they like.

dimod/include/dimod/adjvectordqm.h
```diff
@@ -0,0 +1,356 @@
...
+                quadratic_biases[case_u * num_cases_v + case_v] =
+                    (span.first)->second;
+                span.first++;
+            }
+        }
+        return true;
+    }
+
+    template <class io_bias_type>
+    bool set_quadratic(variable_type u, variable_type v, io_bias_type *biases) {
```
arcondello, line 291: I'd rather accept an iterator than a pointer (applies other places). Let me know if you disagree.

dimod/discrete/cydiscrete_quadratic_model.pyx
arcondello: Right, I know I told you the old behavior was a bug, but I think it is in fact meaningful for `dqm.set_quadratic(u, v, {})` to set the interaction, but not biases. I know it's weird, but I suppose I can imagine a case where I populate the structure of the DQM before setting the biases where I would want to do that.

dimod/include/dimod/adjvectordqm.h
```diff
@@ -0,0 +1,356 @@
...
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#ifndef DIMOD_ADJVECTORDQM_H_
+#define DIMOD_ADJVECTORDQM_H_
+
+#include <stdio.h>
```
arcondello, line 18: Is this used?

dimod/include/dimod/adjvectordqm.h
```diff
@@ -0,0 +1,356 @@
...
+        adj_.resize(num_variables);
+#pragma omp parallel for
+        for (variable_type v = 0; v < num_variables; v++) {
+            adj_[v].insert(adj_[v].begin(), adjset[v].begin(), adjset[v].end());
+            std::sort(adj_[v].begin(), adj_[v].end());
+        }
+    }
+
+    template <class io_variable_type, class io_bias_type>
+    void extract_data(io_variable_type *case_starts,
```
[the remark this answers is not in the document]
arcondello: Alternatively, since this is just reading the DQM, I think it makes sense to leave this code in cython.

dimod/include/dimod/adjvectordqm.h
```diff
@@ -0,0 +1,356 @@
...
+    using variable_type = V;
+    using size_type = std::size_t;
+
+    AdjVectorBQM<variable_type, bias_type> bqm_;
+    std::vector<variable_type> case_starts_;  // len(adj_) + 1
+    std::vector<std::vector<variable_type>> adj_;
+
+    AdjVectorDQM() { case_starts_.push_back(0); }
+
+    explicit AdjVectorDQM(const AdjVectorDQM &dqm) {
```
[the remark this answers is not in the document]
amahmudwave: I have to dig deeper into it. I followed the AdjVectorBQM, which does use a template over here, but in the cppbqm.pxd file it has only the declaration for something like AdjVectorBQM(cons AdjVectorBQM & bqm), we are not passing a template parameter for specifying the type of the other BQM.

dimod/include/dimod/adjvectordqm.h
```diff
@@ -0,0 +1,356 @@
...
+                if ((span.first != span.second) &&
+                    ((span.first)->first < case_starts_[v + 1])) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    bool connection_present(variable_type u, variable_type v) {
```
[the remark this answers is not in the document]
amahmudwave: This needs to be used inside cython, to raise value error, when no connection is present.

dimod/include/dimod/adjvectordqm.h
```diff
@@ -0,0 +1,356 @@
...
+        bool connected = true;
+        auto it = std::lower_bound(adj_[u].begin(), adj_[u].end(), v);
+        if (it == adj_[u].end() || *it != v) {
+            connected = false;
+        }
+        return connected;
+    }
+
+    void connect_variables(variable_type u, variable_type v) {
```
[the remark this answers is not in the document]
amahmudwave: I made it private initially but the problem is for the bug I mentioned above this is needed, to create the connection in cython.

dimod/include/dimod/adjvectordqm.h
amahmudwave: If we pass in an iterator while passing a pointer we are sort of hiding the real information, we can always use the pointer as an iterator. The thing is we might end up doing pointer arithmetic which does not look too good on an iterator.

amahmudwave (commented):
Addressed the review comments, regarding taking to_coo (previously extract_data) , if we want to parallelize it, it will be a lot of code, so I think its better to keep inside c++ instead of cython.

dimod/include/dimod/adjvectordqm.h
amahmudwave: Removed

dimod/include/dimod/adjvectordqm.h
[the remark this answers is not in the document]
amahmudwave: Templated it the way adjvectorbqm copy constructor is templated.

dimod/include/dimod/adjvectordqm.h
arcondello: I guess it seems to me that the iterator is the more general object. Specifically here we mean a random access iterator, since as you say we might end up doing arithmetic. But using an iterator instead would allow us to use strided arrays for instance.

arcondello: Let me think about it a bit, this is also coming up in the BQM case.

arcondello:
Closing as stale

PR #784, opened 2021-01-27
Bqm add constraint

hsadeghidw:
Add a method for converting a linear constraint to a quadratic objective.


docs/reference/bqm_classes/adjmapbqm.rst
```diff
@@ -82,6 +82,7 @@ Methods
    ~AdjMapBQM.scale
    ~AdjMapBQM.set_linear
    ~AdjMapBQM.set_quadratic
+   ~AdjVectorBQM.add_constraint_as_quadratic
```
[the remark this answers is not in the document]
arcondello: Also, and I know how pedantic this is, these should be in alphabetical order :smile:

dimod/bqm/shapeablebqm.pyx.src
```diff
@@ -914,6 +914,52 @@ cdef class cyAdj@name@BQM:
 
         self.bqm_.set_linear(vi, b)
 
+    def add_constraint_as_quadratic(self, object terms, Bias
```
arcondello, line 917: Methods should be in alphabetical order (after special methods and properties). Private methods can be placed near their main methods.

randomir, line 917: This method sounds lot more general than it is.

Call it perhaps `add_linear_equality_constraint`? Or `add_linear_equality_constraint_as_penalty`?

dimod/bqm/shapeablebqm.pyx.src
```diff
@@ -914,6 +914,52 @@ cdef class cyAdj@name@BQM:
...
+                                    lagrange_multiplier, Bias constant):
```
randomir, line 918: Doesn't have to be implemented here, but let's plan to have (in version 1) an adaptive default value for `lagrange_multiplier`. (To avoid chain strength-like issues)

arcondello: I think for now, I prefer the explicit value. I know it passes complexity to the user, but for now when we're lacking best practices to me it makes sense.

dimod/bqm/shapeablebqm.pyx.src
arcondello: We would need to also update the DQM version. I am not necessarily opposed (it hasn't been released yet). Of the two, I prefer `add_linear_equality_constraint`. But would defer to @hsadeghidw ultimately.

dimod/bqm/shapeablebqm.pyx.src
arcondello: But totally agree we would want that in the long run.

randomir (commented):
Overall, I'm not sure this particular method belongs to the BQM/model class.

Not just because there are multiple ways to implement this conversion/penalization, but also because it implies inequality constraints and constraints over discrete/integer/etc variables will also live on the model.

Unless we plan to specialize these methods for different models, so we would convert these constraints to penalty on BQM, but store as-is on a more general model? If that's the case, it would be nicer to implement constraints in a reversable way, e.g. via some setter method. Assumes the compile step, though.

dimod/bqm/shapeablebqm.pyx.src
hsadeghidw: The tuning of `lagrange_multiplier` is not handled by the BQM class. It must be an explicit value.

dimod/bqm/shapeablebqm.pyx.src
hsadeghidw: good call. The name should change, also for DQM solver.

arcondello:
AFAIK, this is more-or-less the only way to add this type of constraint. I am also not sure it implies the we'll also support constraints that require the addition of auxiliary variables (which inequality and other variable types would).

The intention is, in the long run, to support these methods "natively" in a more general model object. I do think that the relationship between BQM and CQM models will need to be one-way without significant changes to the BQM (storing the diff maybe with a compile step like you say).

I do think that having mirroring methods on both objects is valuable though.

hsadeghidw:
I agree that the name should change to reflect the main purpose of the method. I don't think this method implies that constraints are "part" of bqm, but rather it enables adding constraints "as an objective". The use of this function is so common that it naturally should be a method, IMO.

dimod/bqm/shapeablebqm.pyx.src
```diff
@@ -914,6 +914,52 @@ cdef class cyAdj@name@BQM:
...
+        """Add a linear constraint as a quadratic objective.
+        Adds a linear constraint of the form
+        :math:`\sum_{i} a_{i} x_{i} + C = 0`
+        to the binary quadratic model as a quadratic objective.
+
+        Args:
+            terms: A list of tuples of the type (variable, bias).
+                Each tuple is evaluated to the term (bias * variable).
+                All terms in the list are summed.
+            lagrange_multiplier: The coefficient or the penalty strength
```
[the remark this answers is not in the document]
hsadeghidw: This parameter is not part of the equation above, but it's a parameter used for gluing this to the objective. So maybe we need an extra equation that shows this constraint is added to the objective?

dimod/bqm/shapeablebqm.pyx.src
```diff
@@ -914,6 +914,52 @@ cdef class cyAdj@name@BQM:
...
+        cdef Bias lbias, qbias
+        self.offset += lagrange_multiplier * constant * constant
+        for i in range(num_terms):
+            v = variables[i]
+            lbias = lagrange_multiplier * biases[i] * (2 * constant + biases[i])
+            self.bqm_.set_linear(v, lbias + self.bqm_.get_linear(v))
+            for j in range(i + 1, num_terms):
+                u = variables[j]
+                qbias = 2 * lagrange_multiplier * biases[i] * biases[j]
+                self.bqm_.set_quadratic(u, v,
```
arcondello, line 960: This will be slow, but I think we should wait until I've added some more methods to C++ (forthcoming)

arcondello:
For how generally one can specify the linear constraint (arbitrary linear weights equaling an arbitrary value), this is the only one I know of used in practice. Especially without adding auxiliary variables. Others should weight in if I am missing something.

For the utility point, we could imagine something akin to `dimod.combinations()` for this, something like
```
constraint_bqm = dimod.linear_equality_constraint(terms, lagrange, constant)
bqm.update(constraint_bqm)
```
The method is more performant, and, to me more importantly, it disambiguates the DQM and BQM. Otherwise we might need something like
```
from dimod.generators.bqm import linear_equality_constraint
from dimod.generators.dqm import linear_equality_constraint
```
Not necessarily bad, but I prefer the method as a matter of aesthetics.

rahmani15:
How about something new like this!
```
from dimod.applications import linear_equality_constraint
```
We are thinking to include a few other methods like `merge_dqm` which merge several dqm objects into one object.

arcondello:
@rahmani15 , for the record you can merge bqms already with `bqm.update(other_bqm)`, or, if you want both of the original ones preserved
```
new = bqm.copy()
new.update(other_bqm)
```

rahmani15:
@arcondello could be used to do something like this:
```
def merge_dqm(dqm_list: List[DiscreteQuadraticModel],
              lagrange_parameters: Union[float, List[float]] = 1.0):
    """Given a list of DQM objects, add linear and quadratic biases.
    Args:
        dqm_list (List[DiscreteQuadraticModel]): A list of quadratic model
        that you want to add together. The dqm objects need not to have the
        same variables, but if a variable exist in more than one dqm,
        it must have the same number of cases.
        lagrange_parameters (List[float]): A list of parameters that is
        multiplied by each dqm.
    Returns:
        (DiscreteQuadraticModel): The output DQM object that is the sum of
        the input DQMs.
    """
```

arcondello:
@rahmani15, that's a good idea. Let's take this to a github issue, I'll make a new one and add it here

hsadeghidw:
@randomir 
This method is a very general method for all linear constraints. For some specific constraints, it may be more efficient to use other techniques, but this one seems to occur a lot.

I think function works as well. but if `linear_equality_constraint` is always followed by `bqm.update()` why not make it a method?

arcondello (approved):
LGTM. @randomir do you want to do another review before we merge?

randomir (approved):

PR #790, opened 2021-02-02
operator overloading prototype for AdjDictBQM (WIP)

hsadeghidw:


dimod/bqm/adjdictbqm.py
```diff
@@ -652,3 +652,89 @@ def set_quadratic(self, u, v, bias):
 
         adj = self._adj
         adj[u][v] = adj[v][u] = bias
+
+    def __rmul__(self, other):
+        return self.__mul__(other)
+
+    def __mul__(self, other):
+        if isinstance(other, (int, float)):
```
arcondello, line 660: `numbers.Number`

dimod/bqm/adjdictbqm.py
```diff
@@ -652,3 +652,89 @@ def set_quadratic(self, u, v, bias):
...
+            bqm = AdjDictBQM(vartype=self.vartype)
```
arcondello, line 661:
```
bqm = self.copy()
bqm.scale(other)
```

dimod/bqm/adjdictbqm.py
```diff
@@ -652,3 +652,89 @@ def set_quadratic(self, u, v, bias):
...
+                bqm.add_interaction(u, v, bias * other)
+
+        elif isinstance(other, BQM):
+            if other.shape[1] > 0 or self.shape[1] > 0:
+                raise ValueError("Multiply is supported for BQMs with "
+                                 "linear terms only")
+            if other.vartype != self.vartype:
+                other = other.change_vartype(self.vartype)
+
+            bqm = AdjDictBQM(vartype=self.vartype)
```
arcondello, line 676: Again, I would probably start with a `bqm.copy()`, though at the very least should do `bqm = type(self)(vartype=self.vartype)` to accommodate subclasses. Applies other places.

dimod/bqm/adjdictbqm.py
```diff
@@ -652,3 +652,89 @@ def set_quadratic(self, u, v, bias):
...
+                        else:
+                            bqm.add_offset(bias * bias_u)
+                    else:
+                        bqm.add_interaction(u, v, bias_u * bias)
+        else:
+            raise ValueError("other must be a numeric or an instance of BQM")
+        return bqm
+
+    def __sub__(self, other):
+        return self.__add__(-other)
```
arcondello, line 697: So this will make an intermediate BQM when it calls `__neg__`, I think you probably can do this in one step.

dimod/bqm/adjdictbqm.py
```diff
@@ -652,3 +652,89 @@ def set_quadratic(self, u, v, bias):
...
+        return self.__add__(other)
+
+    def __add__(self, other):
+
+        if isinstance(other, (int, float)):
+            bqm = self.copy()
+            bqm.add_offset(other)
+
+        elif isinstance(other, BQM):
+            if other.vartype != self.vartype:
```
arcondello, line 712: `self.update(other)`

dimod/bqm/adjdictbqm.py
```diff
@@ -652,3 +652,89 @@ def set_quadratic(self, u, v, bias):
...
+
+            for (u, v), bias in other.quadratic.items():
+                bqm.add_interaction(u, v, bias)
+        else:
+            raise ValueError("other must be a numeric or an instance of BQM")
+        return bqm
+
+    def __pow__(self, power, modulo=None):
+        if modulo is not None:
+            raise ValueError("Modulo is not supported")
```
arcondello, line 727: I believe python raises a `TypeError` in this case, e.g. `[] % 4`.

dimod/bqm/adjdictbqm.py
```diff
@@ -652,3 +652,89 @@ def set_quadratic(self, u, v, bias):
...
+            for v, bias in other.linear.items():
+                bqm.add_variable(v, bias)
+
+            for (u, v), bias in other.quadratic.items():
+                bqm.add_interaction(u, v, bias)
+        else:
+            raise ValueError("other must be a numeric or an instance of BQM")
+        return bqm
+
+    def __pow__(self, power, modulo=None):
```
arcondello, line 725: One question is, should we allow this at all or just have the user do `bqm * bqm`, which is more explicit.

arcondello (commented):
Overall, this should go in the BQM abstract base class, rather than be implemented on the individual models. Since that abstract base class is being changed as part of the 0.10.x work, I wouldn't make that change right now.

arcondello:
For the CI errors, I have been getting that error sporadically since yesterday. Not 100% sure what's happening but it fixed itself at least briefly yesterday evening.

dimod/bqm/adjdictbqm.py
hsadeghidw: for the product of two `bqm`s, I think making a copy makes it convoluted, I basically have to replace every bias.

dimod/bqm/adjdictbqm.py
hsadeghidw: change this 

`__neg__` is define to be `-1*self`, but when subtracting, I don't call `__neg__` anymore.

dimod/bqm/adjdictbqm.py
hsadeghidw: removed `__pow__` altogether

PR #796, opened 2021-02-10
Updating samplers to swallow unknown kwargs and raise warning

hhtong:


dimod/reference/samplers/exact_solver.py
```diff
@@ -85,6 +86,10 @@ def sample(self, bqm):
             :obj:`~dimod.SampleSet`
 
         """
+        for kw in kwargs:
+            msg = "Ignoring unknown kwarg: {!r}".format(kw)
+            warnings.warn(msg, RuntimeWarning)
```
randomir, line 91: Perhaps subclass `UserWarning` as something like `UnknownSamplerArgumentWarning`, or similar, and raise that? FWIW, we do that in dwave-system for problem/solution warnings (e.g. `ChainBreakWarning` or `EnergyScaleWarning`).

dimod/reference/samplers/identity_sampler.py
```diff
@@ -97,6 +99,12 @@ def sample(self, bqm, *args, **kwargs):
 
 
         """
+        for kw in kwargs.copy():
+            if kw not in self.parameters:
+                msg = "Ignoring unknown kwarg: {!r}".format(kw)
+                warnings.warn(msg, RuntimeWarning)
```
randomir, line 105: Instead of repeating the same pattern, perhaps extract (some variant or portion of) it to `dimod.Sampler` base class?

dimod/core/sampler.py
```diff
@@ -146,6 +146,7 @@ def sample(self, bqm, **parameters):
 
             **kwargs:
                 See the implemented sampling for additional keyword definitions.
+                Unknown keywords will be swallowed with a raised warning.
```
JoelPasvolsky, line 149: Much as I find this phrasing more enjoyable, I think we're stuck with something prosaic such as "Unknown keywords are accepted but a warning is raised"

dimod/reference/samplers/exact_solver.py
JoelPasvolsky: I would maybe change to SamplerUnknownArgumentWarning so it's clear it's the argument, not the sampler, that is unknown

JoelPasvolsky (approved):
This PR makes me happy.

dimod/reference/samplers/exact_solver.py
arcondello, line 91: We probably also want to increase the stacklevel so that the warning shows where the `.sample()` call is.

hhtong: I added a SamplerUnknownArgWarning to sampler.py, but I'm wondering if there is a better place to put it.

arcondello: `dimod.exceptions` seems like a good place for now.

dimod/core/sampler.py
```diff
@@ -238,3 +239,23 @@ def sample_qubo(self, Q, **parameters):
         """
         bqm = BinaryQuadraticModel.from_qubo(Q)
         return self.sample(bqm, **parameters)
+
+    def validate_kwargs(self, **kwargs):
```
arcondello, line 243: I am worried that this is too general a name. I might expect it to validate the correctness of my kwargs. Maybe `validate_kwarg_set`?

dimod/core/sampler.py
```diff
@@ -238,3 +239,23 @@ def sample_qubo(self, Q, **parameters):
...
+
+        Args:
+            **kwargs:
+                Keyword arguments to be validated.
+
+        Returns:
+            dict: Updated `kwargs`
+
+        """
+        for kw in kwargs.copy():
```
arcondello, line 255: Better to do `kw in list(kwargs)` rather than copying the full thing. Shouldn't make much difference because it's a shallow copy, but might as well only make copies of the things we care about (the keys).

dimod/core/sampler.py
hhtong: What about something more descriptive like `remove_unknown_kwargs()`?

dimod/core/sampler.py
arcondello: Or even
```
for kw in [kw for kw in kwargs if kw not in self.parameters]:
    ....
```

dimod/core/sampler.py
```diff
@@ -238,3 +239,22 @@ def sample_qubo(self, Q, **parameters):
...
+        Returns:
+            dict: Updated `kwargs`
+
+        """
+        for kw in [k for k in kwargs if k not in self.parameters]:
+            msg = "Ignoring unknown kwarg: {!r}".format(kw)
+            warnings.warn(msg, SamplerUnknownArgWarning, stacklevel=3)
+            kwargs.pop(kw)
+
+        return kwargs
```
randomir, line 260: I find this confusing. `kwargs` are modified inplace, and that modified/original dict is returned.

Can we either:
1) return a shallow copy, with unknown items removed, or
2) not return anything (like `dict.update`)

randomir: Please ignore my previous comment. `kwargs` are constructed *on call*, even if expanded from `**kwargs`, so the change to them can't be propagated to the caller.

randomir (approved):

dimod/core/sampler.py
```diff
@@ -238,3 +239,22 @@ def sample_qubo(self, Q, **parameters):
         """
         bqm = BinaryQuadraticModel.from_qubo(Q)
         return self.sample(bqm, **parameters)
+
+    def remove_unknown_kwargs(self, **kwargs):
```
arcondello, line 243: I guess this has already been merged but I would say that this also needs to be documented somewhere in the "creating new samplers" documentation. Especially if we're expecting all samplers to do it. An example would be https://github.com/dwavesystems/dimod/pull/796/files#diff-4f1806f800c56e578e5ede71ae3a6c117743302bebb7aeddbe29404040ad052bR85 but probably it needs to be noted elsewhere.

PR #799, opened 2021-02-16
Update sampler docs for unknown kwarg handling

hhtong:


dimod/core/sampler.py
```diff
@@ -23,6 +23,10 @@
 or :meth:`~.Sampler.sample_qubo`. The :class:`.Sampler` class provides the complementary
 methods as mixins and ensures consistent responses.
 
+All sample methods must be able to accept unknown kwargs with a raised warning.
```
JoelPasvolsky, line 26: What about a slight rephrasing to something like, 
"Implemented sample methods must accept, and warn on, unknown keyword arguments ``**kwargs``.

dimod/core/sampler.py
```diff
@@ -23,6 +23,10 @@
...
+This means that all sample methods must have the `**kwargs` parameter and
```
JoelPasvolsky, line 27: "... parameter and call the :meth:`~.Sampler.remove_unknown_kwargs` method"?

dimod/core/sampler.py
```diff
@@ -115,6 +121,8 @@ class Sampler(metaclass=SamplerABCMeta):
     Provides all methods :meth:`~.Sampler.sample`, :meth:`~.Sampler.sample_ising`,
     :meth:`~.Sampler.sample_qubo` assuming at least one is implemented.
 
+    Also includes utility method :meth:`~.Sampler.remove_unknown_kwargs`, which
+    should be called in each sample method.
```
JoelPasvolsky, line 125: "... each implemented sample method"?

dimod/core/sampler.py
hhtong: That sounds better, thanks!

dimod/core/sampler.py
```diff
@@ -23,6 +23,10 @@
 or :meth:`~.Sampler.sample_qubo`. The :class:`.Sampler` class provides the complementary
 methods as mixins and ensures consistent responses.
 
+Implemented sample methods must accept, and warn on, unknown keyword arguments
+`**kwargs`. This means that all sample methods must have the `**kwargs`
+parameter and call the :meth:`~.Sampler.remove_unknown_kwargs` method.
```
arcondello, line 28: The user technically does not have to call the method. Maybe 
```
...must have `**kwargs` parameter. The :meth:`~.Sampler.remove_unknown_kwargs` convenience method is provided for this purpose.
```
or something like that?

hhtong: That's true. I've updated with this change.

arcondello (approved):

dimod/core/sampler.py
```diff
@@ -115,6 +122,8 @@ class Sampler(metaclass=SamplerABCMeta):
     Provides all methods :meth:`~.Sampler.sample`, :meth:`~.Sampler.sample_ising`,
     :meth:`~.Sampler.sample_qubo` assuming at least one is implemented.
```
JoelPasvolsky, line 124: According to comment https://github.com/dwavesystems/dimod/pull/799#discussion_r577249029, the implementation does not have to call the method but below you say ", which  should be called in each implemented sample method"

JoelPasvolsky (approved):

PR #800, opened 2021-02-17
Deprecate Variables.index mapping and add permissive kwarg

arcondello:
This kwarg argument is useful for other refactoring work I am doing but wanted to break it out so it's easier to review.

### DIFF
--- a/dimod/variables.py
+++ b/dimod/variables.py
@@ -221,8 +221,28 @@ class Variables(abc.Sequence, abc.Set):
         # everything is unique
         return int(v in self)
 
-    def index(self, v):
-        if v not in self:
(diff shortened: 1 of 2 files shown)


dimod/variables.py
```diff
@@ -271,15 +221,27 @@ def count(self, v):
         # everything is unique
         return int(v in self)
 
-    # index is set by __init__ for now
+    def index(self, v, permissive=False):
+        """Return the index of v.
+
+        Args:
+            v (hashable): A variable.
+            permissive (bool, optional, default=False): If True, the variable
```
JoelPasvolsky, line 229: Add blank line

PR #805, opened 2021-03-04
Sample sets from nonblocking_sample_method can signal .done()

arcondello:


arcondello:
For context, the sample set returned by `ScaleComposite` was always indicated that it was done (see [code](https://github.com/dwavesystems/dimod/blob/dc98404439114f075aa7c0ff979ad30079eab2e4/dimod/sampleset.py#L910)).
This then led to `SampleSet.change_vartype` to resolve eagerly (see [code](https://github.com/dwavesystems/dimod/blob/dc98404439114f075aa7c0ff979ad30079eab2e4/dimod/sampleset.py#L1134)).

dimod/reference/composites/scalecomposite.py
```diff
@@ -135,7 +135,7 @@ def sample(self, bqm, scalar=None, bias_range=1, quadratic_range=None,
 
         sampleset = self.child.sample(bqm, **parameters)
 
-        yield  # wait until sampleset is read for the remaining part
+        yield sampleset  # so that SampleSet.done() works
```
randomir, line 138: How about adding a regression test for this?

randomir (approved):
LGTM!

dimod/reference/composites/scalecomposite.py
arcondello: As far as I know the only place we actually used this (so far) was in `ScaleComposite`. Also, the previous behaviour was explicitly to ignore whatever was yielded first. I do have an integration test ready for dwave-system that I'll add once this is deployed.

PR #806, opened 2021-03-06
dropping variables from a sampleset

pau557:
It simply removes variables from a sampleset

### DIFF
--- a/dimod/sampleset.py
+++ b/dimod/sampleset.py
@@ -20,7 +20,7 @@ import numbers
 import collections.abc as abc
 
 from collections import namedtuple
-from typing import Any, Callable
+from typing import Any, Callable, Iterable
 
 import numpy as np
 from numpy.lib import recfunctions
@@ -35,11 +35,20 @@ from dimod.serialization.utils import (pack_samples as _pack_samples,
                                        serialize_ndarrays,
                                        deserialize_ndarrays)
 from dimod.sym import Sense
+from dimod.typing import Variable
 from dimod.variables import Variables, iter_deserialize_variables
 from dimod.vartypes import as_vartype, Vartype, DISCRETE
 from dimod.views.samples import SampleView, SamplesArray
 
-__all__ = ['append_data_vectors', 'append_variables', 'as_samples', 'concatenate', 'SampleSet']
+__all__ = ['append_data_vectors',
+           'append_variables',
+           'as_samples',
+           'concatenate',
+           'drop_variables',
+           'keep_variables',
+           'SampleSet',
+           ]
+
 
 def append_data_vectors(sampleset, **vectors):
     """Create a new :obj:`.SampleSet` with additional fields in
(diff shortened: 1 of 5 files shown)


tests/test_sampleset.py
```diff
@@ -489,6 +489,16 @@ def test_invalid(self):
...
+class TestDropVariables(unittest.TestCase):
+    def test_variables_were_dropped(self):
+        samples = np.triu(np.ones((5, 5)))
+        labels = 'abcde'
+        sampleset = dimod.SampleSet.from_samples((samples, labels), vartype='BINARY', energy=1)
+        variables_to_drop="ae"
+        expected_variables = set([label for label in labels if label not in variables_to_drop])
+        obtained_variables = set(dimod.sampleset.drop_variables(sampleset, variables_to_drop).variables)
+        self.assertTrue(obtained_variables==expected_variables)
+
```
arcondello, line 501: Can we also check that the correct sample values were returned?

tests/test_sampleset.py
```diff
@@ -489,6 +489,16 @@ def test_invalid(self):
         with self.assertRaises(ValueError):
             sampleset = dimod.append_data_vectors(sampleset, sets=[{0}, {1}])
 
+class TestDropVariables(unittest.TestCase):
+    def test_variables_were_dropped(self):
+        samples = np.triu(np.ones((5, 5)))
+        labels = 'abcde'
+        sampleset = dimod.SampleSet.from_samples((samples, labels), vartype='BINARY', energy=1)
```
arcondello, line 496: Also, need to either import `dimod` at the top or just drop the `dimod.` here.

dimod/sampleset.py
```diff
@@ -453,6 +453,27 @@ def infer_vartype(samples_like):
...
     raise ValueError("given samples_like is of an unknown vartype")
 
+def drop_variables(sampleset, variables_to_drop):
+    """Drop variables from a sampleset
+
+    Args:
+        sampleset (:obj:`.SampleSet`):
+            Original sampleset.
+
+        variables_to_drop (iter):
```
arcondello, line 463: should be `container`, not `iter`. An iterator won't actually work in this context.

dimod/sampleset.py
```diff
@@ -453,6 +453,27 @@ def infer_vartype(samples_like):
 
     raise ValueError("given samples_like is of an unknown vartype")
 
+def drop_variables(sampleset, variables_to_drop):
+    """Drop variables from a sampleset
+
+    Args:
+        sampleset (:obj:`.SampleSet`):
```
randomir, line 460: One should [use `class`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-python-objects) for a specific type.
```suggestion
        sampleset (:class:`.SampleSet`):
```

dimod/sampleset.py
```diff
@@ -453,6 +453,27 @@ def infer_vartype(samples_like):
 
     raise ValueError("given samples_like is of an unknown vartype")
 
+def drop_variables(sampleset, variables_to_drop):
+    """Drop variables from a sampleset
```
randomir, line 457: We should be clear we are returning _a new_ sampleset (a copy), not modifying the original one.

dimod/sampleset.py
```diff
@@ -453,6 +453,27 @@ def infer_vartype(samples_like):
...
+
+    Args:
+        sampleset (:obj:`.SampleSet`):
+            Original sampleset.
+
+        variables_to_drop (iter):
+            Iterable containing the variables to be dropped
+
+    Returns:
+        :obj:`.SampleSet`: A sampleset without the variables to be dropped.
```
randomir, line 467:
```suggestion
        :class:`.SampleSet`: A sampleset without the variables to be dropped.
```

dimod/sampleset.py
```diff
@@ -453,6 +453,27 @@ def infer_vartype(samples_like):
...
+    """
+    
+    reduced_variables = [var for var in sampleset.variables if var not in variables_to_drop]
+    reduced_samples = sampleset.samples()[:, reduced_variables]
+
+    return dimod.SampleSet.from_samples((reduced_samples, reduced_variables), 
+                                        vartype=sampleset.vartype, 
+                                        energy=sampleset.record.energy,
+                                        info=sampleset.info)
```
randomir, line 476: Should make a (deep)copy of the original info dict.

dimod/sampleset.py
```diff
@@ -453,6 +453,27 @@ def infer_vartype(samples_like):
...
+            Original sampleset.
+
+        variables_to_drop (iter):
+            Iterable containing the variables to be dropped
+
+    Returns:
+        :obj:`.SampleSet`: A sampleset without the variables to be dropped.
+    """
+    
+    reduced_variables = [var for var in sampleset.variables if var not in variables_to_drop]
```
randomir, line 470: I'd "settify" `variables_to_drop` before doing N existence queries.

dimod/sampleset.py
randomir, line 463: Consider renaming to `variables`. Should be clear from the context, and it's easier on the eyes.

dimod/sampleset.py
```diff
@@ -450,6 +459,58 @@ def infer_vartype(samples_like):
...
+    return keep_variables(sampleset, sampleset.variables - variables)
+
+
+def keep_variables(sampleset: 'SampleSet', variables: Iterable[Variable]) -> 'SampleSet':
+    """Return a new sample set with only the given variables.
+
+    Args:
+        sampleset: A sample set.
+        variables: The variables to be kept. Must be a subset of the variables
+            in the sample set.
```
randomir, line 484: For symmetry with `drop_variables`, why not allow variables not in the sampleset?

randomir (approved):
LGTM

dimod/sampleset.py
arcondello: In this case the asymmetry is deliberate. I think of it as "drop any of these variables" and "keep all of these variables". So if I was able to provide variables not in the sample set to keep, I would expect the created sample set to have all of the ones specified.

randomir: Sure, I get that. Personally, I would appreciate consistency, so at least for drop to require a subset... But I really don't have strong feeling about it. :shrug:

PR #811, opened 2021-03-24
Fix missing .pxd files in windows wheels

arcondello:
setuptools in windows couldn't handle the trailing `/`.

I tested with a different CI script. I don't think it makes sense to add it permanently to the CI as a test


randomir (commented):
:tada:

hbarovertwo (approved):
LGTM

PR #812, opened 2021-03-25
Cythonize the variables object

arcondello:
Results in 10-20x speedup on many common operations in SampleSet and DQM classes. Also opens up the path for accessing it from cython for even more improvement for performance-critical applications.

### DIFF
--- a/dimod/cyvariables.pyx
+++ b/dimod/cyvariables.pyx
@@ -51,13 +51,12 @@ cdef class cyVariables:
         if self._is_range():
             v = pyidx
         else:
-            # I am reasonably confident that this is safe and it's faster
-            # than self._index_to_label.get
+            # faster than self._index_to_label.get
             obj = PyDict_GetItemWithError(self._index_to_label, pyidx)
             if obj == NULL:
                 v = pyidx
             else:
-                v = <object>obj
+                v = <object>obj  # correctly handles the ref count
 
         return v
 
@@ -179,7 +178,7 @@ cdef class cyVariables:
 
         # handle other numeric types
         if isinstance(v, Number):
-            v_int = int(v)  # assume this is safe...
+            v_int = int(v)  # assume this is safe because it's a number
             if v_int == v:
                 return self._count_int(v_int)  # it's an integer afterall!
 
@@ -216,12 +215,11 @@ cdef class cyVariables:
         if self._is_range():
             return v if PyLong_Check(v) else int(v)
 
-        # I am reasonably confident that this is safe and it's faste
-        # than self._label_to_index.get
+        # faster than self._label_to_index.get
         cdef PyObject* obj = PyDict_GetItemWithError(self._label_to_index, v)
         if obj == NULL:
             pyobj = v
         else:
-            pyobj = <object>obj
+            pyobj = <object>obj  # correctly updates ref count
 
         return pyobj if PyLong_Check(pyobj) else int(pyobj)
--- a/dimod/cyvariables.pyx
+++ b/dimod/cyvariables.pyx
@@ -66,6 +66,22 @@ cdef class cyVariables:
     cpdef object _append(self, object v=None, bint permissive=False):
         """Append a new variable.
 
+        Args:
+            v (hashable, optional):
+                Add a new variable. If `None`, a new label will be generated.
+
+            permissive (bool, optional, default=False):
+                If `False`, appending a variable that already exists will raise
+                a `ValueError`. If `True`, appending a variable that already
+                exists will not change the container.
+
+        Returns:
+            hashable: The label of the appended variable.
+
+        Raises:
+            ValueError: If the variable is present and `permissive` is
+            False.
+
         This method is semi-public. it is intended to be used by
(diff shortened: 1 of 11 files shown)


dimod/cyvariables.pyx
```diff
@@ -0,0 +1,263 @@
...
+
+    def __len__(self):
+        return self._stop
+
+    cpdef object _append(self, object v=None, bint permissive=False):
+        """Append a new variable.
+
+        Args:
+            v (hashable, optional):
+                Add a new variable. If `None`, a new label will be generated.
```
JoelPasvolsky, line 71: Be nice to say something on how, something like,  "If `None`, generates a new label that is the lowest currently unused integer ..."

dimod/cyvariables.pyx
```diff
@@ -0,0 +1,263 @@
...
+
+        if idx != v:
+            self._label_to_index[v] = idx
+            self._index_to_label[idx] = v
+
+        self._stop += 1
+        return v
+
+    cpdef bint _is_range(self):
+        """Return whether the Variables are current labelled [0, n)."""
```
JoelPasvolsky, line 113: variables

dimod/cyvariables.pyx
```diff
@@ -0,0 +1,263 @@
...
+                    self._label_to_index[new] = idx
+                    self._index_to_label[idx] = new  # overwrites old idx
+                else:
+                    self._index_to_label.pop(idx, None)
+
+    def _relabel_as_integers(self):
+        """Relabel the variables as integers in-place.
+
+        Returns:
+            dict: The mapping that would restore the original labels.
```
JoelPasvolsky, line 186: What does "original" mean here? Are the `self._label_to_index` and `self._index_to_label` keeping a history of label indices?

dimod/cyvariables.pyx
```diff
@@ -0,0 +1,263 @@
...
+        if self._is_range():
+            return 0 <= vi < self._stop
+
+        # need to make sure that we're not using the integer elsewhere
+        return (0 <= vi < self._stop
+                and not PyDict_Contains(self._index_to_label, v)
+                or PyDict_Contains(self._label_to_index, v))
+
+    cpdef Py_ssize_t count(self, object v) except -1:
+        """Return the number of times `v` appears in Variables.
```
JoelPasvolsky, line 210: Either "in the variables" or "in the :class:`.Variables` class", otherwise it looks 17th-century

dimod/variables.py
```diff
@@ -12,6 +12,20 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+"""
+A class and utilities for encoding variable objects.
+
+The :class:`Variables` class is intended to be used as an attribute of other
+classes, such as :class:`.DiscreteQuadraticModel` and :class:`.SampleSet`.
+
+The goals of the class are:
```
JoelPasvolsky, line 21: requirements?

docs/reference/variables.rst
```diff
@@ -0,0 +1,49 @@
...
+-------
+
+.. autosummary::
+   :toctree: generated/
+
+    ~Variables.index
+    ~Variables.count
+    ~Variables.to_serializable
+
+Mutation Methods
```
JoelPasvolsky, line 34: Can the internal nature be more prominent? Maybe add Private to the title or at least bold the warning below.

dimod/cyvariables.pyx
arcondello: I am not sure I want to "lock in" a particular behavior for backwards compatibility reasons.

docs/reference/variables.rst
arcondello: All the methods are prepended by a `_` which I think is a pretty clear signal in a code context. In the docs context, I am not sure bolding it would make it any more likely to be read. It's not like this section is hidden among a lot of other text. We could say "semi-private" or something, but they are not strictly speaking private...

dimod/cyvariables.pyx
arcondello: You missed the `currently`!

dimod/cyvariables.pyx
```diff
@@ -0,0 +1,278 @@
...
+        """Remove the last variable.
+
+        This method is semi-public. it is intended to be used by
+        classes that have :class:`.Variables` as an attribute, not by the
+        the user.
+        """
+        if not self:
+            raise IndexError("Cannot pop when Variables is empty")
+
+        self._stop = idx = self._stop - 1
```
randomir, line 153: Will it handle this:
```python
variables = Variables('x')
variables._pop()
variables._pop()
```
Maybe it's a cython thing, but I'm not sure why would `self` be falsey after line 2 above?

randomir (approved):
LGTM

dimod/cyvariables.pyx
arcondello: Uses [`__len__`](https://docs.python.org/3/reference/datamodel.html#object.__bool__) in that case.

PR #814, opened 2021-04-08
Setup refresh

### LINKED ISSUE #678: Specify build enivornment by PEP 518 standard
https://www.python.org/dev/peps/pep-0518/

arcondello:
Significantly rework/simplify the setup script.

Closes #678

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,8 @@
+[build-system]
+requires = [
+    "setuptools",
+    "wheel",
+    "Cython>=0.29.21,<3.0",
+    "numpy==1.19.4",
+]
+build-backend = "setuptools.build_meta"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-    "setuptools",
+    "setuptools>=42.0.0",
     "wheel",
     "Cython>=0.29.21,<3.0",
     "numpy==1.19.4",
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,40 @@
 [metadata]
-description-file = README.rst
+classifiers =
+    License :: OSI Approved :: Apache Software License
+    Operating System :: OS Independent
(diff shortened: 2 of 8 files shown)


setup.cfg
```diff
@@ -1,2 +1,40 @@
 [metadata]
-description-file = README.rst
+classifiers =
+    License :: OSI Approved :: Apache Software License
+    Operating System :: OS Independent
+    Programming Language :: Python :: 3 :: Only
+    Programming Language :: Python :: 3.5
```
randomir, line 6: Not anymore.

pyproject.toml
```diff
@@ -0,0 +1,8 @@
+[build-system]
+requires = [
+    "setuptools",
```
randomir, line 3: I'd lower-bound setuptools to avoid some issues we had in past. Although, I can't remember what those were, but I remember it was something. :smile:  Environment markers, right?

randomir (approved):
LGTM

pyproject.toml
arcondello: Set it to `>=42.0.0` for now. A bit arbitrary but that's the version that started supporting pep518 so seemed natural :)

randomir: I've found the lower-bound I had in mind -- looks like [it's 31.0 due to namespace packages support](https://github.com/dwavesystems/dwave-cloud-client/pull/43).

OTOH, I'm not sure where is 42.0.0 coming from? Looking at https://github.com/pypa/setuptools/issues/651, seems like PEP-518 support was not needed/not implemented in setuptools.

randomir: Then again, I see from https://github.com/pypa/setuptools/commit/89155abb4222cf5a9dc81120e5c71e26b5af68f9 the full PEP-420 support (native namespace packages) are only supported since 40.1.0.

arcondello: I think the number came from [their changlog](https://setuptools.readthedocs.io/en/latest/history.html#v42-0-0).

PR #819, opened 2021-05-04
C++ BQM.add_quadratic from COO

arcondello:


dimod/include/dimod/utils.h
```diff
@@ -27,7 +29,85 @@ namespace utils {
...
+        stack[++top] = low;
+        stack[++top] = high;
+
+        while (top >= 0) {
+            high = stack[top--];
+            low = stack[top--];
+
+            std::int64_t pivot_choice_index = (low + high + 1) / 2;
+            if (pivot_choice_index != high) {
+                T1 control_temp = control[pivot_choice_index];
```
amahmudwave, line 64: I wonder how in the world this code survived all the randomized tests. , I am swapping the control variables, but not the response variables. here.  just need to add the following lines, but I will follow up with more testing after making the changes myself 

```
            T2 response_temp = response[pivot_choice_index];
            response[pivot_choice_index] = response[high];
            response[high] = response_temp;
```

dimod/include/dimod/quadratic_model.h
```diff
@@ -252,9 +254,38 @@ class Neighborhood {
...
+    /// Sort the neighborhood and sum the biases of duplicate variables.
+    void sort_and_sum() {
+        if (!std::is_sorted(neighbors.begin(), neighbors.end())) {
+            utils::zip_sort(neighbors, quadratic_biases);
+        }
+
+        // now remove any duplicates, adding the biases
+        size_type i = 0;
+        while (i + 1 < size()) {
+            if (neighbors[i] == neighbors[i + 1]) {
```
amahmudwave, line 276: The better thing to do is to go allocate memory using malloc of the same size as the quadratic biases and and neighbors, write into the buffer the summarized results,  swap the current neighbors and quadratic_biases with empty vectors to ensure memory release( std::vector<index_type>().swap(neighbors) ) . Then use std::insert to insert from the allocated memory. 

Saves you from doing potentially n^2 work when there is a duplicate for each element.

dimod/include/dimod/quadratic_model.h
```diff
@@ -555,6 +586,73 @@ class BinaryQuadraticModel : public QuadraticModelBase<Bias, Index> {
...
+
+            if ((size_t)max_label >= base_type::num_variables()) {
+                resize(max_label + 1);
+            }
+        } else if (length < 0) {
+            throw std::out_of_range("length must be positive");
+        }
+
+        // count the number of elements to be inserted into each
+        std::vector<size_type> counts(base_type::num_variables());
```
amahmudwave, line 619: Need to give the default value , but it may be zero by default I think.

dimod/include/dimod/quadratic_model.h
```diff
@@ -555,6 +586,73 @@ class BinaryQuadraticModel : public QuadraticModelBase<Bias, Index> {
...
+                    *std::max_element(row_iterator, row_iterator + length),
+                    *std::max_element(col_iterator, col_iterator + length));
+
+            if ((size_t)max_label >= base_type::num_variables()) {
+                resize(max_label + 1);
+            }
+        } else if (length < 0) {
+            throw std::out_of_range("length must be positive");
+        }
+
```
amahmudwave, line 617: Do you want to do an early exit if length == 0 ?

amahmudwave (changes requested):

dimod/include/dimod/utils.h
amahmudwave: Found the reason why my tests never caught it. I think it will seem familiar to you 

```
112         for (int i = 0; i < num_elements; i++) {
113             if (v_pair[i].first != control[i] &&
114                 v_pair[i].second != response[i]) {
115                 std::cout << "Problem Pairs not sorted properly" << std::endl;
116             }
117         }
```
 
The && should be a ||. :)

amahmudwave: The suggested code snipped fixed it, I also pushed my fix in the repo I originally had.

dimod/include/dimod/quadratic_model.h
amahmudwave: You could do other optimizations like go over the array once and see if there is any duplicate at all and only then go for the copying and allocating ...

amahmudwave: On second thoughts you do not need an extra array, you just need two indices and one temp neighbor storing the last seen neighbor.  one index pointing to the point you are at, one pointing upto the point you just stored the unique values    if duplicate add to the last pointed unique value, else increment the unique value counter and store to that point. In the end resize upto the uniquely pointed value. No extra memory overhead :)

dimod/include/dimod/quadratic_model.h
arcondello: defaults to 0 but I agree we might as well be explicit.

dimod/include/dimod/quadratic_model.h
arcondello: The solution works for `length==0` as-is and I don't think it's a case worth optimizing for.

dimod/include/dimod/quadratic_model.h
```diff
@@ -252,9 +254,52 @@ class Neighborhood {
...
+    void reserve(index_type n) {
+        neighbors.reserve(n);
+        quadratic_biases.reserve(n);
+    }
+
     /// Return the size of the neighborhood.
     size_type size() const { return neighbors.size(); }
 
+    /// Sort the neighborhood and sum the biases of duplicate variables.
+    void sort_and_sum() {
```
amahmudwave, line 268: Do you want to move this to the utils like the zip sort one ? Up to you.

dimod/include/dimod/utils.h
```diff
@@ -27,7 +29,85 @@ namespace utils {
...
+        stack[++top] = low;
+        stack[++top] = high;
+
+        while (top >= 0) {
+            high = stack[top--];
+            low = stack[top--];
+
+            std::int64_t pivot_choice_index = (low + high + 1) / 2;
+            if (pivot_choice_index != high) {
+                T2 response_temp = response[pivot_choice_index];
```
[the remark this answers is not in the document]
arcondello: this is what I get for being lazy and not adding tests :laughing: . Ok I will do it the right way...

dimod/include/dimod/quadratic_model.h
arcondello: Eh, for now I'll leave it be and move it when it becomes relevant.

amahmudwave (approved):

amahmudwave (approved):

PR #820, opened 2021-05-06
Added dqm exact solver

necaisej:


dimod/reference/samplers/exact_solver.py
```diff
@@ -159,6 +159,45 @@ def sample_poly(self, polynomial, **kwargs):
         """
         return ExactSolver().sample(polynomial, **kwargs)
 
+class ExactSolverDQM(Sampler):
```
arcondello, line 162: Nit - I would call this `ExactDQMSolver` to be consistent with the convention we use for `LeapHybridDQMSampler`

dimod/reference/samplers/exact_solver.py
```diff
@@ -159,6 +159,45 @@ def sample_poly(self, polynomial, **kwargs):
...
+    """A simple exact solver for testing and debugging code using your local CPU.
+
+    Notes:
+        This solver becomes slow for problems with 18 or more
```
arcondello, line 166: Did we test at this size or is it just copied from the above?

dimod/reference/samplers/exact_solver.py
```diff
@@ -159,6 +159,45 @@ def sample_poly(self, polynomial, **kwargs):
...
+        Args:
+            dqm (:obj:`~dimod.DiscreteQuadraticModel`):
+                Discrete quadratic model to be sampled from.
+
+        Returns:
+            :obj:`~dimod.SampleSet`
+
+        """
+        kwargs = self.remove_unknown_kwargs(**kwargs)
+
```
arcondello, line 189: extra space

dimod/reference/samplers/exact_solver.py
```diff
@@ -159,6 +159,45 @@ def sample_poly(self, polynomial, **kwargs):
...
+        
+        
+        n = dqm.num_variables()
+        if n == 0:
+            return SampleSet.from_samples([], 'DISCRETE', energy=[])
+
+        possible_samples = as_samples(_all_cases_dqm(dqm))
+        energies = dqm.energies(possible_samples[0])
+        
+        response = SampleSet.from_samples(possible_samples, 'DISCRETE', energies).relabel_variables(dict(enumerate(dqm.variables)))
```
arcondello, line 198: rather than relabeling, you can just pass in `SampleSet.from_samples((samples, labels), 'DISCRETE',...`

dimod/reference/samplers/exact_solver.py
```diff
@@ -159,6 +159,45 @@ def sample_poly(self, polynomial, **kwargs):
...
+        """
+        kwargs = self.remove_unknown_kwargs(**kwargs)
+        
+        
+        n = dqm.num_variables()
+        if n == 0:
+            return SampleSet.from_samples([], 'DISCRETE', energy=[])
+
+        possible_samples = as_samples(_all_cases_dqm(dqm))
+        energies = dqm.energies(possible_samples[0])
```
arcondello, line 196: Maybe I am missing something, but why `[0]`?

dimod/reference/samplers/exact_solver.py
```diff
@@ -177,3 +216,18 @@ def _graycode(bqm):
...
 
     return samples
+
+
+def _all_cases_dqm(dqm):
+    """Get a numpy array containing all possible samples as lists of integers"""
+    # developer note: there may be better ways to do this, but because we're
+    # limited in performance by the energy calculation, this is probably fine
+    
+    dqm_ = dqm.copy()
```
arcondello, line 226: Don't need to do a copy, could just do
```
cases = [range(dqm.num_cases(v)) for v in dqm.variables]
```
to get the same effect

arcondello (changes requested):
Needs tests

dimod/reference/samplers/exact_solver.py
randomir: If we actually tested it, then a critical piece is missing -- number of cases per variable.

necaisej: The largest I tested was 10 variables and 4 cases, haven't tested it further. I assumed if BQM isn't feasible past 18, this can't handle beyond the product of number of cases over all variables is approx 2^18. They both calculate the energy of all possible configurations

dimod/reference/samplers/exact_solver.py
JoelPasvolsky: And also 'ExactPolySolver'

JoelPasvolsky (commented):
Also needs to update https://github.com/dwavesystems/dimod/blob/main/docs/reference/sampler_composites/samplers.rst

dimod/reference/samplers/exact_solver.py
```diff
@@ -159,6 +159,53 @@ def sample_poly(self, polynomial, **kwargs):
...
+        self.parameters = {}
+
+    def sample_ising(self, h, J, **parameters):
+        return ExactSolver().sample_ising(h, J, **parameters)
+
+    def sample_qubo(self, Q, **parameters):
+        return ExactSolver().sample_qubo(Q, **parameters)
+
+    def sample(self, dqm, **parameters):
+        return self.sample_dqm(dqm, **parameters)
```
mcfarljm, line 184: If all the class does is sample DQM's, then I like the shortened version (as opposed to or in addition to `sample_dqm`), but it wouldn't be consistent with `LeapHybridDQMSampler`.  On the other hand, if you keep `sample_ising` and `sample_qubo`, `sample` might be ambiguous in this context.

dimod/reference/samplers/exact_solver.py
```diff
@@ -159,6 +159,53 @@ def sample_poly(self, polynomial, **kwargs):
...
+
+    def __init__(self):
+        self.properties = {}
+        self.parameters = {}
+
+    def sample_ising(self, h, J, **parameters):
+        return ExactSolver().sample_ising(h, J, **parameters)
+
+    def sample_qubo(self, Q, **parameters):
+        return ExactSolver().sample_qubo(Q, **parameters)
```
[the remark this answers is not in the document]
arcondello: Agreed. I would only keep `.sample_dqm` to be consistent with the DQM solver.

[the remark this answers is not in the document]
necaisej: I believe there is an issue of instantiating the sampler using the Sampler base class without sample_ising and sample_qubo methods; otherwise I agree with you both.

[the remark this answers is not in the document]
mcfarljm: Do you need to subclass from `Sampler`?  It looks like `LeapHybridDQMSampler` does not.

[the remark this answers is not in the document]
necaisej: My bad, you only need one of `sample`, `sample_ising`, `sample_qubo`. I can remove the latter two and still make use of `Sampler`.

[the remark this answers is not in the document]
arcondello: `Sampler` is only defined for BQM samplers (if we had known we were going to support other types we would have called it `BQMSampler` but :shrug: ). Agree with John, it just shouldn't subclass.

dimod/reference/samplers/exact_solver.py
arcondello: Right - I imagine this number should be smaller.

dimod/reference/samplers/exact_solver.py
```diff
@@ -159,6 +161,63 @@ def sample_poly(self, polynomial, **kwargs):
...
+
+    """
+    properties = None
+    parameters = None
+
+    def __init__(self):
+        self.properties = {}
+        self.parameters = {}
+
+    def remove_unknown_kwargs(self, **kwargs):
```
arcondello, line 179: I think `remove_unknown_kwargs = Sampler.remove_unknown_kwargs` should work fine.

dimod/testing/sampler.py
```diff
@@ -57,6 +57,16 @@ def test_sample_qubo(self, sampler, Q):
         self.assertAlmostEqual(dimod.qubo_energy(sample, Q), en)
 
 
+def test_sample_dqm(self, sampler, dqm):
```
arcondello, line 60: What is this?

arcondello: I still don't see where this function is being used.

docs/reference/sampler_composites/samplers.rst
```diff
@@ -44,6 +44,30 @@ Methods
...
+
+.. autoclass:: ExactDQMSolver
+
+Methods
+~~~~~~~
+
+.. autosummary::
+   :toctree: generated/
+
+   ExactDQMSolver.sample
```
arcondello, line 68: need to remove `.sample`.

dimod/reference/samplers/exact_solver.py
```diff
@@ -159,6 +161,47 @@ def sample_poly(self, polynomial, **kwargs):
...
+
+        Args:
+            dqm (:obj:`~dimod.DiscreteQuadraticModel`):
+                Discrete quadratic model to be sampled from.
+
+        Returns:
+            :obj:`~dimod.SampleSet`
+
+        """
+        remove_unknown_kwargs = Sampler.remove_unknown_kwargs(self, **kwargs)
```
arcondello, line 193: If you're just calling it in the method, you don't need the LHS. Can just do `Sampler.remove_unknown_kwargs(self, **kwargs)`

dimod/reference/samplers/exact_solver.py
```diff
@@ -21,13 +21,15 @@
     energy for every possible sample, they are very slow.
 """
 import numpy as np
+import warnings
```
arcondello, line 24: I don't think this is used.

arcondello (approved):

PR #827, opened 2021-05-21
Refactor c++ iterators

arcondello:
For neighborhood iterators:
* Made them random access
* Move them out from being nested in Neighborhood
* Fix warnings generated from empty constructor
* Implement operator->()

Also added a new quadratic bias iterator.

### DIFF
--- a/dimod/include/dimod/quadratic_model.h
+++ b/dimod/include/dimod/quadratic_model.h
@@ -31,6 +31,12 @@ enum Vartype {
     INTEGER  ///< Variables that are integer valued.
 };
 
+template <class Bias, class Index>
+class NeighborhoodIterator;
+
+template <class Bias, class Index>
+class ConstNeighborhoodIterator;
+
 /**
  * Used internally by QuadraticModelBase to sparsely encode the neighborhood of
  * a variable.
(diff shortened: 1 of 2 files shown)


dimod/include/dimod/quadratic_model.h
```diff
@@ -322,8 +229,233 @@ class Neighborhood {
...
+    neighborhood_type* neighborhood_;
+    index_type i_;
+
+    value_type* pair_;
+
+    void update_pair() {
+        if (pair_) {
+            delete pair_;
+        }
+        if (i_ < (index_type)neighborhood_->size()) {
```
amahmudwave, line 338: what if i < 0

dimod/include/dimod/quadratic_model.h
```diff
@@ -474,6 +617,81 @@ class QuadraticModelBase {
     std::vector<Neighborhood<bias_type, index_type>> adj_;
 
     bias_type offset_;
+
+    friend class ConstQuadraticIterator<Bias, Index>;
+};
+
+template <class Bias, class Index>
+class ConstQuadraticIterator {
```
amahmudwave, line 625: I am not the user of this class but I do not know how much it helps, having to remember the internal working of this class seems like a hard thing for me, I would just iterate over the neighborhood using the other tools above.

dimod/include/dimod/quadratic_model.h
```diff
@@ -322,8 +229,233 @@ class Neighborhood {
...
+        }
+        if (i_ < (index_type)neighborhood_->size()) {
+            pair_ = new value_type(neighborhood_->neighbors[i_],
+                                   neighborhood_->quadratic_biases[i_]);
+        }
+    }
+};
+
+template <class Bias, class Index>
+class ConstNeighborhoodIterator {
```
amahmudwave, line 346: can you do something like this ? 

template<clas Bias, class Index> 
BaseNeighborhoodIterator  -> define only this one. 

template<class Bias, class Index>
using NeighborhoodIterator = BaseNeighborhoodIterator<Bias, Index>

template <class Bias, class Index> 
using ConstNeighborhoodIteraotr = BaseNeighborhoodIterator<const Bias, Index>

dimod/include/dimod/quadratic_model.h
```diff
@@ -322,8 +229,233 @@ class Neighborhood {
...
 };
 
+template <class Bias, class Index>
+class NeighborhoodIterator {
+ public:
+    using bias_type = Bias;
+    using index_type = Index;
+    using neighborhood_type = Neighborhood<Bias, Index>;
+
+    using value_type = std::pair<const index_type&, bias_type&>;
```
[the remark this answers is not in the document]
arcondello: At least in this PR it is `Bias, Index` on all of the template constructors, since the `Bias` is the required one and the `Index` is optional.

dimod/include/dimod/quadratic_model.h
arcondello: I don't think we could mark `const_neighborhood_iterator cbegin() const {}` in that case, since both would need to accept a non-const pointer.

I do totally hate the redundancy but as far as I tell you have to duplicate... Yay c++.

Issue #757, opened 2020-11-19
Improve performance of BQM.update for cyBQMs


arcondello:
Right now `BQM.update` uses the views. In the case of the cyBQMs this could be greatly improved.

Issue #747, opened 2020-11-06
Typo in BQM repr


randomir:
```
>>> dimod.BQM.from_ising({}, {'ab': 1})
BinaryQuadraticModel({a: 0.0, b: 0.0}, {('a', 'b'): 1}, 0, 'SPIN')
                      ▲       ▲
                      ┗━━━━━━━┻━━━━━━━  should be quoted
```
Also, at least for the `AdjDictBQM`, linear biases are stored (and printed) as floats, and quadratic/offset preserve the input type.

arcondello:
Good catch, comes from https://github.com/dwavesystems/dimod/blob/4941d2301aca0dc2c61b586605d0d3c69e2a7320/dimod/core/bqm.py#L59

Will fix.

mcfarljm:
Possible fix:
```python
            conv = str if isinstance(key, tuple) else repr
            stream.write('{}: {!s}'.format(conv(key), value))
```
Gives:
```
In [2]: dimod.BQM.from_ising({}, {'ab':1})
Out[2]: BinaryQuadraticModel({'a': 0.0, 'b': 0.0}, {('a', 'b'): 1}, 0, 'SPIN')
```

Also there is a line in the doc for `flip_variable` that has this issue as well:
https://github.com/dwavesystems/dimod/blob/dc98404439114f075aa7c0ff979ad30079eab2e4/dimod/core/bqm.py#L548-L552

arcondello:
I think just `stream.write('{!r}: {!r}'.format(key, value))` is the most explicit fix

randomir:
Yes, `repr`/`{!r}` is the correct solution, IMO, since the whole output is `repr(bqm)` in spirit.

Another manifestation of this bug:
```
>>> dimod.BQM.from_ising({0: 1, '0': 1}, {})                                                                                                                                                                  
BinaryQuadraticModel({0: 1, 0: 1}, {}, 0, 'SPIN')
                      ▲     ▲
                      ┗━━━━━┻━━━━━━━  indiscernible
```

Issue #699, opened 2020-09-09
Support multiple data types in BQMs


arcondello:
Right now all BQMs are float64, it would be good to support integer and float32 dtypes.

PR #828, opened 2021-05-26
Reimplement BQM with new C++ code

arcondello:
Closes https://github.com/dwavesystems/dimod/issues/254
Closes https://github.com/dwavesystems/dimod/issues/397 since all BQMs are ordered (also required python 3.6+)
Closes https://github.com/dwavesystems/dimod/issues/489
Closes https://github.com/dwavesystems/dimod/issues/528
Closes https://github.com/dwavesystems/dimod/issues/580 as no longer relevant
Closes https://github.com/dwavesystems/dimod/pull/690 as redundant
Closes https://github.com/dwavesystems/dimod/issues/699
Closes https://github.com/dwavesystems/dimod/issues/747
Closes https://github.com/dwavesystems/dimod/issues/757
Closes https://github.com/dwavesystems/dimod/pull/794 as redundant
Includes commits from and therefore closes https://github.com/dwavesystems/dimod/pull/824


arcondello:
There is still much to do including:
* Fix documentation
* Remove all the old BQM boilerplate
* Update the rest of dimod to default to the new BQM
But I think that makes sense to do in followup PRs.

dimod/binary/pybqm.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def add_variable(self, v: Optional[Variable] = None,
+                     bias: Any = 0) -> Variable:
+        if v is None:
+            # match the behaviour of dimod.Variables
+            adj = self._adj
+
+            v = len(adj)
+
+            if v in adj:
```
amahmudwave, line 99: Given you are checking all the elements of adj anyways, you could just do v = max(adj) + 1;

dimod/binary/pybqm.py
```diff
@@ -0,0 +1,324 @@
...
+
+        energies = samples.dot(ldata)
+        energies += (samples[:, irow]*samples[:, icol]).dot(qdata)
+        energies += ldata.dtype.type(self.offset)
+
+        return np.asarray(energies, dtype=dtype)
+
+    def get_linear(self, v: Variable) -> Any:
+        try:
+            return self._adj[v][v]
```
amahmudwave, line 184: Separating out the linear from _adj has its own advantages during simulated annealing. But I guess this bqm is not going to be used for simulated annealing.

dimod/binary/pybqm.py
```diff
@@ -0,0 +1,324 @@
...
+    def add_quadratic(self, u: Variable, v: Variable, bias: Any):
+        if u == v:
+            raise ValueError(f"{u!r} cannot have an interaction with itself")
+
+        self.add_variable(u)
+        self.add_variable(v)
+
+        zero = type(bias)(0)
+
+        self._adj[u][v] = self._adj[v][u] = bias + self._adj[v].get(u, zero)
```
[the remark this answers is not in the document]
arcondello: it might not exist yet - so we use `get` to handle that case.

dimod/binary/pybqm.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def add_linear_equality_constraint(self, *args, **kwargs):
+        raise NotImplementedError  # defer to caller
+
+    def add_linear_from_array(self, linear: ArrayLike):
+        for v, bias in enumerate(np.asarray(linear)):
+            self.add_linear(v, bias)
+
+    def add_quadratic(self, u: Variable, v: Variable, bias: Any):
+        if u == v:
```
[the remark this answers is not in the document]
arcondello: That's deliberate. We only support it in the dense case because using matrices to represent QUBOs is pretty common.

dimod/binary/pybqm.py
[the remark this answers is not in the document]
amahmudwave: I guessed that but you are using that element on the left side of the equation, does using [] operator create one ?

dimod/binary/pybqm.py
arcondello: yeah, in this case this is just meant to be a simple python BQM, not really useful for solving on.

dimod/binary/pybqm.py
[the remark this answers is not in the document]
arcondello: On the LHS it is being created or overwritten.

dimod/binary/pybqm.py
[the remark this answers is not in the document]
arcondello: Turns out there was a bug here, I must have misunderstood you. Handle the spin/binary distinction correctly now.

dimod/binary/pybqm.py
arcondello: yes, though we want to "fill in" missing values so try to keep it as close to index-labelled as possible.

dimod/binary/pybqm.py
```diff
@@ -84,10 +84,22 @@ def add_quadratic_from_dense(self, quadratic: ArrayLike):
         for u in range(num_variables):
             for v in range(num_variables):
                 if u == v:
-                    self.add_linear(u, quadratic[u, v])
+                    continue
```
amahmudwave, line 87: The modified code looks okay. But now the code under this elif condition caught my eye. You add a quadratic only if the quadratic exists ? What about if it does not exist. Secondly what if it exists and the value is positive 0, will the condition still be exited ?

arcondello: In this case we only add non-zero biases from the off diagonal, which makes sense since our representation is sparse.

arcondello: E.g.
```
In [5]: from dimod.binary.pybqm import pyBQM

In [6]: bqm = pyBQM('SPIN')

In [7]: bqm.add_quadratic_from_dense([[0, 0, 1], [0, 0, 0], [0, 0, 0]])

In [10]: list(bqm.iter_quadratic())
Out[10]: [(0, 2, 1)]
```
in this case only one interaction is added.

amahmudwave: Oh got it thanks.

amahmudwave: Just curious does python for an if condition consider if all the bits are zero or not , or it actually tries to find out if the value is zero or not. That is what if someone passses [-0 , -0 , -0] will it be added ?

dimod/binary/cybqm/cybqm_float64.pxd
```diff
@@ -0,0 +1,28 @@
...
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+cimport numpy as np
+
+ctypedef np.float64_t bias_type
+ctypedef np.int64_t index_type
```
amahmudwave, line 21: Why do we need the type of index to be of 64 bit ? There will be quite a bit of wastage of memory.

dimod/binary/cybqm/cybqm_float64.pyx
```diff
@@ -0,0 +1,25 @@
...
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+include "cybqm_template.pyx.pxi"
+
+import numpy as np
+
+BIAS_DTYPE = np.float64
+INDEX_DTYPE = np.int64
```
amahmudwave, line 20: Same comment as before.

dimod/binary/pybqm.py
```diff
@@ -0,0 +1,336 @@
...
+        while n < self.num_variables():
+            self.remove_variable()
+
+    def set_linear(self, v: Variable, bias: Any):
+        self._adj.setdefault(v, dict())[v] = bias
+
+    def set_quadratic(self, u: Variable, v: Variable, bias: Any):
+        if u == v:
+            raise ValueError(f"{u!r} cannot have an interaction with itself")
+        self.add_variable(u)
```
amahmudwave, line 328: What if the variables u and v already exists ?

dimod/binary/pybqm.py
```diff
@@ -0,0 +1,336 @@
...
+
+                # remove the old adj for old
+                del adj[old]
+
+    def relabel_variables_as_integers(self) -> Mapping[int, Variable]:
+        mapping = dict((v, i) for i, v in enumerate(self.variables) if i != v)
+        self.relabel_variables(mapping)
+        return dict((i, v) for v, i in mapping.items())
+
+    def remove_variable(self, v: Optional[Variable] = None) -> Variable:
```
amahmudwave, line 298: The linear / quadratic of a variable got absorbed to the offset in case of spin based bqm. Do we want to account for that when removing a variable ? Like remove the contribution it made to the offset ?

dimod/cyutilities.pyx
```diff
@@ -0,0 +1,110 @@
...
+    # lower-triangular
+    cdef Py_ssize_t i
+    for i in range(num_interactions):
+        if row[i] > col[i]:
+            row[i], col[i] = col[i], row[i]
+
+    quicksort_coo(row, col, data, 0, num_interactions - 1)
+
+
+cdef void quicksort_coo(Integer[:] row, Integer[:] col, cython.floating[:] data,
```
amahmudwave, line 74: I do not know how important performance is, but it would be much better to use a non recursive function. The zip sort was not recursive. Although the performance boost will be less than 30%

dimod/binary/pybqm.py
arcondello: All of them evaluate False
```
In [11]: bool(0)
Out[11]: False

In [12]: bool(0.)
Out[12]: False

In [13]: bool(-0.)
Out[13]: False

In [14]: bool(-0)
Out[14]: False
```
so none would get picked up, which I think it fine

dimod/binary/pybqm.py
arcondello: `add_variable(u)` does nothing when it already exists.

dimod/binary/pybqm.py
arcondello: That's what `fix_variable` does, `remove_variable` is just a raw removal.

dimod/cyutilities.pyx
arcondello: Yeah, this one is old. I know we could write a better performance one but it's never really been a bottleneck.

tests/test_bqm.py
```diff
@@ -1636,15 +1571,33 @@ def test_exclusions(self, name, BQM):
...
+        for _, bias in bqm.quadratic.items():
+            self.assertIsInstance(bias, int)
+
+    def test_dtypes_ndarray_ints(self):
+        obj = np.asarray([[0, 1], [1, 2]], dtype=np.int32)
+
+        bqm = DictBQM(obj, 'BINARY')
+
+        for _, bias in bqm.quadratic.items():
+            self.assertIsInstance(bias, np.int32)
```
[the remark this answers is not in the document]
arcondello: default int is np.int64

dimod/binary/cybqm/cybqm_float64.pxd
arcondello: I got this from you! :laughing: I must have misunderstood some benchmarks you sent me. I thought it was odd that it would matter they two different vectors have items of the same length.

amahmudwave: Okay if they have same length data type the performance will be better lets say 20% better for getting the effective biases from the CQM , it is true. Getting efffective bias is all that matters for binary and more or less for discrete sampling. For integers, it is the sampling that dominates. Lets say in total we will have 15% better performance if bias_type and index_type have same width. But as I understand 30% storage is more important than 15% performance in our case. If it is not the case we can definitely keep them of the same width :) 

If the variables are all of discrete type this will take up more than 30% of the space as I explained earlier for discrete variables we have two extra vectors of index_type. This is your call here :) . If performance is the goal we should keep this.

arcondello: No, you're right, in this case it's memory that we care about. I will make the change (luckily it's super easy to do).

amahmudwave: FYI, when I gather the 4 double precision biases, I have to supply 4 long long int indices to the cpu , so if they are of type int32_t  I need to convert them to int64_t that conversion takes up that extra 20% time in the tiny loop.

amahmudwave: Okay this discussion was fruitful I had missed one instruction that can actually gather without the conversion I will try it out and in that case we may save space and still have some performance :) 

https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=gather&techs=AVX,AVX2&expand=2999

amahmudwave (approved):

dimod/binary/pybqm.py
```diff
@@ -0,0 +1,336 @@
...
+        try:
+            return self._adj[u][v]
+        except KeyError:
+            if default is None:
+                raise ValueError(
+                    f"{u!r} and {v!r} have no interaction") from None
+            return default
+
+    def is_linear(self) -> bool:
+        return not any(len(Nv) > 1 for Nv in self._adj.values())
```
randomir, line 214: Looks like a :bug:.

Consider `dimod.BQM.from_ising({}, {'ab': 1, 'cd': 1})`. Has quadratic biases, but this function would return true.

dimod/serialization/fileview.py
```diff
@@ -207,345 +125,10 @@ def loads_data(self, data):
...
-        elif whence == SEEK_QUADRATIC:
-            self.pos = self.quadratic_start + offset
-        else:
-            raise ValueError("unknown value for 'whence'")
-        return self.pos
-
-    def seekable(self):
-        return True
+def FileView(bqm, version=(1, 0), ignore_labels=False):
+    warnings.warn("FileView is deprecated, please use `bqm.to_file` instead",
```
randomir, line 129: :tada:

randomir (approved):

dimod/binary/pybqm.py
arcondello: In adjdict we *always* keep the linear in the same dict, that's why we use `> 1`. I did add an extra test just to be sure :smile:

PR #832, opened 2021-06-02
check bqm feature first pass

### LINKED ISSUE #830: StructuredSampler.check_bqm or similar
Something like
```
def check_bqm(self, bqm):
    return all(v in self.adjacency for v in bqm.variables) and all(u in self.adjacency[v] for u, v in bqm.quadratic)
```
See https://github.com/dwavesystems/dwave-system/issues/401

hbarovertwo:
Feels a little clunky, reviews appreciated :)

Closes #830

### DIFF
--- a/dimod/core/structured.py
+++ b/dimod/core/structured.py
@@ -74,6 +74,7 @@ Examples:
 import abc
 
 from collections import namedtuple
+from dimod.exceptions import BinaryQuadraticModelStructureError
 
 __all__ = ['Structured']
 
@@ -144,3 +145,25 @@ class Structured(abc.ABC):
         G.add_nodes_from(self.nodelist)
 
         return G
+
+    def check_bqm_structure(self, bqm: "BinaryQuadraticModel"):
+        """Validate that problem defined by :class:`dimod.BinaryQuadraticModel`
+        matches the graph provided by the sampler.
+
+        Args:
+            bqm: :class:`dimod.BinaryQuadraticModel` object to validate.
+
+        Returns:
+            True if BQM structure matches that of sampler.
+
+        Raises:
+            :exception:`dimod.exceptions.BinaryQuadraticModelStructureError`
+            if structure doesn't match.
+
+        """
+        if all(variable in self.adjacency for variable in bqm.variables) and \
+           all(neighbor in self.adjacency[vertex] for neighbor, vertex in bqm.quadratic):
+            return True
+        raise BinaryQuadraticModelStructureError(
+            "BQM structure does not match graph provided by sampler."
+        )
--- a/dimod/core/structured.py
+++ b/dimod/core/structured.py
@@ -146,7 +146,7 @@ class Structured(abc.ABC):
 
         return G
 
-    def check_bqm_structure(self, bqm: "BinaryQuadraticModel"):
+    def valid_bqm_structure(self, bqm: "BinaryQuadraticModel"):
         """Validate that problem defined by :class:`dimod.BinaryQuadraticModel`
         matches the graph provided by the sampler.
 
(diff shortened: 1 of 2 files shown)


dimod/core/structured.py
```diff
@@ -144,3 +145,25 @@ def to_networkx_graph(self):
...
+            True if BQM structure matches that of sampler.
+
+        Raises:
+            :exception:`dimod.exceptions.BinaryQuadraticModelStructureError`
+            if structure doesn't match.
+
+        """
+        if all(variable in self.adjacency for variable in bqm.variables) and \
+           all(neighbor in self.adjacency[vertex] for neighbor, vertex in bqm.quadratic):
+            return True
```
arcondello, line 166: I would either return nothing and raise when it doesn't match, or just return True/False

hbarovertwo: That makes sense. I think returning True/False is ok in this case, as the decorator on sampling methods will raise this exception.

arcondello: In that case maybe the method should be called `.valid_bqm_structure(bqm)` or something like that.

dimod/core/structured.py
```diff
@@ -144,3 +145,19 @@ def to_networkx_graph(self):
...
+            bqm: :class:`dimod.BinaryQuadraticModel` object to validate.
+
+        Returns:
+            Boolean indicating validity of BQM structure compared to sampler graph.
+
+        """
+        if all(variable in self.adjacency for variable in bqm.variables) and \
+           all(neighbor in self.adjacency[vertex] for neighbor, vertex in bqm.quadratic):
+            return True
+        return False
```
arcondello, line 163: or even just `return all(....` :smile:

dimod/core/structured.py
JoelPasvolsky: Would "graph" be clearer than "structure" to most users?

dimod/core/structured.py
hbarovertwo: Good point, I actually had it like that prior to pushing 🤕

dimod/core/structured.py
hbarovertwo: Sounds more clear. For a second leaned towards `valid_bqm_graph_structure`, or we can go with `valid_bqm_graph`.

arcondello: I am fine with any of these. Shorter is nicer in general.

dimod/core/structured.py
```diff
@@ -144,3 +145,17 @@ def to_networkx_graph(self):
         G.add_nodes_from(self.nodelist)
 
         return G
+
+    def valid_bqm_graph(self, bqm: "BinaryQuadraticModel"):
```
randomir, line 149: Perhaps a more general approach?
```
def has_subgraph(self, linear, quadratic):
```

I'm fine with either, though.

dimod/core/structured.py
```diff
@@ -74,6 +74,7 @@ def sample(self, bqm):
 import abc
 
 from collections import namedtuple
+from dimod.exceptions import BinaryQuadraticModelStructureError
```
randomir, line 77: Not used.

dimod/core/structured.py
randomir: The only reason I'm suggesting this is because we don't mention/specialize for BQM here so far.

dimod/core/structured.py
hbarovertwo: Thanks

dimod/core/structured.py
hbarovertwo: At least the docstring above mentions "A structured sampler can only sample from binary quadratic models with a specific graph."

arcondello: I do like the generality, but one reason not to is that we can improve the performance if we're using a BQM specifically. I don't mind it being specific to BQMs because at the moment `Sampler` is only defined for BQMs.

randomir: Ideally, I think, we would add `.samples_from(bqm)` to a `StructuredSampler` class (which we don't have :laughing: )

arcondello (approved):

Issue #833, opened 2021-06-07
QuadraticModel::scale in C++ code


arcondello:
something like
```
template<class Bias, class Index>
void QuadraticModel<Bias, Index>::scale(Bias scalar)
```
First step to address https://github.com/dwavesystems/dimod/blob/be747fa4be2c898f7460f77573f82d35f634df57/dimod/binary/binary_quadratic_model.py#L1290

Issue #624, opened 2020-02-28
Fix default typing in AdjDictBQM


arcondello:
Failing example:
```
>>> from dimod import AdjDictBQM
>>> from fractions import Fraction
>>> AdjDictBQM({'a': Fraction(1, 3)}, {'ab': Fraction(2, 7)}, 'SPIN')
AdjDictBQM({a: 0.3333333333333333, b: 0.0}, {('a', 'b'): 2/7}, 0.0, 'SPIN')
```
Would expect to see
```
>>> AdjDictBQM({'a': Fraction(1, 3)}, {'ab': Fraction(2, 7)}, 'SPIN')
AdjDictBQM({a: 1/3, b: 0.0}, {('a', 'b'): 2/7}, 0.0, 'SPIN')
```

This happens because the quadratic input is parsed first, which assigns a `0.0` linear bias to `'a'` and `'b'`. When the linear bias is later added
```
>>> 0.0 + Fraction(1, 3)
0.3333333333333333
```

PR #834, opened 2021-06-07
Add symbolic manipulation of BQM objects

arcondello:
Add very simple operator overloading to the BQM object.

#### Alternatives Considered
Right now `__add__` etc create a new BQM object, rather than something like
```
Add = namedtuple('Add', ['lhs', 'rhs'])

class BQM:
    ...
    def __add__(self, other):
        return Add(self, other)
```

Returning a relation object like above has performance advantages (at the cost of memory), but it's a lot less flexible and a lot less self-contained. For instance, right now
```
ss = ExactSolver().sample(bqm0 + bqm1)
```
is defined without needing some sort of resolution step.

One way around this would be to make comparisons full-fledged BQM data classes (e.g. [pyBQM](https://github.com/dwavesystems/dimod/blob/be747fa4be2c898f7460f77573f82d35f634df57/dimod/binary/pybqm.py#L35) or [VartypeView](https://github.com/dwavesystems/dimod/blob/be747fa4be2c898f7460f77573f82d35f634df57/dimod/binary/vartypeview.py#L56)) and then do something like
```
class BQM:
    ...
    def __add__(self, other):
        bqm = self.__new__(type(self))
        bqm.data = AddBQM(self, other)
```

but for now I think it makes sense to keep things simple.

#### Additional Context

Performance of the multiplication operator would be improved by #833 
Closes #790 as redundant.

### DIFF
--- a/dimod/binary/binary_quadratic_model.py
+++ b/dimod/binary/binary_quadratic_model.py
@@ -38,6 +38,7 @@ from dimod.binary.pybqm import pyBQM
 from dimod.binary.vartypeview import VartypeView
 from dimod.decorators import forwarding_method
 from dimod.serialization.fileview import SpooledTemporaryFile, _BytesIO, VariablesSection
+from dimod.sym import symbolic
 from dimod.typing import Bias, Variable
 from dimod.variables import Variables, iter_deserialize_variables
 from dimod.vartypes import as_vartype, Vartype
@@ -48,6 +49,7 @@ __all__ = ['BinaryQuadraticModel',
            'Float32BQM',
            'Float64BQM',
            'as_bqm',
+           'Spin', 'Binary',
            ]
 
 BQM_MAGIC_PREFIX = b'DIMODBQM'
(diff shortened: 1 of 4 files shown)


dimod/binary/binary_quadratic_model.py
```diff
@@ -390,30 +392,161 @@ def __deepcopy__(self, memo):
...
+    def __add__(self, other: Union['BinaryQuadraticModel', Bias]):
+        # in python 3.8+ we could do this is functools.singledispatchmethod
+        if isinstance(other, BinaryQuadraticModel):
+            if other.num_variables and other.vartype != self.vartype:
+                # future: return QuadraticModel
+                raise TypeError("cannot add BQMs with different vartypes")
+            new = self.copy()
+            new.update(other)
+            return new
+        if isinstance(other, Number):
```
randomir, line 414: Doesn't `DictBQM` support `object` for bias `dtype`?

Currently `NotImplemented` is raised, even if that object implements `+`.

In other places as well.

dimod/sym.py
```diff
@@ -0,0 +1,60 @@
...
+        except AttributeError:
+            return False
+
+
+class Ge(Comparison, sense='>='):
+    pass
+
+
+class Le(Comparison, sense='<='):
+    pass
```
randomir, line 60: You could compare lhs.offset with numeric rhs. Probably doesn't make much sense, but it seems consistent.

randomir (approved):
LGTM!

dimod/sym.py
arcondello: IMO, makes sense to keep the implementation simpler. Though if we get a feature request for that I am happy to support it.

dimod/binary/binary_quadratic_model.py
arcondello: Hmm, you're right. And I just checked that `isinstance(sympy.Symbol('a'), Number) == False`... I'll take a look.

arcondello: Ok, I played with a couple different variations of trying to handle this case and this way lies madness. I think (for now) only allowing symbolic manipulation for numeric types strikes a good balance between nice sugar and not having to deal with a million edge cases.

Users can always make use of the `bqm.offset += val` and `bqm.scale(val)` directly, which will handle other types reasonably well (at least when combined with #624 )

randomir: Yeah, among other things, a bias can be just another BQM :laughing:

PR #837, opened 2021-06-08
add_linear_inequality_constraint to dimod

rahmani15:
This PR adds `add_linear_inequality_constraint` method for both DQM and BQM objects. For BQM object it adds Log2 (slack_range) number of binary slack variables to the constraint. For DQM Object there are two options: 1) log2 adds log2(slack_range) number of dqm variables each with two cases 2)  log10 method adds log10 dqm variables each with up to 10 cases. 
Note that this method only handles integral coefficient correctly. When there are fractional coefficients for variables it is assumed that user will multiply both sides of equations by factor of 10s to remove fractional coefficients.


dimod/discrete/discrete_quadratic_model.py
```diff
@@ -203,6 +203,94 @@ def add_linear_equality_constraint(self, terms: LinearTriplets,
         self._cydqm.add_linear_equality_constraint(
             index_terms, lagrange_multiplier, constant)
 
+    def add_linear_inequality_constraint(self, terms: LinearTriplets,
```
hbarovertwo, line 206: Would it be desirable to implement this in the Cython DQM object, like `add_linear_equality_constraint`?

rahmani15: I am not sure, the `add_linear_equality_constraint` which is already implemented in Cython does the heavy lifting in this function.

dimod/binary/binary_quadratic_model.py
```diff
@@ -624,6 +624,84 @@ def add_linear_equality_constraint(
...
+        :math:`\\sum_{i} a_{i} x_{i} + C <= 0`
+        to the binary quadratic model as a quadratic objective. Only integral coefficient are handled correctly.
+        Try multiplying both sides of equations by factor of 10s to remove fractional coefficients.
+
+        Args:
+            terms (iterable/iterator):
+                An iterable of 2-tuples, (variable, bias).
+                Each tuple is evaluated to the term (bias * variable).
+                All terms in the list are summed.
+            lagrange_multiplier:
```
JoelPasvolsky, line 647: We often use this name for input parameters like these but @mcfarljm has pointed out that in many cases that's inaccurate and should be something like ``penalty_weight`` or similar.

dimod/binary/binary_quadratic_model.py
```diff
@@ -624,6 +624,84 @@ def add_linear_equality_constraint(
...
+                The coefficient or the penalty strength. This value is
```
JoelPasvolsky, line 648: "The coefficient or the penalty strength" --> I think "coefficient " here might be confusing. Perhaps "A weight or ..."

"and added to the binary quadratic model. It doesn't appear explicitly in the equation above.

dimod/binary/binary_quadratic_model.py
```diff
@@ -624,6 +624,84 @@ def add_linear_equality_constraint(
...
+                multiplied by the entire constraint objective and added to the
+                bqm (it doesn't appear explicitly in the equation above).
+            constant:
+                The constant value of the constraint, C, in the equation above
+            label:
+                Label assigned to the new slack variables.
```
JoelPasvolsky, line 654: Looks like this should be something like, "Prefix used to label the slack variables used to create the new objective."

dimod/binary/binary_quadratic_model.py
```diff
@@ -624,6 +624,84 @@ def add_linear_equality_constraint(
...
+            slack_method:
+                The method for adding slack. Only log2 method currently supported for BQM object. log2 adds
```
JoelPasvolsky, line 656: "The method for adding slack variables. Supported methods are: 

 * log2: Adds log2(``slack_range``) binary variables to the constraint.

"

dimod/binary/binary_quadratic_model.py
```diff
@@ -624,6 +624,84 @@ def add_linear_equality_constraint(
...
+        Args:
+            terms (iterable/iterator):
+                An iterable of 2-tuples, (variable, bias).
+                Each tuple is evaluated to the term (bias * variable).
+                All terms in the list are summed.
+            lagrange_multiplier:
+                The coefficient or the penalty strength. This value is
+                multiplied by the entire constraint objective and added to the
+                bqm (it doesn't appear explicitly in the equation above).
+            constant:
```
JoelPasvolsky, line 651: constant (int):

dimod/binary/binary_quadratic_model.py
```diff
@@ -624,6 +624,84 @@ def add_linear_equality_constraint(
...
+                The constant value of the constraint, C, in the equation above
+            label:
+                Label assigned to the new slack variables.
+            slack_method:
+                The method for adding slack. Only log2 method currently supported for BQM object. log2 adds
+                 log2(slack_range) number of binary variables to the constraint.
+            slack_range:
+                could be used to limit the range of slack variables.
+            discontinuous_slack:
+                If true it allows the lhs of the constraint be zero when slack_range is provided.
```
JoelPasvolsky, line 661: If true, the lefthand side of the constraint can be zero when ``slack_range`` is provided.

JoelPasvolsky (commented):
So far just looked at the docstring for add_linear_inequality_constraint

dimod/binary/binary_quadratic_model.py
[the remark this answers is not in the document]
arcondello: This is now covered by the type annotations

dimod/binary/binary_quadratic_model.py
```diff
@@ -624,6 +624,84 @@ def add_linear_equality_constraint(
...
+            label: str,
+            slack_method: str = "log2",
+            slack_range: int = None,
+            discontinuous_slack: bool = False
+    ):
+        """Add a linear inequality constraint as a quadratic objective.
+
+        Adds a linear inequality constraint of the form
+        :math:`\\sum_{i} a_{i} x_{i} + C <= 0`
+        to the binary quadratic model as a quadratic objective. Only integral coefficient are handled correctly.
```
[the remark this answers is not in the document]
rahmani15: Thanks, this is fixed.

dimod/binary/binary_quadratic_model.py
```diff
@@ -624,6 +624,84 @@ def add_linear_equality_constraint(
...
+        Try multiplying both sides of equations by factor of 10s to remove fractional coefficients.
+
+        Args:
+            terms (iterable/iterator):
+                An iterable of 2-tuples, (variable, bias).
+                Each tuple is evaluated to the term (bias * variable).
```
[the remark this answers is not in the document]
rahmani15: Should we also fix this in the  `add_linear_equality_constraint`? I'd like to stay consistent with the docstring we have for that method.

dimod/binary/binary_quadratic_model.py
rahmani15: Should we also fix this in the  `add_linear_equality_constraint`? I'd like to stay consistent with the docstring we have for that method.

dimod/binary/binary_quadratic_model.py
rahmani15: Should we also fix this in the  `add_linear_equality_constraint`? I'd like to stay consistent with the docstring we have for that method.

dimod/binary/binary_quadratic_model.py
```diff
@@ -624,6 +624,84 @@ def add_linear_equality_constraint(
...
+                All terms in the list are summed.
+            lagrange_multiplier:
+                The coefficient or the penalty strength. This value is
+                multiplied by the entire constraint objective and added to the
+                bqm (it doesn't appear explicitly in the equation above).
+            constant:
+                The constant value of the constraint, C, in the equation above
```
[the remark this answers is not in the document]
rahmani15: Thanks! fixed this

dimod/binary/binary_quadratic_model.py
rahmani15: Do we need to define type here as well? the function argument says its type is Bias @arcondello

dimod/binary/binary_quadratic_model.py
```diff
@@ -624,6 +624,84 @@ def add_linear_equality_constraint(
...
+            label:
+                Label assigned to the new slack variables.
+            slack_method:
+                The method for adding slack. Only log2 method currently supported for BQM object. log2 adds
+                 log2(slack_range) number of binary variables to the constraint.
+            slack_range:
+                could be used to limit the range of slack variables.
```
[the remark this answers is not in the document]
rahmani15: fixed

dimod/binary/binary_quadratic_model.py
rahmani15: fixed

dimod/binary/binary_quadratic_model.py
```diff
@@ -758,6 +758,93 @@ def add_linear_equality_constraint(
                     u, v, 2 * lagrange_multiplier * ubias * vbias)
         self.offset += lagrange_multiplier * constant * constant
 
+    def add_linear_inequality_constraint(
+            self, terms: Iterable[Tuple[Variable, Bias]],
+            lagrange_multiplier: Bias,
+            label: str,
```
arcondello, line 764: this label is not kept anywhere so I don't think it should be an input.

dimod/binary/binary_quadratic_model.py
```diff
@@ -758,6 +758,93 @@ def add_linear_equality_constraint(
...
+            constant: int = 0,
+            lb: int = - np.Infinity,
```
arcondello, line 766: `np.Infinity` is actually a float, not an int.

dimod/binary/binary_quadratic_model.py
```diff
@@ -758,6 +758,93 @@ def add_linear_equality_constraint(
...
+            ub: int = 0,
+            slack_method: str = "log2",
```
arcondello, line 768: if we only support one `slack_method`, why make it an argument at all? We can always add it later if needed.

dimod/binary/binary_quadratic_model.py
```diff
@@ -758,6 +758,93 @@ def add_linear_equality_constraint(
...
+            ub:
+                upper bound for the constraint
+            slack_method:
+                "The method for adding slack variables. Supported methods are:
+                - log2: Adds log2(slack_range) number of binary variables to the constraint.
+            cross_zero
+                If True, includes zero as the domain of constraint if not include already
+
+        """
+        assert (slack_method in ['log2'])
```
arcondello, line 804: As I said elsewhere, if we only support one method, then let's just not make it an argument until we support two.

Also, in python you shouldn't use asserts for input checking, it should be
```
if slack_method not in ['log2']:
    raise ValueError("unexpected slack_method")
```
or something like that.

dimod/binary/binary_quadratic_model.py
```diff
@@ -758,6 +758,93 @@ def add_linear_equality_constraint(
...
+            slack_method: str = "log2",
+            cross_zero: bool = False
+    ):
+        """Add a linear inequality constraint as a quadratic objective.
+
+        Adds a linear inequality constraint of the form:
+
+        math:'lb <= \sum_{i,k} a_{i,k} x_{i,k} + constant <= ub'
+        to the binary quadratic model as a quadratic objective.
+        For constraints with fractional coefficients, multiply both sides of the inequality by an
```
arcondello, line 777: If this is a requirement, then we need to enforce it in code and raise an exception or at the very least a warning. Users don't read docstrings :smile:

dimod/binary/binary_quadratic_model.py
```diff
@@ -758,6 +758,93 @@ def add_linear_equality_constraint(
...
+            lb:
+                lower bound for the constraint
+            ub:
+                upper bound for the constraint
+            slack_method:
+                "The method for adding slack variables. Supported methods are:
+                - log2: Adds log2(slack_range) number of binary variables to the constraint.
+            cross_zero
+                If True, includes zero as the domain of constraint if not include already
+
```
arcondello, line 802: Need to document the return value, and add it to the function signature as a type annotation.

dimod/binary/binary_quadratic_model.py
```diff
@@ -758,6 +758,93 @@ def add_linear_equality_constraint(
...
+        ub_c = min(terms_upper_bound, ub - constant)
+        lb_c = max(terms_lower_bound, lb - constant)
+        slack_terms = []
+
+        if terms_upper_bound <= ub_c and terms_lower_bound >= lb_c:
+            print(f'Did not add constraint {label}. This constraint is feasible with any value for state variables.')
+            return slack_terms
+
+        if ub_c <= lb_c:
+            print(f'Did not add constraint {label}. This constraint is infeasible with any value for state variables.')
```
arcondello, line 817: if the inputs are ill-defined, then we should raise a `ValueError`, not return.
```
      if ub_c <= lb_c:
            raise ValueError('the given constraint is infeasible with any value for state variables.')
```

dimod/binary/binary_quadratic_model.py
```diff
@@ -758,6 +758,93 @@ def add_linear_equality_constraint(
...
+            return slack_terms
+
+        slack_upper_bound = int(ub_c - lb_c)
+        if slack_upper_bound == 0:
+            print(f'add constraint {label} as equality constraint.')
```
arcondello, line 822: Either raise an error telling the user to use `add_linear_equality_constraint` or just accept it quietly without the print statement.

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -217,6 +214,104 @@ def add_linear_equality_constraint(self, terms: LinearTriplets,
...
+                All terms in the list are summed.
+            lagrange_multiplier:
+                The coefficient or the penalty strength
+            label:
+                Prefix used to label the slack variables used to create the new objective.
+            constant:
+                The constant value of the constraint.
+            lb:
+                lower bound for the constraint
+            ub:
```
arcondello, line 248: missing description

tests/requirements.txt
```diff
@@ -6,4 +6,4 @@ parameterized==0.7.4
 pandas==0.25.3; python_version < '3.9'
 networkx==2.5
 simplejson==3.16.0
-pymongo==3.7.1; python_version < '3.9'
+pymongo==3.7.1; python_version < '3.9'
```
arcondello, line 9: not sure why this file was touched?

dimod/binary/binary_quadratic_model.py
arcondello: Changing it in ` add_linear_equality_constraint` would be a backwards compatibility break, so we should keep them consistent.

dimod/binary/binary_quadratic_model.py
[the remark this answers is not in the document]
arcondello: Changing the wording in both places is fine.

dimod/binary/binary_quadratic_model.py
arcondello: Changing the wording in both places is fine.

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -217,6 +214,104 @@ def add_linear_equality_constraint(self, terms: LinearTriplets,
...
+            slack_method:
+                "The method for adding slack variables. Supported methods are:
+                - log2: Adds log2(slack_range) number of dqm variables each with two cases to the constraint.
+                - log10: Adds log10 dqm variables each with up to 10 cases.
+
+                upper bound for the constraint
+            cross_zero
+                If True, includes zero as the domain of constraint if not include already
+       """
+        assert (slack_method in ['log2', 'log10'])
```
arcondello, line 258: Rather than using an assert, we should raise a proper error. E.g.
```
if slack_method not in ['log2', 'log10']:
    raise ValueError(r"expected slack_method to be 'log2' or 'log10', but got {slack_method!r}")
```

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -217,6 +214,104 @@ def add_linear_equality_constraint(self, terms: LinearTriplets,
...
+
+        terms_upper_bound = sum(v for _, _, v in terms if v > 0)
```
arcondello, line 260: Since we are passing through the terms more than once, we actually require terms to be a `Collection`. Right now I could pass in a generator and this would fail since the generator would be exhausted by this line.

Example:
```
>>> terms_list = [('a', 'b', 1), ('b', 'c', 2), ('c', 'd', -1)]
>>> print(sum(v for _, _, v in terms_list if v > 0))
3
>>> print(sum(v for _, _, v in terms if v < 0))
-1
```
works fine, but 
```
>>> terms_generator = (t for t in [('a', 'b', 1), ('b', 'c', 2), ('c', 'd', -1)])
>>> print(sum(v for _, _, v in terms_generator if v > 0))
3
>>> print(sum(v for _, _, v in terms_generator if v < 0))
0
```
both `terms_list` and `terms_generator` are `Iterable`.

An alternative would be something like

```
if isinstance(terms, Iterator):
    terms = list(terms)

File: dimod/discrete/discrete_quadratic_model.py [hunk 24]
```diff
@@ -217,6 +214,104 @@ def add_linear_equality_constraint(self, terms: LinearTriplets,
...
+        terms_lower_bound = sum(v for _, _, v in terms if v < 0)
+        ub_c = min(terms_upper_bound, ub - constant)
+        lb_c = max(terms_lower_bound, lb - constant)
+        slack_terms = []
+
+        if terms_upper_bound <= ub_c and terms_lower_bound >= lb_c:
+            print(f'Did not add constraint {label}. This constraint is feasible with any value for state variables.')
```
arcondello, line 267: See above, raise error.

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -217,6 +214,104 @@ def add_linear_equality_constraint(self, terms: LinearTriplets,
...
+            return slack_terms
+
+        if ub_c <= lb_c:
+            print(f'Did not add constraint {label}. This constraint is infeasible with any value for state variables.')
```
arcondello, line 271: See above, raise error

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -217,6 +214,104 @@ def add_linear_equality_constraint(self, terms: LinearTriplets,
...
+            return slack_terms
+
+        slack_upper_bound = int(ub_c - lb_c)
+        if slack_upper_bound == 0:
+            print(f'add constraint {label} as equality constraint.')
```
arcondello, line 276: See above, let this pass silently or raise error.

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -396,7 +491,7 @@ def from_numpy_vectors(cls, case_starts, linear_biases, quadratic,
             if len(labels) != obj._cydqm.num_variables():
                 raise ValueError(
                     "labels does not match the length of the DQM"
-                    )
+                )
```
arcondello, line 494: It's fine for this PR but in the future it would be good not to touch unrelated lines of code. It confuses the git history and makes merging harder.

arcondello (changes requested):

dimod/binary/binary_quadratic_model.py
randomir: ...and the reason for not using `assert` is because it gets eliminated from optimized bytecode (both with `-O` and `-OO`).

dimod/binary/binary_quadratic_model.py
rahmani15: The label goes to slack_terms and slack_terms are returned.

dimod/binary/binary_quadratic_model.py
rahmani15: I'll remove slack_method.

dimod/binary/binary_quadratic_model.py
arcondello: oops, missed that! thanks

dimod/binary/binary_quadratic_model.py
rahmani15: I'll add a warning. The formulation/method when there are fractional values are approximate and not necessarily invalid.

dimod/binary/binary_quadratic_model.py
rahmani15: I'll go with accept it quietly without print as user might not be able to define if a constraint binds.

tests/requirements.txt
rahmani15: I made sure to add/commit only changes I need. I don't know what happened that this file got changed. It is a also not obvious what has changed both of these look the same. I'll resolve the conflicts.

arcondello: You should be able to restore it with `git checkout main -- tests/requirements.txt`

My guess is that the difference is a newline.

arcondello:
Also, one more thing, can you add the new methods to the release notes? See https://github.com/dwavesystems/dimod#release-notes. Let me know if you have trouble or it's not clear.

Something like
```
Add `BinaryQuadraticModel.add_linear_inequality_constraint` and `DiscreteQuadraticModel.add_linear_inequality_constraint` methods.
```

dimod/discrete/discrete_quadratic_model.py
rahmani15: what is 'r' before and after the message. does that define the color? It is already red :)

arcondello: oh, typo, sorry!

arcondello: the end `r` is meaningful
```
>>> a = 'a'
>>> print(f"{a}")
a
>>> print(f"{a!r}")
'a'
```
the latter uses `__repr__` rather than `__str__`. See [SO](https://stackoverflow.com/questions/1436703/what-is-the-difference-between-str-and-repr).

dimod/discrete/discrete_quadratic_model.py
rahmani15: I think I should avoid using PyCharm automatic formatting. I'll roll back these changes.

dimod/binary/binary_quadratic_model.py
```diff
@@ -763,6 +764,120 @@ def add_linear_equality_constraint(
...
+
+        Returns:
+            slack_terms:  An iterable of 2-tuples for the new slack variables
+             (variable, int), with each tuple constituting a term in
+             :math:`\sum_{i} b_{i} slack {i},
+             with i being the length of the iterable and b being the coefficient
+             for slack variable.
+        """
+
+        if isinstance(terms, Iterator):
```
arcondello, line 816: Can we add a test for this case? Something like
```
bqm.add_linear_inequality_constraint(iter(terms), ...)
```
will do it.

dimod/binary/binary_quadratic_model.py
```diff
@@ -763,6 +764,120 @@ def add_linear_equality_constraint(
...
+            terms = list(terms)
+
+        all_terms = [constant, lb, ub] + [v for _, v in terms]
```
arcondello, line 819: I would replace this whole block with
```
if int(constant) != constant or int(lb) != lb or int(ub) != ub or any(int(bias) != bias for _, bias in terms):
    warnings.warn(...)
```

dimod/binary/binary_quadratic_model.py
```diff
@@ -763,6 +764,120 @@ def add_linear_equality_constraint(
...
+                    warnings.warn(
+                        "For constraints with fractional coefficients, "
+                        "multiply both sides of the inequality by an "
+                        "appropriate factor of ten to attain or "
+                        "approximate integer coefficients. ")
+                    break
+            else:
+                raise ValueError("unexpected input value")
+
+        terms_upper_bound = sum(v for _, v in terms if v > 0)
```
arcondello, line 834: I think we should be careful to use `bias` or `b` for biases and `v` for variables.

dimod/binary/binary_quadratic_model.py
rahmani15: smart!

dimod/binary/binary_quadratic_model.py
```diff
@@ -763,6 +764,120 @@ def add_linear_equality_constraint(
...
+        terms_lower_bound = sum(v for _, v in terms if v < 0)
+        ub_c = min(terms_upper_bound, ub - constant)
+        lb_c = max(terms_lower_bound, lb - constant)
+
+        if terms_upper_bound <= ub_c and terms_lower_bound >= lb_c:
+            warnings.warn(
+                f'Did not add constraint {label}.'
+                f' This constraint is feasible'
```
arcondello, line 842: Don't need the `f'...` when there isn't a `{}`.

arcondello (commented):
Getting close! Some of my comments apply to both.

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -217,6 +216,150 @@ def add_linear_equality_constraint(self, terms: LinearTriplets,
...
+            slack_terms:  A list of tuples of the type (variable, case, bias)
+                for the new slack variables.
+                Each tuple is evaluated to the term (bias * variable_case).
+                All terms in the list are summed.
+
+       """
+
+        if slack_method not in ['log2', 'log10']:
+            raise ValueError(
+                r"expected slack_method to be 'log2' or 'log10', "
```
arcondello, line 277: don't need the `r` on the first line, and I think you mean `f` on the second one.

dimod/binary/binary_quadratic_model.py
rahmani15: added this to one of the existing tests.

arcondello (approved):

Issue #744, opened 2020-10-28
add_offset for DQM


hay-k:
**Application**
In some applications the energy value of a sample has an important meaning to it. For instance I am working on a graph partitioning problem, and due to the structure of the cost function for this problem, the energy value of a sample (offsets included) is exactly the number of cut edges between the partitions. You get the energy of a sample and you know the number of cut edges right away and you don't need to traverse the graph one more time to count the cut edges. When formulating a problem in terms of a DQM, one usually throws away any constant offset which appear on the way, but in problems like this it is sometimes convenient to keep the constants. BQM has a function called `add_offset` which allows to do this.

**Proposed Solution**
Implement `add_offset` function for DQM as well. 

**Alternatives Considered**
I am currently calculating the offset in a separate variable, so that I can add it to the energy of the result after sampling.

arcondello:
I agree with adding an offset to DQM. However, I would just add a property so the syntax would be `dqm.offset += 1`rather than `dqm.add_offset(1)`

PR #838, opened 2021-06-09
Add offset to dqm

hbarovertwo:
Note: Publishing before it's ready to merge to get eyes on it.

Implement offset in DQM object via underlying `cyDiscreteQuadraticModel` object.
Update methods in backwards compatibility safe way.
Add new tests and update existing tests to reflect offset.

Closes https://github.com/dwavesystems/dimod/issues/744

### DIFF
--- a/dimod/discrete/cydiscrete_quadratic_model.pxd
+++ b/dimod/discrete/cydiscrete_quadratic_model.pxd
@@ -40,7 +40,7 @@ cdef class cyDiscreteQuadraticModel:
     cdef vector[CaseIndex] case_starts_  # len(adj_) + 1
     cdef vector[vector[VarIndex]] adj_
 
-    cdef Bias offset_
+    cdef public Bias offset
 
     cdef readonly object dtype
     cdef readonly object case_dtype
--- a/dimod/discrete/cydiscrete_quadratic_model.pxd
+++ b/dimod/discrete/cydiscrete_quadratic_model.pxd
@@ -40,6 +40,8 @@ cdef class cyDiscreteQuadraticModel:
     cdef vector[CaseIndex] case_starts_  # len(adj_) + 1
     cdef vector[vector[VarIndex]] adj_
 
+    cdef Bias offset_
+
     cdef readonly object dtype
     cdef readonly object case_dtype
 
--- a/dimod/discrete/cydiscrete_quadratic_model.pyx
+++ b/dimod/discrete/cydiscrete_quadratic_model.pyx
@@ -18,7 +18,6 @@ cimport cython
 
 from cython.operator cimport preincrement as inc, dereference as deref
 from cython.operator cimport preincrement as preinc  # for when it matters
-from libcpp cimport bool
 from libcpp.algorithm cimport lower_bound, sort, unique
 from libcpp.unordered_set cimport unordered_set
 
@@ -687,7 +686,7 @@ cdef class cyDiscreteQuadraticModel:
                 qi += 1
         
 
-    def to_numpy_vectors(self, bool return_offset=False):
+    def to_numpy_vectors(self, bint return_offset=False):
         
         cdef Py_ssize_t num_variables = self.num_variables()
         cdef Py_ssize_t num_cases = self.num_cases()
--- a/dimod/discrete/cydiscrete_quadratic_model.pyx
+++ b/dimod/discrete/cydiscrete_quadratic_model.pyx
@@ -74,7 +74,7 @@ cdef class cyDiscreteQuadraticModel:
     def add_linear_equality_constraint(self, object terms,
                                        Bias lagrange_multiplier, Bias constant):
         # adjust energy offset
-        self.offset_ += lagrange_multiplier * constant * constant
+        self.offset += lagrange_multiplier * constant * constant
 
         # resolve the terms from a python object into a C++ object
         cdef vector[LinearTerm] cppterms
@@ -223,7 +223,7 @@ cdef class cyDiscreteQuadraticModel:
         dqm.bqm_ = self.bqm_
         dqm.case_starts_ = self.case_starts_
         dqm.adj_ = self.adj_
-        dqm.offset_ = self.offset_
+        dqm.offset = self.offset
 
         return dqm
 
@@ -242,7 +242,7 @@ cdef class cyDiscreteQuadraticModel:
         cdef Py_ssize_t num_samples = samples.shape[0]
         cdef VarIndex num_variables = samples.shape[1]
 
-        cdef Bias[:] energies = np.full(num_samples, self.offset_, dtype=self.dtype)
+        cdef Bias[:] energies = np.full(num_samples, self.offset, dtype=self.dtype)
 
         cdef Py_ssize_t si, vi
         cdef CaseIndex cu, case_u, cv, case_v
@@ -382,7 +382,7 @@ cdef class cyDiscreteQuadraticModel:
                     raise ValueError("A variable has a self-loop")
 
         # add provided offset to dqm
-        dqm.offset_ = offset
+        dqm.offset = offset
 
         return dqm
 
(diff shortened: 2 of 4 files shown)


dimod/discrete/discrete_quadratic_model.py
```diff
@@ -279,12 +291,18 @@ def _from_file_numpy(cls, file_like):
 
         data = np.load(file_like)
 
+        try:
+            offset = data['offset']
```
arcondello, line 295: could just do `data.get('offset', 0)`

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -577,7 +598,7 @@ def set_quadratic_case(self, u, u_case, v, v_case, bias):
             self.variables.index(v), v_case,
             bias)
 
-    def _to_file_numpy(self, file, compress):
+    def _to_file_numpy(self, file, compress, save_offset=False):
```
arcondello, line 601: IMO always save the offset in `.to_file()`, but minor increment the version. to account for that.

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -587,20 +608,30 @@ def _to_file_numpy(self, file, compress):
...
             save = np.savez
 
-        save(file,
-             case_starts=vectors.case_starts,
-             linear_biases=vectors.linear_biases,
-             quadratic_row_indices=vectors.quadratic.row_indices,
-             quadratic_col_indices=vectors.quadratic.col_indices,
-             quadratic_biases=vectors.quadratic.biases,
-             )
+        if save_offset:
```
arcondello, line 618: probably a bit cleaner to make
```
data = dict(case_starts=....)
if save_offset:
    data.update(offset=vectors.offset)
save(file, **data)
```

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -609,7 +640,7 @@ def _to_file_numpy(self, file, compress):
         file.seek(end)
 
     def to_file(self, compress=False, compressed=None, ignore_labels=False,
-                spool_size=int(1e9)):
+                spool_size=int(1e9), save_offset=False):
```
arcondello, line 643: See above. IMO no kwarg. Also need to increment the serialization version.

dimod/discrete/cydiscrete_quadratic_model.pyx
```diff
@@ -18,6 +18,7 @@ cimport cython
 
 from cython.operator cimport preincrement as inc, dereference as deref
 from cython.operator cimport preincrement as preinc  # for when it matters
+from libcpp cimport bool
```
arcondello, line 21: no harm using `bool`, but `bint` does not need to be cimported.

dimod/discrete/cydiscrete_quadratic_model.pxd
```diff
@@ -40,6 +40,8 @@ cdef class cyDiscreteQuadraticModel:
     cdef vector[CaseIndex] case_starts_  # len(adj_) + 1
     cdef vector[vector[VarIndex]] adj_
 
+    cdef Bias offset_
```
arcondello, line 43: rather than making a private attribute with getter/setter, you can just do `cdef public Bias offset`

dimod/discrete/discrete_quadratic_model.py
hbarovertwo: Thanks, didn't think to check for a get method in that object

arcondello: I am only speculating that it has one, but worth checking :slightly_smiling_face:

dimod/discrete/cydiscrete_quadratic_model.pyx
hbarovertwo: I came across `bint` too and for the sake of import reduction, I'll use it

dimod/discrete/discrete_quadratic_model.py
hbarovertwo: Yep, there is one.

dimod/discrete/discrete_quadratic_model.py
hbarovertwo: So just update the second element of: https://github.com/dwavesystems/dimod/blob/a4bca1031f32d4d12235170704c6dab4144986bf/dimod/discrete/discrete_quadratic_model.py#L41 to 1?

arcondello: yup

hbarovertwo: Is your suggestion to set this to arg to `True` or to do away with it?

dimod/discrete/discrete_quadratic_model.py
hbarovertwo: I guess the above doesn't apply if we do away with `save_offset`

dimod/discrete/cydiscrete_quadratic_model.pxd
hbarovertwo: Done

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -793,13 +813,20 @@ def to_numpy_vectors(self):
             :meth:`~DiscreteQuadraticModel.from_numpy_vectors`
 
         """
-        case_starts, linear_biases, quadratic = self._cydqm.to_numpy_vectors()
+        if not return_offset:
+            case_starts, linear_biases, quadratic = self._cydqm.to_numpy_vectors()
+
+            return DQMVectors(case_starts,
```
arcondello, line 819: so I think this is actually still a backwards compatibility break, since
```
starts, ldata, quadratic = dqm.to_numpy_vectors()
```
will break (I think).

Unfortunately I think we need to have two different `DQMVectors`, maybe one called `LegacyDQMVectors` or something

tests/test_discrete_quadratic_model.py
```diff
@@ -813,7 +823,7 @@ def test_random_shuffled_quadratic(self):
         vectors = dqm.to_numpy_vectors()
 
         # suffle the quadratic vectors so they are not ordered anymore
-        starts, ldata, (irow, icol, qdata), labels = vectors
+        starts, ldata, (irow, icol, qdata), labels, offset = vectors
```
arcondello, line 826: see above, I think we need to keep this test the way it was.

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -793,13 +813,20 @@ def to_numpy_vectors(self):
             :meth:`~DiscreteQuadraticModel.from_numpy_vectors`
 
         """
-        case_starts, linear_biases, quadratic = self._cydqm.to_numpy_vectors()
+        if not return_offset:
+            case_starts, linear_biases, quadratic = self._cydqm.to_numpy_vectors()
```
arcondello, line 817: IMO raise a deprecation warning here saying that we will be making `return_offset` default to `True` in the future.

dimod/discrete/discrete_quadratic_model.py
hbarovertwo: I actually had both `DQMVectors` and `DQMVectorsWithOffset` at one point. But the above naming scheme makes sense along with the deprecation warning. Ended up changing it previously due to the default value feature of named tuples.

dimod/discrete/discrete_quadratic_model.py
hbarovertwo: Updated. Can we also add @hsadeghidw as a reviewer?

arcondello: I think he can just start reviewing if he wants to

hsadeghidw (approved):

arcondello (approved):

PR #839, opened 2021-06-10
Add minimal CQM implementation

arcondello:
Allows the construction of binary constrained quadratic models.

### DIFF
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -14,6 +14,7 @@
 
 import json
 import os.path
+import re
 import tempfile
 import uuid
 import zipfile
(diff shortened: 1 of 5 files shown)


arcondello:
Still need to fix the windows bug obviously. This also will need a pretty serious second pass when we add `QuadraticModel`

dimod/constrained.py
```diff
@@ -0,0 +1,395 @@
...
+
+        data = dict(num_variables=len(self.variables),
+                    num_constraints=len(self.constraints),
+                    num_biases=self.num_biases(),
+                    )
+
+        header_data = json.dumps(data, sort_keys=True).encode('ascii')
+        header_data += b'\n'
+        header_data += b' '*(64 - (len(prefix) + 6 + len(header_data)) % 64)
+        header_len = np.dtype('<u4').type(len(header_data)).tobytes()
```
randomir, line 364: I did just a quick check, but it looks `.tobytes()` in this setup actually always uses **native** byteorder. I believe your intention was to use little endian.

randomir (approved):
Apart from a few minor issues, LGTM!

dimod/constrained.py
```diff
@@ -0,0 +1,395 @@
...
+        return objective
+
+    def add_constraint(self, data, *args, **kwargs):
+        """A convenience wrapper for other methods that add constraints."""
+        if isinstance(data, BinaryQuadraticModel):
+            self.add_constraint_from_bqm(data, *args, **kwargs)
+        elif isinstance(data, Comparison):
+            self.add_constraint_from_comparison(data, *args, **kwargs)
+        elif isinstance(data, Iterable):
+            self.add_constraint_from_iterable(data, *args, **kwargs)
```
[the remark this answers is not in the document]
arcondello: Ah, used to have a comment about that, not sure where it went. But alas, `singledispatchmethod` is only available in 3.8+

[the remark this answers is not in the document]
randomir: Ah, I checked in docs for `singledispatch` (available from 3.4), but not for the method variant.

dimod/constrained.py
arcondello: Yup, excellent catch
```
In [6]: np.frombuffer(np.dtype('<u4').type(256).tobytes(), '<u4')
Out[6]: array([256], dtype=uint32)

In [7]: np.frombuffer(np.dtype('>u4').type(256).tobytes(), '<u4')
Out[7]: array([256], dtype=uint32)

In [8]: np.frombuffer(np.dtype('>u4').type(256).tobytes(), '>u4')
Out[8]: array([65536], dtype=uint32)

In [9]: np.frombuffer(np.dtype('<u4').type(256).tobytes(), '>u4')
Out[9]: array([65536], dtype=uint32)
```

randomir: I see we have the same bug in BQM and DQM.

arcondello: Indeed. Investigating

arcondello: Have a fix I'll make in a followup PR.

Issue #840, opened 2021-06-10
DQM variable case labeling


seatim:
**Application**

It is common for optimization practitioners to use labels to refer to cases of discrete variables in the same way as one does binary variables (e.g. "x11", "x12", "x13" for a discrete variable "x1" with three cases).

Another use case is where the cases of different variables have the same meaning, but the variables have a different number (or set) of cases.  For example, say variable "c1" has cases "red", "green", and "blue", and variable "c2" has cases "blue", and "yellow", and "brown".

DQM currently only supports the use of integers 0 .. N-1 to identify cases of variables with N cases.

**Proposed Solution**

Summary

Introduce an ability to refer to cases of discrete variables using labels, as an alternative to integers.  Case labels may be model-unique (unique within a model), or shared (unique within each variable, but shared among variables).

_Model-unique case labels_ are useful as they are a type of notation that is familiar to optimization practitioners.  These are semantically equivalent to sets of binary variables subject to one-hot constraints.

_Shared case labels_ are useful in that they allow cases to be labeled consistently by variables that have different numbers or sets of cases.

High-level Design

1. Introduce a subclass of DQM called `CaseLabelDQM` to offer both kinds of variable case labels.
2. Introduce `DQMCaseLabelComposite` class to map samples, translating both types of case labels.

CaseLabelDQM Methods

1. `add_variable(cases, label=None)`
    1. accept `Iterable` as value for `cases` parameter in addition to int.
    2. when `cases` is an `Iterable`:
        1. its values are case labels.  Labels can be any hashable object.
        2. when `label` is not `None`, shared case labels are defined. `ValueError` is raised if any label is repeated in cases.
        3. when `label` is `None`, model-unique case labels are defined. `ValueError` is raised if any label is not unique within the model among model-unique case labels and discrete variable labels.

    3. when `cases` is an int:
        1. no case labels are defined.
                OR: shared case labels `{ 1, ..., cases }` are created and `label` must not be `None`
                    NOTE: These are equivalent and the latter could be an implementation detail EXCEPT for the "label must not be None" part.
        2. behaves same as base class EXCEPT that if `label` is not `None`, it must be unique among model-unique case labels and discrete variable labels, else `ValueError` will be raised.

2. `get_linear(v)`
    `set_linear(v, biases)`
    - `v` may be a model-unique case label.

3. `get_linear_case(v, case)`
    `set_linear_case(v, case, bias)`
    - `case` may be a shared case label.

[...]

PR #844, opened 2021-06-14
Introduce CaseLabelDQM

seatim:
For #840.


seatim:
Note: this is a work in progress.  `CaseLabelDQM` is functional, `DQMCaseLabelComposite` has not been added yet.

requirements.txt
```diff
@@ -1,2 +1,3 @@
 numpy==1.19.4
 cython==0.29.21
+parameterized==0.8.1
```
arcondello, line 3: This is already in `tests/requirements.txt`

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -810,3 +808,315 @@ def to_numpy_vectors(self, return_offset: bool = False):
...
+            v: A variable in the discrete quadratic model.
+
+            v_case: The case of `v`.
+
+            bias (float): The quadratic bias.
+
+        """
+        u_case = self._lookup_shared_case(u, u_case)
+        v_case = self._lookup_shared_case(v, v_case)
+        DQM.set_quadratic_case(self, u, u_case, v, v_case, bias)
```
arcondello, line 1122: In this case is there a reason to use `DQM` rather than `super()`?

requirements.txt
seatim: Thanks, fixed.

dimod/discrete/discrete_quadratic_model.py
seatim: It's explicit.

arcondello: Sure, but what does the explicitness give us in this case? Because on the down side it makes later subclassing harder and/or messes with multiple inheritance. Especially true for the one in the `__init__`.

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -810,3 +808,315 @@ def to_numpy_vectors(self, return_offset: bool = False):
...
+
+                If `u` and `v` are unique case labels, the quadratic bias is a
+                real number.
+
+        Raises:
+            ValueError: If `u` is a unique case label and `v` is not.
+
+        """
+        u_k = self._unique_label_case.get(u)
+        if u_k:
```
arcondello, line 1095: unless I am missing something, this can be Falsey for lots of cases. E.g. if the label is `''` or `0`.

seatim: 🤦 You're right, of course!

dimod/discrete/discrete_quadratic_model.py
seatim: Explicitness reduces surprises.  But, I understand the scenarios you allude to where having used `super()` here would be desirable.  So, happy to make the change.

dimod/discrete/discrete_quadratic_model.py
seatim: Leaving aside the code and considering the design space, is there a good reason to support `''` or non-strings as labels?  Supporting non-empty strings only, as labels, covers both example use cases given in #840.  The "labels can be any hashable object" design can be revised.

dimod/discrete/discrete_quadratic_model.py
seatim: Did that.

dimod/discrete/discrete_quadratic_model.py
arcondello: Frozensets and tuples are both useful in some cases. And if we're going to use strings, why not include `''`? And in general, why restrict the user? Hashable is well defined in python. For what it's worth, there is actually one restriction, which is that we don't allow `None`, which is hashable. In this case I think you could just do `if u_k is not None`.

randomir (approved):
LGTM

dimod/discrete/discrete_quadratic_model.py
seatim: Update: `u_k` will be either a 2-tuple - a (var, case) pair - or `None`.  So this code is correct.  I do need to add a check for case labels being `None` to `add_variable()` though.

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -810,3 +808,315 @@ def to_numpy_vectors(self, return_offset: bool = False):
 
 # register fileview loader
 load.register(DQM_MAGIC_PREFIX, DiscreteQuadraticModel.from_file)
+
+
+class CaseLabelDQM(DQM):
```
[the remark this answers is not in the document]
seatim: I considered those options and thought they were a bit unwieldy.  But, I'm certainly open to changing the name if there's consensus.

dimod/discrete/discrete_quadratic_model.py
seatim: Hmm, or perhaps better yet, we could allow `None` as a case label.

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -810,3 +808,315 @@ def to_numpy_vectors(self, return_offset: bool = False):
...
+        """
+        map_ = self._shared_label_case.get(v)
+        if map_:
+            if case not in map_:
+                raise ValueError(f'unknown case {case} of variable {v}')
+            return map_[case]
+        return case
+
+    def get_linear(self, v):
+        """The linear bias(es) associated with variable `v`.
```
JoelPasvolsky, line 940: "The linear biases associated ... " is more readable and no one will hold the one-case corner case against you .
(Also in the returns description here and some other places)

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -810,3 +808,315 @@ def to_numpy_vectors(self, return_offset: bool = False):
...
+
+        Args:
+            v: A variable in the discrete quadratic model, or a unique case
+                label.
+
+        Returns:
+            The linear bias(es).  If `v` is a variable, returns a
+                :meth:`~DiscreteQuadraticModel.num_cases(v)` by 1 numpy array.
```
JoelPasvolsky, line 948: "numpy" --> Numpy
Would be easier to parse as "... returns a Numpy array of size :meth:`~DiscreteQuadraticModel.num_cases(v)` by 1"
(Also elsewhere below)

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -810,3 +808,315 @@ def to_numpy_vectors(self, return_offset: bool = False):
...
+
+    def set_linear(self, v, biases):
+        """Set the linear bias(es) associated with `v`.
+
+        Args:
+            v: A variable in the discrete quadratic model, or a unique case
+                label.
+
+            biases (float or array-like):  If `v` is a variable, the linear
+                biases in an array.  Otherwise, the linear bias is a real
```
JoelPasvolsky, line 1044: "in" --> is

dimod/discrete/discrete_quadratic_model.py
arcondello: We could for sure. FWIW, the reason that we don't allow `None` as a variable label is so that we can allow labels as optional keyword arguments easily. E.g. `variables._append(label=None)` does not mean that the variable should be labelled `None`, but rather that a new label should be generated. It might make sense to reserve that case for the future, or not :shrug:

dimod/discrete/discrete_quadratic_model.py
seatim: Just to note I see we use "numpy" in three pre-existing doc strings: `DQM.to_numpy_vectors()`, `DQM.from_numpy_vectors()`, and `DQM.get_quadratic()`, so this will introduce a minor inconsistency, unless we change them as well.

arcondello: We're inconsistent about this all over Ocean. `NumPy` is actually how they refer to themselves so I think we should use that when we remember.

JoelPasvolsky: Please do. And it's actually NumPy (sorry, my typo)

arcondello: I don't think that change should be made as part of this PR.

dimod/discrete/discrete_quadratic_model.py
seatim: Fixed.

dimod/discrete/discrete_quadratic_model.py
seatim: Fixed.

dimod/discrete/discrete_quadratic_model.py
seatim: Done.

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -810,3 +808,315 @@ def to_numpy_vectors(self, return_offset: bool = False):
...
+    1. Unique case labels are unique among variable labels and themselves.
+
+    2. Shared case labels are unique among cases for a variable, but may be
+        reused among variables.
+
+    Examples:
+
+        Declare variables with unique case labels.
+
+        >>> dqm = dimod.DiscreteQuadraticModel()
```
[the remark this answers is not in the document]
seatim: Fixed.

seatim:
@arcondello Where should `DQMCaseLabelComposite` be defined?  New file in `dimod/discrete/`?

arcondello:
We don't "officially" support composites for DQM samplers. That said, IMO should be in a followup PR in under the `dimod/composites/` namespace.

Another approach would be adding a `flatten_sampleset` function that turns `INTEGER` samplesets into `BINARY` ones. That is the main functionality added, right?

arcondello:
What is the behavior when I do `dqm.to_numpy_vectors` etc. Specifically I believe
```
new_dqm = CaseLabelDQM.from_file(dqm.to_file())
assert new_dqm == dqm
```
will fail. Same with `.to_numpy_vectors` etc.

seatim:
> Another approach would be adding a flatten_sampleset function that turns INTEGER samplesets into BINARY ones. That is the main functionality added, right?

For unique case labels, yes.  For shared, the _values_ in the sampleset are mapped to the labels.  Both types of transformation are performed.

seatim:
> What is the behavior when I do `dqm.to_numpy_vectors` etc.

Correct, you won't get a `CaseLabelDQM` back, at this point.

arcondello:
> Correct, you won't get a `CaseLabelDQM` back, at this point.

To me that seems like a poor user experience. Especially since that's what will happen if you submit a `CaseLabelDQM` to `LeapHybridDQMSampler`. I think we should at least overwrite the methods to raise `NotImplementedError`.

seatim:
> > What is the behavior when I do `dqm.to_numpy_vectors` etc.
> 
> Correct, you won't get a `CaseLabelDQM` back, at this point.

Actually, that assert may not trigger.  I don't know `DQM` well enough to say.  Case labels are not serialized though, at this point.

arcondello:
You should get a `CaseLabelDQM` back, but it won't have labels on the cases.

seatim:
> Also, I don't think __eq__ is defined for DQM, but you get the idea.

Ha, yeah, I just reached the same conclusion.

> [..] but you get the idea.

Not totally sure I do.  I don't think we should speculate about some behavior that is important.  So, if conserving that (assumed) `DQM` behavior is important, we should first have a unit test to cover that behavior in `DQM`.  That would mean defining `DQM.__eq__()`.  Then second we should have a unit test to cover the conservation of behavior in `CaseLabelDQM`.  Is that what you are thinking?

arcondello:
We use https://github.com/dwavesystems/dimod/blob/43c5c57e8b2bd5c29e25b54275fbbbc534f42fdc/tests/test_discrete_quadratic_model.py#L275 in the unittests already.

I don't think we should implement `__eq__` for `DQM`, I somewhat regret doing so for `BQM`. But you could imagine doing `DQM.is_equal(other)`. Similar to https://github.com/dwavesystems/dimod/blob/43c5c57e8b2bd5c29e25b54275fbbbc534f42fdc/dimod/binary/binary_quadratic_model.py#L1198

seatim:
Right on, thanks.  I'll use `assertDQMEqual` for this.

arcondello:
BTW, that equality might be missing the offset added in https://github.com/dwavesystems/dimod/pull/838. Not really that relevant in this case but on the change you're planning to generalize.

seatim:
> Note: this is a work in progress. `CaseLabelDQM` is functional, `DQMCaseLabelComposite` has not been added yet.

This PR is functionally complete.  The design changed to use a utility method `CaseLabelDQM.map_sample()` instead of a composite.

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -810,3 +808,372 @@ def to_numpy_vectors(self, return_offset: bool = False):
...
+        if map_:
+            return [map_[case] for case in range_]
+
+        elif v in self._unique_label_vars:
+            return [self._unique_case_label[(v, case)] for case in range_]
+
+        else:
+            return list(range_)
+
+    def to_file(self, *args, **kwargs):
```
arcondello, line 1148: Think we need to do the same thing with `to_` and `from_numpy_vectors`.

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -810,3 +808,372 @@ def to_numpy_vectors(self, return_offset: bool = False):
...
+        raise NotImplementedError
+
+    @classmethod
+    def from_file(cls, *args, **kwargs):
```
arcondello, line 1152: I think we could actually leave this as-is, and support the
```
dqm = DQM()
...
cldqm = CaseLabelDQM.from_file(dqm.to_file())
```

On the other hand, it might be confusing to support `from_` but not `to_`. Up to you :shrug:

arcondello (commented):
Looking good! Two more things:
1) Can you add it to the docs? See https://github.com/dwavesystems/dimod/blob/main/docs/reference/dqm.rst
2) Can you add it to the release notes? See https://github.com/dwavesystems/dimod#release-notes. Let me know if you have trouble or it's not clear.

dimod/discrete/discrete_quadratic_model.py
seatim: There's one issue with that: `to_numpy_vectors()` is called indirectly by `DQM.to_file()`, so making it raise `NotImplementedError` means that [this change](https://github.com/dwavesystems/dwave-system/pull/409/commits/cace0f698747cf58d7bcae6a400dfffbbfa3ff14#diff-65eb42d3a55da8bbe2c8474ba04bfe2a047b27372655f8e2f14141a3da3ea5d2R450) won't work.  🤔

arcondello: Ah, good point. Ok, I think we can just leave it for now. See if anyone complains. The longer solution would be to add
```
LabelDQMVectors = namedtuple('DQMVectors', ['case_starts', 'linear_biases', 'quadratic', 'labels', 'offset', 'case_labels'])
```
or something like that, but maybe we can make an issue for that and call it done for now.

arcondello (approved):

JoelPasvolsky (approved):

PR #852, opened 2021-06-21
Fix subtracting a BQM from a number

arcondello:


randomir (approved):
LGTM.

Also, unrelated, love having release notes in each PR.

Issue #664, opened 2020-06-09
Integer Vartype


arcondello:
Add a new `Vartype.Integer`. This would allow us to handle integer quadratic models and discrete quadratic models.

Obviously this would be changing a core assumption in dimod so must be done very carefully.

PR #853, opened 2021-06-21
Add Vartype.INTEGER

arcondello:
Closes https://github.com/dwavesystems/dimod/issues/664

### DIFF
--- a/dimod/vartypes.py
+++ b/dimod/vartypes.py
@@ -55,11 +55,26 @@ Examples:
 """
 import enum
 import typing
+import warnings
+
+from collections.abc import Container, Iterable
 
 __all__ = ['as_vartype',
            'Vartype', 'ExtendedVartype',
-           'SPIN', 'BINARY', 'DISCRETE',
+           'SPIN', 'BINARY', 'DISCRETE', 'INTEGER',
            ]
+
+
+class Integers(Container):
+    """Container for testing integer membership."""
+    def __contains__(self, item):
+        try:
+            return int(item) == item
+        except TypeError:
+            return False
+
+
+integers = Integers()
 
 
 class Vartype(enum.Enum):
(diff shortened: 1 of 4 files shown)


dimod/vartypes.py
```diff
@@ -75,24 +90,38 @@ class Vartype(enum.Enum):
...
+def _vartype_miss(vartype, extended):
+    if extended:
+        candidates = ("Vartype.SPIN, 'SPIN', {-1, 1}, "
+                      "Vartype.BINARY, 'BINARY', {0, 1},"
+                      "Vartype.INTEGER, or 'INTEGER'")
+    else:
+        candidates = ("Vartype.SPIN, 'SPIN', {-1, 1}, "
+                      "Vartype.BINARY, 'BINARY', or {0, 1}")
+    return TypeError(f"expected input vartype to be one of: {candidates!s}; "
+                     f"recieved {vartype!r}.")
```
randomir, line 119: typo

dimod/vartypes.py
```diff
@@ -75,24 +90,38 @@ class Vartype(enum.Enum):
...
+
+
+# In the future we may wish to make extended default to True. However, for
+# now even raining a PendingDeprecationWarning would be a big change.
```
randomir, line 123: raining?

dimod/vartypes.py
```diff
@@ -75,24 +90,38 @@ class Vartype(enum.Enum):
...
-    DISCRETE = int
+# Deprecated alias
+ExtendedVartype = Vartype
 
 
 SPIN = Vartype.SPIN
 BINARY = Vartype.BINARY
-DISCRETE = ExtendedVartype.DISCRETE
+INTEGER = Vartype.INTEGER
+DISCRETE = Vartype.DISCRETE
```
randomir, line 104: sorting by line length for the win! :grin:

dimod/vartypes.py
```diff
@@ -55,13 +55,28 @@
...
+class Integers(Container):
+    """Container for testing integer membership."""
+    def __contains__(self, item):
+        try:
+            return int(item) == item
+        except TypeError:
+            return False
+
+
+integers = Integers()
```
randomir, line 77: I'm sure there's a good reason for this, but it's not obvious to me. Where is membership testing used?

randomir (approved):
LGTM

dimod/vartypes.py
arcondello: An example would be something like https://github.com/dwavesystems/dimod/blob/63d17243b2aab52852df8e4e8760150ee2bc40c2/dimod/core/bqm.py#L1372

dimod/vartypes.py
arcondello: :umbrella:

dimod/vartypes.py
randomir: Of course. Enum's value is actually a domain of the type. Forgot that.

PR #855, opened 2021-06-21
Loosen numpy version range in pyproject.toml

arcondello:


arcondello:
Nevermind, I guess it actually updates the NumPy version even if one within range in pre-installed. Maybe there is a way to configure that, will look into it later.

arcondello:
For posterity, can use `pip wheel . --no-build-isolation`

PR #856, opened 2021-06-22
README update

arcondello:


README.rst
```diff
@@ -60,24 +57,17 @@ Installation
...
 
-    pip install dimod[all]
-
-To install from source:
-
-.. code-block:: bash
-
-    pip install -r requirements.txt
-    python setup.py install
+    pip install .
```
JoelPasvolsky, line 70: I get an error:
```
ERROR: Could not install packages due to an OSError: [('C:\\Users\\jpasvolsky\\!git_ADTT\\test_dimod\\dimod\\docs\\LICENSE', 'C:\\Users\\jpasvolsky\\AppData\\Local\\Temp\\pip-req-build-w6ovc33b\\docs\\LICENSE', 'symbolic link privilege not held'), ('C:\\Users\\jpasvolsky\\!git_ADTT\\test_dimod\\dimod\\docs\\README.rst', 'C:\\Users\\jpasvolsky\\AppData\\Local\\Temp\\pip-req-build-w6ovc33b\\docs\\README.rst', 'symbolic link privilege not held')]
```

arcondello: Incredible. Windows is why we cannot have nice things.

README.rst
```diff
@@ -89,6 +79,35 @@ in place:
 
 .. installation-end-marker
 
+Contributing
```
JoelPasvolsky, line 82: For consistency: we usually have (1) license before contrib and (2) most have "Contribution" though personally I prefer "Contributing"

README.rst
```diff
@@ -89,6 +79,35 @@ in place:
...
+------------
+
+Ocean's `contributing guide <https://docs.ocean.dwavesys.com/en/stable/contributing.html>`_
+has guidelines for contributing to Ocean packages.
+
+dimod includes some formatting customization in the
+`.clang-format <.clang-format>`_ and `setup.cfg <setup.cfg>`_ files.
+
+Reno
```
JoelPasvolsky, line 91: Can we have the header "Release Notes" instead? That will be self-explanatory also to users who are not familiar with this particular package

JoelPasvolsky (approved):

README.rst
arcondello: I will do (1) but not (2)

README.rst
arcondello: For posterity, you can also do `pip install . --use-feature=in-tree-build` which will become the default behaviour in the future. For now I have reverted to the old syntax.

PR #857, opened 2021-06-22
Add `scale` method to `QuadraticModelBase`, with test

### LINKED ISSUE #833: QuadraticModel::scale in C++ code

hbarovertwo:
Implement `scale` method in `QuadraticModelBase`.
Adjust offset, linear biases, and interactions by a scale factor.
Add test to reflect addition.

Closes #833

### DIFF
--- a/dimod/include/dimod/quadratic_model.h
+++ b/dimod/include/dimod/quadratic_model.h
@@ -613,8 +613,7 @@ class QuadraticModelBase {
     }
 
     /// Scale offset, linear biases, and interactions by a factor
-    void scale(double scale_factor, int ignored_variables = NULL,
-               int ignored_interactions = NULL) {
+    void scale(double scale_factor) {
         // adjust offset
         offset() *= scale_factor;
 
--- a/dimod/include/dimod/quadratic_model.h
+++ b/dimod/include/dimod/quadratic_model.h
@@ -618,11 +618,11 @@ class QuadraticModelBase {
         offset() *= scale_factor;
 
         // adjust linear biases and quadratic interactions
-        for (size_type ui = 0; ui < num_variables(); ui++) {
-            linear_biases_[ui] *= scale_factor;
+        for (size_type u = 0; u < num_variables(); u++) {
+            linear_biases_[u] *= scale_factor;
 
-            auto begin = adj_[ui].begin();
-            auto end = adj_[ui].end();
+            auto begin = adj_[u].begin();
+            auto end = adj_[u].end();
             for (auto nit = begin; nit != end; ++nit) {
                 (*nit).second *= scale_factor;
             }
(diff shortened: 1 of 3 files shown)


dimod/include/dimod/quadratic_model.h
```diff
@@ -612,6 +612,23 @@ class QuadraticModelBase {
         }
     }
 
+    /// Scale offset, linear biases, and interactions by a factor
+    void scale(double scale_factor) {
+        // adjust offset
+        offset() *= scale_factor;
+
+        // adjust linear biases and quadratic interactions
+        for (size_type ui = 0; ui < num_variables(); ui++) {
```
arcondello, line 621: This is a nit, but for consistency: we use `u` and `v` in the C++ code.

It's only in cython that we use `ui` and `vi` to distinguish the integer labels from the hashable labels.

arcondello (approved):
LGTM

dimod/include/dimod/quadratic_model.h
hbarovertwo: That is consistent with most of the file, although later in `change_vartype` I saw `ui` was used. Updated this.

arcondello:
Oh, sorry, one more (new) step to the process. Can you add the new method to the release notes? See https://github.com/dwavesystems/dimod#release-notes. Let me know if you have trouble or it's not clear. This is a brand new thing we're trying in dimod. See #848.

hbarovertwo:
> Oh, sorry, one more (new) step to the process. Can you add the new method to the release notes? See https://github.com/dwavesystems/dimod#release-notes. Let me know if you have trouble or it's not clear. This is a brand new thing we're trying in dimod. See #848.

Added the release note, let me know if it seems clear enough

releasenotes/notes/quadratic_model_scale_method-52f5a73ec045801c.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    Add new `scale` method to `QuadraticModelBase`. Eventually to be ported to python
```
arcondello, line 4: I think this can be more simply
```
Add `QuadraticModelBase::scale` method. Scales the offset, linear biases and quadratic biases by a provided scale factor.
```
chances are the port will be included in the same release so I don't think we need to say anything about it.

hbarovertwo: Thanks

PR #859, opened 2021-06-24
Add `dimod.views.quadratic.QuadraticViewsMixin`

arcondello:
Can be used in the upcoming `QuadraticModel` implementation.

I don't think any of these changes are user-facing enough to be worth adding to the changelog.


randomir (approved):
lgtm

hhtong (approved):
LGTM

PR #860, opened 2021-06-26
Make use of QuadraticModelBase::scale in dimod.binary.BQM

hbarovertwo:
Port `QuadraticModelBase::scale` to Python layer of BQM.
Benchmark done using `dimod.generators.random.gnm_random_bqm`, fixed with 100,000 variables and 1,000,000 interactions. Comparison done with `cls = AdjVectorBQM` and `cls = dimod.binary.BinaryQuadraticModel` respectively:

`1.4 s ± 7.43 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)` (`AdjVectorBQM` scale method pure python)

`102 ms ± 651 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)` (`BinaryQuadraticModel` defaulting to C++ code)

### DIFF
--- a/dimod/binary/binary_quadratic_model.py
+++ b/dimod/binary/binary_quadratic_model.py
@@ -1225,8 +1225,7 @@ class BinaryQuadraticModel(QuadraticViewsMixin):
 
     def scale(self, scalar, ignored_variables=None, ignored_interactions=None,
               ignore_offset=False):
-        """Multiply all the biases by the specified scalar. Default to C++
-        implementation with default args, otherwise fallback on Python.
+        """Multiply all the biases by the specified scalar. 
 
         Args:
             scalar (number):
@@ -1244,14 +1243,14 @@ class BinaryQuadraticModel(QuadraticViewsMixin):
                 If True, the offset is not scaled.
 
         """
-        if (
-            ignored_variables, ignored_interactions, ignore_offset
-        ) == (None, None, False) and isinstance(
-            self.data, (cyBQM_float32, cyBQM_float64)
-        ):
-            self.data.scale(scalar)
-            return
-        
+        if ignored_variables is None and ignored_interactions is None \
+            and ignore_offset is False:
+            try:
+                self.data.scale(scalar)
+                return
+            except AttributeError:
+                pass
+
         if ignored_variables is None:
             ignored_variables = set()
         elif not isinstance(ignored_variables, abc.Container):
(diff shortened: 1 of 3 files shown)


dimod/binary/binary_quadratic_model.py
```diff
@@ -1225,7 +1225,8 @@ def resize(self, n: int):
 
     def scale(self, scalar, ignored_variables=None, ignored_interactions=None,
               ignore_offset=False):
-        """Multiply all the biases by the specified scalar.
+        """Multiply all the biases by the specified scalar. Default to C++
```
arcondello, line 1228: I would not include the default/fallback behavior in the first line of the docstring, see [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings).

I probably wouldn't document this implementation detail at all, though if you want to include it, the second paragraph would be the appropriate place.

dimod/binary/binary_quadratic_model.py
```diff
@@ -1243,8 +1244,14 @@ def scale(self, scalar, ignored_variables=None, ignored_interactions=None,
                 If True, the offset is not scaled.
 
         """
-        # this could be cythonized for performance
-
+        if (
```
arcondello, line 1247: It's a minor thing, but IMO is more readable and extensible in the future.
```
if ignored_variables is None and ignored_interactions is None and not ignore_offset:
    try:
        self.data.scale(scalar)
        return
    except AttributeError:
        pass
```

dimod/binary/cybqm/cybqm_template.pxd.pxi
```diff
@@ -72,6 +72,7 @@ cdef extern from "dimod/quadratic_model.h" namespace "dimod" nogil:
         size_type num_interactions(index_type)
         bint remove_interaction(index_type, index_type) except +
         void resize(index_type)
+        void scale(cython.double)
```
arcondello, line 75: since the underlying C++ code takes `bias_type`, you should just use `bias_type` here as well. The compiler will figure it out either way but no point making it harder for it :smile:

dimod/binary/binary_quadratic_model.py
hbarovertwo: Will remove

dimod/binary/cybqm/cybqm_template.pxd.pxi
hbarovertwo: I had a feeling to go that way but wasn't sure if that would cause any issues. If it's not an issue, then it is preferable to use the template type alias.

arcondello (approved):

PR #865, opened 2021-06-30
Add `QuadraticModel`

arcondello:
Still missing:
- docs
- serialization
- symbolic manipulation for `<=`, `==`, `>=`
- integration into `CQM`
- Support for specifying lower and upper bounds for integer variables

But I think this is enough to review for now, can add other things in subsequent PRs.


dimod/include/dimod/quadratic_model.h
```diff
@@ -873,6 +609,7 @@ class BinaryQuadraticModel : public QuadraticModelBase<Bias, Index> {
     /// Add quadratic bias for the given variables.
     void add_quadratic(index_type u, index_type v, bias_type bias) {
         if (u == v) {
+            assert(0 <= u && static_cast<size_t>(u) < this->num_variables());
```
randomir, line 612: Shouldn't you want to keep these checks in release version as well (not just debug)?

arcondello: This is following C++'s style of allowing undefined behaviour in the interests of speed. So at the C++ level we allow segfaults. We do check this stuff at the Python level Python doesn't allow segfaults.

randomir: If we'll stick to allowing segfaults in c++, once we stabilize/open c++ interface for devs, we'll need to heavily document that.

arcondello: Totally agree. I intend to take a pass through and mark the undefined behavior and/or exception safety appropriately. Just focusing on the functionality for the moment.

dimod/include/dimod/iterators.h
```diff
@@ -0,0 +1,319 @@
...
+        }
+        if (i_ >= 0 && i_ < (index_type)neighborhood_ptr_->size()) {
+            pair_ptr_ = new value_type(neighborhood_ptr_->neighbors[i_],
+                                       neighborhood_ptr_->quadratic_biases[i_]);
+        }
+    }
+};
+
+template <class Bias, class Index>
+class ConstQuadraticIterator {
```
amahmudwave, line 247: The workings of ConstQuadraticIterator is much more complex than the NeighborhoodIterators.  It is better to add more elaborate comments to the advance function, with complete sentences.

dimod/include/dimod/quadratic_model.h
```diff
@@ -143,6 +139,12 @@ class Neighborhood {
         return iterator(this, std::distance(neighbors.begin(), it));
     }
 
+    /// Return an iterator to the first element that does not come before `v`.
```
amahmudwave, line 142: before v or after v ?

dimod/include/dimod/quadratic_model.h
```diff
@@ -480,6 +260,17 @@ class QuadraticModelBase {
 
     QuadraticModelBase() : offset_(0) {}
 
+    /// Add quadratic bias for the given variables.
+    void add_quadratic(index_type u, index_type v, bias_type bias) {
+        assert(0 <= u && static_cast<size_t>(u) < this->num_variables());
```
amahmudwave, line 265: I generally write it as (assert u >= 0 && ..... ) 
The second part of your condition is like that too.

dimod/include/dimod/quadratic_model.h
```diff
@@ -583,14 +374,22 @@ class QuadraticModelBase {
...
+     */
     size_type num_interactions() const {
         size_type count = 0;
+        for (size_type v = 0; v < this->num_variables(); ++v) {
+            count += this->adj_[v].size();
 
-        for (auto it = adj_.begin(); it != adj_.end(); ++it) {
-            count += (*it).size();
+            // account for self-loops
+            auto lb = this->adj_[v].lower_bound(v);
```
amahmudwave, line 388: I think you need to subtract 1 if self loop is present, other wise keep it unchanged.

dimod/include/dimod/quadratic_model.h
```diff
@@ -629,6 +428,16 @@ class QuadraticModelBase {
         }
     }
 
+    void set_quadratic(index_type u, index_type v, bias_type bias) {
+        assert(0 <= u && static_cast<size_t>(u) < this->num_variables());
+        assert(0 <= v && static_cast<size_t>(v) < this->num_variables());
```
amahmudwave, line 433: Same as above.

dimod/include/dimod/quadratic_model.h
```diff
@@ -1094,6 +849,123 @@ class BinaryQuadraticModel : public QuadraticModelBase<Bias, Index> {
...
+    VarInfo(Vartype vartype, T lb, T ub) : vartype(vartype), lb(lb), ub(ub) {}
+};
+
+template <class Bias, class Index = int>
+class QuadraticModel : public QuadraticModelBase<Bias, Index> {
+ public:
+    /// The type of the base class.
+    using base_type = QuadraticModelBase<Bias, Index>;
+
+    /// The first template parameter (Bias).
```
amahmudwave, line 867: If the template parameter order is changed, these comments need to be changed. ie first, second.

dimod/include/dimod/quadratic_model.h
```diff
@@ -1094,6 +849,123 @@ class BinaryQuadraticModel : public QuadraticModelBase<Bias, Index> {
...
+            this->varinfo_.insert(this->varinfo_.begin(), bqm.num_variables(),
+                                  VarInfo<bias_type>(Vartype::BINARY, 0, 1));
+        } else {
+            throw std::logic_error("unexpected vartype");
+        }
+    }
+
+    void add_quadratic(index_type u, index_type v, bias_type bias) {
+        if (u == v) {
+            assert(0 <= u && static_cast<size_t>(u) < this->num_variables());
```
amahmudwave, line 894: Same as before. I generally write a macro for these asserts. If they are going to appear too many times in a file.

#define ASSERT_CHECK_VARIABLE(var_type, v) \\
    assert(v >= 0 && v < this->num_variables(var_type))

dimod/include/dimod/iterators.h
```diff
@@ -0,0 +1,319 @@
...
+    value_type* pair_ptr_;
+
+    void update_pair() {
+        if (pair_ptr_) {
+            delete pair_ptr_;
+        }
+        if (i_ >= 0 && i_ < (index_type)neighborhood_ptr_->size()) {
+            pair_ptr_ = new value_type(neighborhood_ptr_->neighbors[i_],
+                                       neighborhood_ptr_->quadratic_biases[i_]);
+        }
```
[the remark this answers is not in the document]
arcondello: No, this case happens typically when the iterator goes "out of range", e.g. for the end.

dimod/include/dimod/quadratic_model.h
arcondello: We divide by 2 at the end. So we need to double count the self-loops. In this case we want to count self-loops as interactions. E.g. `i**2` when `i` is an integer is considered an interaction.

dimod/include/dimod/quadratic_model.h
arcondello: I like the explicitness of putting the actual assert in.

arcondello:
For some of the style comments on iterators I made a new issue (https://github.com/dwavesystems/dimod/issues/871). This PR just moved that code into another file so I'd rather make those changes as part of an independent PR.

dimod/include/dimod/quadratic_model.h
arcondello: Haha, I am used to python style `assert 0 <= u < self.num_variables())`

dimod/include/dimod/quadratic_model.h
amahmudwave: You can add this reply as a comment :)

PR #872, opened 2021-07-02
Use new C++ BQM in DiscreteQuadraticModel

arcondello:
Use the new C++ BQM introduced in https://github.com/dwavesystems/dimod/pull/818 in `DiscreteQuadraticModel`.

*Additional Context*
`DQM.add_linear_equality_constraint` performance has changed. Using the following script
```
import time

import numpy as np
import dimod

num_variables = 1000
num_cases = 10
num_trials  = 5

ordered = True


rng = np.random.default_rng(42)


terms_list = [[(v, rng.choice(num_cases), 1) for v in range(num_variables)] for _ in range(10)]

if not ordered:
    for terms in terms_list:
        rng.shuffle(terms)

single = 0
additive = 0

for _ in range(num_trials):
    dqm = dimod.DQM()
        dqm.add_variable(num_cases)

    t = time.time()
    dqm.add_linear_equality_constraint(terms_list[0], 1, 1)
    single += time.time() - t

    t = time.time()
    for terms in terms_list:
        dqm.add_linear_equality_constraint(terms, 1, 1)
    additive += time.time() - t

print('single', single)
print('additive', additive)
```
to compare between this PR and https://github.com/dwavesystems/dimod/tree/06d2fd18b7bd8c12f75976675f4e7f7f32b6b512 on my system gives
* Main
    * single: ~.06
    * additive: ~5.77
* PR
    * single: ~.13
    * additive: ~4.7

So took a performance hit when adding to an empty DQM but improved performance when adding multiple constraints.

### DIFF
--- a/dimod/discrete/cydiscrete_quadratic_model.pxd
+++ b/dimod/discrete/cydiscrete_quadratic_model.pxd
@@ -23,10 +23,8 @@ cimport numpy as np
 from dimod.bqm.cppbqm cimport AdjVectorBQM as cppAdjVectorBQM
 from dimod.bqm.common cimport Integral32plus, Numeric, Numeric32plus
 
-
-ctypedef np.float64_t Bias
-ctypedef np.int64_t CaseIndex
-ctypedef np.int64_t VarIndex
+ctypedef np.float64_t bias_type
+ctypedef np.int64_t index_type  # todo: int32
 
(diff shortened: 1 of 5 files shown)


arcondello:
Forgot to mention, the performance is about the same for `ordered=True` and `ordered=False`.

Issue #873, opened 2021-07-05
Specify setup_requires in setup.py.


29rou:
**Application**
Now, We need "pip install -r requirements.txt", before run setup.py. 
However, I think that is a bit cumbersome.

**Proposed Solution**
Specify setup_requires in setup.py.

PR #874, opened 2021-07-05
Specify setup requires

29rou:
For issue https://github.com/dwavesystems/dimod/issues/873#issue-936620223


arcondello:
We use `pyproject.toml` (see [PEP 518](https://www.python.org/dev/peps/pep-0518/)) for build requirements. We don't include `reno` in the `setup.py` or `setup.cfg` because it is not required to use the package, just to make contributions.

29rou:
Thank you, I understand.

PR #886, opened 2021-07-06
Improve performance of copy operations

arcondello:


dimod/cyvariables.pyx
```diff
@@ -12,6 +12,7 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+from copy import deepcopy
```
hhtong, line 15: Don't think we need this import

hhtong (approved):

dimod/binary/vartypeview.py
```diff
@@ -58,6 +59,14 @@ def __init__(self, data, vartype: Vartype):
         self.data = data
         self._vartype = vartype
 
+    def __copy__(self):
+        # since we'd need to copy he underlying data anyway, let's just return
+        # that instead. It doesn't really make sense to maintain the view
+        # of a detatched copy.
```
hhtong, line 65: typos (the, detached)

hhtong (approved):

PR #887, opened 2021-07-07
Implement copy constructor and assignment operator for C++ iterators

arcondello:


dimod/include/dimod/iterators.h
```diff
@@ -45,6 +45,23 @@ class NeighborhoodIterator {
         update_pair();
     }
 
+    NeighborhoodIterator(const NeighborhoodIterator<Bias, Index>& other)
```
amahmudwave, line 48: This looks good but declaring an explicit copy constructor may invalidate the implicit destructor and move constructor. The rules are listed here, it is not absolutely needed to define the move constructor if it is deleted due to the declaration of explicit copy constructor, but we may want to do that to improve speed. And the destructor is needed to avoid memory leaks. 

Happy reading :) 
https://stackoverflow.com/questions/46402983/rule-of-five-and-implicitly-deleted-functions

dimod/include/dimod/iterators.h
```diff
@@ -37,14 +37,47 @@ class NeighborhoodIterator {
...
+
+    NeighborhoodIterator<Bias, Index>& operator=(
+            NeighborhoodIterator<Bias, Index> const& other) {
+        if (this == &other) return *this;
+        neighborhood_ptr_ = other.neighborhood_ptr_;
+        i_ = other.i_;
+        update_pair();  // handles pair_ptr_
+    }
+
+    NeighborhoodIterator<Bias, Index>& operator=(
```
amahmudwave, line 75: Dont you want to have the same check if (this == &other) like the other one ?

amahmudwave: Is swapping okay over here ?  or we need to use std::move ? and just before that delete the elements of the one being assigned to ?

Swapping makes sense for the move constructor though.

arcondello: I think so? https://stackoverflow.com/questions/276173/what-are-your-favorite-c-coding-style-idioms/2034447#2034447

amahmudwave: I think swap will swap the data, but I think we want the rhs side objects data to be invalidated. For cases such as these. 

 vector<int> func() {
    ....
    return vec;
}

x = func();

This is where we want the move constructor to work instead of the copy constructor, thus x will get all the data of the return value, but the return value should be invalidated and not take on the value of x, Thus what was inside x is not freed and that may cause memory leakage.

arcondello: I think it's ok, I see the same pattern in https://en.cppreference.com/w/cpp/language/rule_of_three

amahmudwave: Okay got it, since you have an explicit destructor, the temporary object will be destroyed and there wont be any memory leak.

PR #888, opened 2021-07-09
Support specifying lower and upper bounds for integer variables in quadratic models

arcondello:


dimod/quadratic/quadratic_model.py
```diff
@@ -493,6 +520,11 @@ def iter_neighborhood(self, v: Variable) -> Iterator[Tuple[Variable, Bias]]:
     def iter_quadratic(self) -> Iterator[Tuple[Variable, Variable, Bias]]:
         return self.data.iter_quadratic
 
+    @forwarding_method
+    def lower_bound(self, v: Variable) -> Bias:
```
hhtong, line 524: should this be `-> Integer` instead? (same for `upper_bound`)

arcondello: Yeah, it's weird, but no. For various performance reasons we store the bounds under the hood as `bias_type`, the same as bias.

hhtong (approved):
LGTM!

PR #890, opened 2021-07-09
Remove dimod.discrete dependency on dimod.bqm

arcondello:


arcondello:
Not sure why this is failing in windows - closing for now.

PR #891, opened 2021-07-09
Deprecate AdjVectorBQM and AdjDictBQM

arcondello:


hhtong (approved):
LGTM, just a question: do we need to specify in the docstrings that these classes are being deprecated?

PR #900, opened 2021-07-15
Cythonize `QM.from_bqm`

arcondello:


arcondello:
I don't believe that the doctest failure is related.

PR #909, opened 2021-07-19
Refresh circleci

arcondello:


.circleci/config.yml
```diff
@@ -433,44 +442,45 @@ workflows:
...
+              python-version: *python-versions
             exclude:
-              - numpy-version: '1.20.2'
-                image: 3.6-jessie
-              - numpy-version: '1.18.5'
-                image: 3.9-buster
-              - numpy-version: '1.17.3'
-                image: 3.9-buster
+              - numpy-version: <1.21
+                image: 3.6.8
```
hhtong, line 466: instead of `image`, should this be `python-version` instead?

arcondello: it should. Huh. Wonder how that's working.

arcondello: Interesting, it's ignoring the `--only-binary`. I'll take a look to see what's going on.

arcondello: for posterity, needed `--only-binary=:all:` rather than `--only-binary=all`

Issue #901, opened 2021-07-16
`import dimod` fails on Python 3.7.10 with dimod 0.10.0


j-i-k-o:
**Description**
`import dimod` on Python 3.7.10 causes a `TypeError` exception.
```shell
❯ python           
Python 3.7.10 (default, Jul 16 2021, 19:59:05) 
[GCC 11.1.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import dimod
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/jiko/.pyenv/versions/3.7.10/lib/python3.7/site-packages/dimod/__init__.py", line 18, in <module>
    from dimod.bqm import *
  File "/home/jiko/.pyenv/versions/3.7.10/lib/python3.7/site-packages/dimod/bqm/__init__.py", line 15, in <module>
    from dimod.bqm.adjvectorbqm import AdjVectorBQM
  File "dimod/bqm/adjvectorbqm.pyx", line 38, in init dimod.bqm.adjvectorbqm
  File "/home/jiko/.pyenv/versions/3.7.10/lib/python3.7/site-packages/dimod/core/__init__.py", line 15, in <module>
    from dimod.core.composite import Composite, ComposedSampler
  File "/home/jiko/.pyenv/versions/3.7.10/lib/python3.7/site-packages/dimod/core/composite.py", line 67, in <module>
    from dimod.core.sampler import Sampler
  File "/home/jiko/.pyenv/versions/3.7.10/lib/python3.7/site-packages/dimod/core/sampler.py", line 110, in <module>
    from dimod.binary_quadratic_model import BinaryQuadraticModel
  File "/home/jiko/.pyenv/versions/3.7.10/lib/python3.7/site-packages/dimod/binary_quadratic_model.py", line 17, in <module>
    from dimod.binary.binary_quadratic_model import BinaryQuadraticModel, BQM
  File "/home/jiko/.pyenv/versions/3.7.10/lib/python3.7/site-packages/dimod/binary/__init__.py", line 15, in <module>
    from dimod.binary.binary_quadratic_model import *
  File "/home/jiko/.pyenv/versions/3.7.10/lib/python3.7/site-packages/dimod/binary/binary_quadratic_model.py", line 30, in <module>
    from numpy.typing import ArrayLike, DTypeLike
[...]

PR #912, opened 2021-07-21
Remove support for numpy 1.21.0 and 1.21.1 due to typing bug

hhtong:
See #901

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -69,6 +69,7 @@ setup(
         'dimod/include/',
         ],
     install_requires=[
+        # ignoring 1.21.0 and 1.21.1 due to https://github.com/dwavesystems/dimod/issues/901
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -69,7 +69,7 @@ def build_extensions(self):
         'dimod/include/',
         ],
     install_requires=[
-        'numpy>=1.17.3,<2.0.0',  # keep synced with circle-ci, pyproject.toml
+        'numpy>=1.17.3,<2.0.0,!=1.21.0,!=1.21.1' # keep synced with circle-ci, pyproject.toml
```
arcondello, line 72: might want to add a `see https://github.com/dwavesystems/dimod/issues/901` comment, otherwise this is somewhat mysterious.

hhtong: done!

arcondello (approved):
LGTM

PR #917, opened 2021-07-22
Fix/add empty bqm

wbernoudy:
Adding an empty BQM to another is causing a segfault.

```
In [1]: dimod.BQM.empty(dimod.BINARY) + dimod.BQM.empty(dimod.BINARY)
[1]    30444 segmentation fault (core dumped)  ipython
```

Seems to be caused by trying to use `std::max_element` on an empty vector https://github.com/dwavesystems/dimod/blob/c39677b4a743574dc795bc140dce703abd61087b/dimod/include/dimod/quadratic_model.h#L590

### DIFF
--- a/dimod/include/dimod/quadratic_model.h
+++ b/dimod/include/dimod/quadratic_model.h
@@ -577,6 +577,10 @@ class BinaryQuadraticModel : public QuadraticModelBase<Bias, Index> {
     template <class B, class I, class T>
     void add_bqm(const BinaryQuadraticModel<B, I>& bqm,
                  const std::vector<T>& mapping) {
+        if (bqm.num_variables() == 0)
+            // nothing to do, other BQM is empty
+            return;
+
         if (bqm.vartype() != this->vartype()) {
             // we could do this without the copy, but for now let's just do
             // it simply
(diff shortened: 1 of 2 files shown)


dimod/include/dimod/quadratic_model.h
```diff
@@ -577,6 +577,10 @@ class BinaryQuadraticModel : public QuadraticModelBase<Bias, Index> {
     template <class B, class I, class T>
     void add_bqm(const BinaryQuadraticModel<B, I>& bqm,
                  const std::vector<T>& mapping) {
+        if (bqm.num_variables() == 0)
+            // nothing to do, other BQM is empty
```
arcondello, line 581: There still might be an offset
```
this->offset() += bqm.offset();
```

wbernoudy: Yep good catch.

tests/test_bqm.py
```diff
@@ -3175,3 +3175,21 @@ def test_spin(self, name, BQM):
 
             self.assertAlmostEqual(bqm.energy(sample),
                                    sum(sample[v]*b for v, b in terms)**2)
+
+
+class TestAddBQM(unittest.TestCase):
+    @parameterized.expand(itertools.product(BQMs.items(), BQMs.items()))
```
arcondello, line 3181: Doesn't actually matter, but you don't need the `BQM_info0` if you just do `itertools.product(BQMs.values(), repeat=2)`

dimod/include/dimod/quadratic_model.h
wbernoudy: This should be fixed now. First convert the BQM if necessary, then add offset, then you're done if empty.

arcondello (approved):

arcondello:
Good catch, thanks.

PR #924, opened 2021-07-26
Make dwave-preprocessing a full dependency

hhtong:


setup.py
```diff
@@ -70,7 +70,7 @@ def build_extensions(self):
         ],
     install_requires=[
         # ignoring 1.21.0 and 1.21.1 due to https://github.com/dwavesystems/dimod/issues/901
-        'numpy>=1.17.3,<2.0.0,!=1.21.0,!=1.21.1' # keep synced with circle-ci, pyproject.toml
-        ],
-    extras_require=dict(preprocessing='dwave-preprocessing>=0.3,<0.4'),
+        'numpy>=1.17.3,<2.0.0,!=1.21.0,!=1.21.1', # keep synced with circle-ci, pyproject.toml
+        'dwave-preprocessing>=0.3,<0.4'
```
arcondello, line 74: might as well add the trailing comma to make it future-proof :smile:

setup.py
```diff
@@ -70,7 +70,7 @@ def build_extensions(self):
...
+        ]
```
arcondello, line 75: ditto on trailing comma

requirements.txt
```diff
@@ -1,5 +1,6 @@
 numpy==1.19.4
 cython==0.29.21
+dwave-preprocessing>=0.3.0,<0.4
```
arcondello, line 3: may as well pin this version

arcondello (commented):
Can you add this to the changelog with `reno`?

releasenotes/notes/preprocessing-full-dependency-2a62f62e36d5ace6.yaml
```diff
@@ -0,0 +1,5 @@
+---
```
arcondello, line 1: Think there also needs to be an `upgrade` section that mentions that we removed the `preprocessing` extra install arg

hhtong: Added! But on second thought -> is this an upgrade or a deprecation?

arcondello: Deprecations are for things that will continue to work for a while. In this case because we're just removing it, it's an upgrade (IMO).

arcondello (approved):

PR #930, opened 2021-07-29
Add a few module-level docs and other structure hints

arcondello:


JoelPasvolsky (approved):
Thank you!
One thought is whether to call more attention to the deprecation by adding it to the release notes (the readme will only be seen by someone looking at that folder in the code)?

arcondello:
It's already in the notes, when the classes were deprecated.

Issue #866, opened 2021-06-30
`CQM.from_dqm` and `CQM.from_bqm` or similar


arcondello:
It would be nice to be able to easily construct constrained quadratic models from binary and discrete quadratic models. Something like
```
cqm = CQM.from_dqm(dqm)
```

PR #936, opened 2021-07-30
Add `CQM.from_bqm`, `CQM.from_dqm` and `CQM.from_qm`

arcondello:
Closes https://github.com/dwavesystems/dimod/issues/866

### DIFF
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -24,11 +24,13 @@ import zipfile
 
 from numbers import Number
 from typing import Hashable, Optional, Union, BinaryIO, ByteString, Iterable, Collection, Dict
+from typing import Callable
 
 import numpy as np
 
 from dimod.core.bqm import BQM as BQMabc
 from dimod.binary.binary_quadratic_model import BinaryQuadraticModel, Binary, Spin, as_bqm
+from dimod.discrete.discrete_quadratic_model import DiscreteQuadraticModel
 from dimod.quadratic import QuadraticModel
 from dimod.sym import Comparison, Eq, Le, Ge, Sense
 from dimod.serialization.fileview import SpooledTemporaryFile, _BytesIO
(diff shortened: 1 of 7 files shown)


dimod/constrained.py
```diff
@@ -442,6 +444,60 @@ def add_variable(self, v: Variable, vartype: VartypeLike):
...
+                    )
+
+        objective.offset = dqm.offset
+
+        cqm.set_objective(objective)
+
+        return cqm
+
+    @classmethod
+    def from_quadratic_model(cls, qm: BinaryQuadraticModel) -> 'ConstrainedQuadraticModel':
```
hhtong, line 495:
```suggestion
    def from_quadratic_model(cls, qm: QuadraticModel) -> 'ConstrainedQuadraticModel':
```

dimod/discrete/discrete_quadratic_model.py
```diff
@@ -537,6 +537,10 @@ def from_numpy_vectors(cls, case_starts, linear_biases, quadratic,
 
         return obj
 
+    def get_cases(self, v):
```
hhtong, line 540: do we need to add this method to dqm.rst?

hhtong (approved):

dimod/discrete/discrete_quadratic_model.py
arcondello: Also the changelog I suppose

hhtong:
Also, is the plan to later add `from_bqm`, `from_dqm` and `from_qm` as aliases?

arcondello:
> Also, is the plan to later add `from_bqm`, `from_dqm` and `from_qm` as aliases?

I definitely considered it. It felt like it was adding clutter, but I am definitely open to the idea. Do you think they're helpful?

hhtong:
> > Also, is the plan to later add `from_bqm`, `from_dqm` and `from_qm` as aliases?
> 
> I definitely considered it. It felt like it was adding clutter, but I am definitely open to the idea. Do you think they're helpful?

I mean personally I prefer the shorthand. I guess it depends on how much we think these functions will be used? They seem pretty helpful to me but I agree that it would be a bit unnecessary if not used often.

randomir:
I would prefer shorthands as well. Maybe even adding something like `.from_dimod_model` that would do a single dispatch...

arcondello:
Ok, reworked, now there is `CQM.from_discrete_quadratic_model` with alias `CQM.from_dqm` and  `CQM.from_quadratic_model` with aliases `CQM.from_qm` and `CQM.from_bqm`.

dimod/constrained.py
```diff
@@ -442,6 +444,80 @@ def add_variable(self, v: Variable, vartype: VartypeLike):
...
             return self.variables._append(vartype, v)
 
+    @classmethod
+    def from_bqm(cls, bqm: BinaryQuadraticModel) -> 'ConstrainedQuadraticModel':
+        """Alias for :meth:`from_quadratic_model`."""
+        return cls.from_quadratic_model(bqm)
+
+    @classmethod
+    def from_discrete_quadratic_model(cls, dqm: DiscreteQuadraticModel, *,
+                                      relabel_func: Callable[[Variable, int], Variable] = lambda v, c: (v, c),
```
randomir, line 454: Might be a rare case, with no practical/obvious abuse scenario that comes to mind, but function objects technically are mutable (so shouldn't be used for default argument values).

dimod/constrained.py
```diff
@@ -442,6 +444,80 @@ def add_variable(self, v: Variable, vartype: VartypeLike):
         else:
             return self.variables._append(vartype, v)
 
+    @classmethod
+    def from_bqm(cls, bqm: BinaryQuadraticModel) -> 'ConstrainedQuadraticModel':
```
randomir, line 448: Do we want to add `from_binary_quadratic_model` for completeness? :shrug:

randomir (approved):
LGTM!

dimod/constrained.py
arcondello: Sure, technically. But as you say, I really cannot think of an abuse scenario is this is clear and simple.

dimod/constrained.py
arcondello: We'll see if anyone requests it.

PR #938, opened 2021-08-02
Unique variable label feature

hbarovertwo:
Uses Python standard library module `uuid` to generate unique (string) ID's for `Spin`, `Binary`, and `Integer` variables by default, thus accepting a value of `None` for the `label` argument of each.

Considerations:

1. `uuid.uuid4()` is used in this code, we could look into different numbered uuid versions as alternatives. Version 4 relies on random/pseudo random number generation, and has a very small near zero chance of collision.
2. Label default values are instances of `uuid.UUID`, can be overwritten whenever a hashable label is supplied. We could represent these as strings of the UUIDs as well.

Please comment if you have any strong feelings about the considerations listed above!
And if I missed any places where the changes should also be made 😄


dimod/binary/binary_quadratic_model.py
```diff
@@ -1823,12 +1824,12 @@ class Float64BQM(BQM, default_dtype=np.float64):
     pass
 
 
-def Binary(label: Variable, bias: Bias = 1,
+def Binary(label: Variable = uuid.uuid4(), bias: Bias = 1,
```
arcondello, line 1827: I believe this will assign the same label to each variable (determined when this module is imported).

arcondello:
It may be overkill, but we could do something like
```
try:
    from uuid import SafeUUID
except ImportError:
    # python 3.6
    from uuid import UUID as SafeUUID
```
for thread safely..

dimod/binary/binary_quadratic_model.py
hbarovertwo: I'll add this test and fix this, thanks.

dimod/decorators.py
```diff
@@ -497,3 +498,21 @@ def wrapper(obj, *args, **kwargs):
...
+
+def unique_variable_labels(f):
+    """Decorator to assign unique labels to variables when no label is passed.
+
+    Designed to be applied to variable methods, :meth:`.dimod.Binary`,
+    :meth:`.dimod.Spin`, :meth:`.dimod.Integer`.
+
+    """
+    @wraps(f)
+    def conditional_unique_label(*args, **kwargs):
```
arcondello, line 510: why not `conditional_unique_label(label=None, **kwargs)` so that it more closely matches the signature we're trying to replace.

dimod/decorators.py
```diff
@@ -497,3 +498,21 @@ def wrapper(obj, *args, **kwargs):
...
+        if len(args) == 0 and len(kwargs) == 0:
```
arcondello, line 511: why do we want `len(kwargs) == 0`? It seems that we don't because I believe `Integer(lower_bound=1)` will fail as currently implemented.

dimod/binary/binary_quadratic_model.py
```diff
@@ -1823,13 +1823,37 @@ class Float64BQM(BQM, default_dtype=np.float64):
...
 
 
-def Binary(label: Variable, bias: Bias = 1,
+@unique_variable_labels
+def Binary(label: Optional[Variable] = None, bias: Bias = 1,
            dtype: Optional[DTypeLike] = None) -> BinaryQuadraticModel:
+    """Return a :class:`.BinaryQuadraticModel` instance representing a binary variable.
+
+    Args:
+        label: Optional hashable label to identify the variable.
```
arcondello, line 1832: can we describe what happens when it's not provided? And specifically we should contrast it with `add_variable`

arcondello, line 1832:
```
label (optional): ...
```

dimod/binary/binary_quadratic_model.py
```diff
@@ -1823,13 +1823,37 @@ class Float64BQM(BQM, default_dtype=np.float64):
     pass
 
 
-def Binary(label: Variable, bias: Bias = 1,
+@unique_variable_labels
+def Binary(label: Optional[Variable] = None, bias: Bias = 1,
            dtype: Optional[DTypeLike] = None) -> BinaryQuadraticModel:
+    """Return a :class:`.BinaryQuadraticModel` instance representing a binary variable.
```
arcondello, line 1829: `Return a binary quadratic model with a single variable.`

releasenotes/notes/unique_variable_label_feature-5e890d6ebd7759a9.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    Labels for :func:`.dimod.Spin`, :func:`.dimod.Binary`, and :func:`dimod.Integer`
+    are no longer required positional arguments. A new decorator has been added to
```
arcondello, line 5: let's not mention the decorator, since we don't really expect users to use it.

dimod/binary/binary_quadratic_model.py
arcondello: We don't normally use the reference links in the first line of the docstring, sphinx doesn't really know what to do with them.

dimod/decorators.py
hbarovertwo: That is right..Was definitely looking at the signatures of `Spin` and `Binary` when writing this but it would break with those keyword only args on `Integer`. Have to rethink along with the above comment.

hbarovertwo: Reworked this. Please double check the new logic!

dimod/binary/binary_quadratic_model.py
hbarovertwo: Is this necessary in docstring if the type hint is present?

dimod/decorators.py
```diff
@@ -497,3 +498,20 @@ def wrapper(obj, *args, **kwargs):
...
+    """Decorator to assign unique labels to variables when no label is passed.
+
+    Designed to be applied to variable methods, :meth:`.dimod.Binary`,
+    :meth:`.dimod.Spin`, :meth:`.dimod.Integer`.
+
+    """
+    @wraps(f)
+    def conditional_unique_label(label = None, *args, **kwargs):
+        if label is None:
+            qm = f(label=uuid.uuid4(), *args, **kwargs)
```
arcondello, line 512: I am actually a bit surprised that Python lets you order it this way. Since for 
```
def f(label=None, *args, **kwargs): pass
```
`f(label=1, 3)` raises a 
```
SyntaxError: positional argument follows keyword argument
```
and `f(3, label=1)` raises a
```
TypeError: f() got multiple values for argument 'label'
```.

You could do 
```
def conditional_unique_label(*args, label = None, **kwargs):
    ...
```
or (my preference) just drop `*args` altogether.

File: dimod/binary/binary_quadratic_model.py (hunk 4 above)
Comment (arcondello) [reply dimod/binary/binary_quadratic_model.py:1832]: Nope, nor is it required in the text :shrug:

File: dimod/decorators.py (hunk 7 above)
Comment (arcondello) [reply dimod/decorators.py:512]: On second thought, I suppose this is fine the way it is, since it allows us to add more keyword arguments in the future while maintaining the position of `label`.

Comment (hbarovertwo) [reply dimod/decorators.py:512]: Those two errors are expected given the arguments can be passed either positionally, positionally followed by keyword (for further down the line positional args), or all as keyword args. We could consider making all of the args keyword only with defaults, but usage would be more verbose. And I don't think I can drop `*args`, because I still want the option to pass the positonal arguments that follow `label` i.e. `Spin('a',1)` should remain valid.

Comment (hbarovertwo) [reply dimod/decorators.py:512]:
```
>>> dimod.Binary('a',1) 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: conditional_unique_label() takes from 0 to 1 positional arguments but 2 were given
```
This is the issue we would see without `*args`

File: dimod/binary/binary_quadratic_model.py (hunk 4 above)
Comment (hbarovertwo) [reply dimod/binary/binary_quadratic_model.py:1832]: Fair point!

File: dimod/decorators.py (hunk 7 above)
Comment (arcondello) [reply dimod/decorators.py:512]: Ah, that is what I was missing. Ok, I stand corrected, thanks!

File: dimod/decorators.py [hunk 8]
```diff
@@ -497,3 +498,21 @@ def wrapper(obj, *args, **kwargs):
...
+
+def unique_variable_labels(f):
+    """Decorator to assign unique labels to variables when no label is passed.
+
+    Designed to be applied to variable methods, :meth:`.dimod.Binary`,
+    :meth:`.dimod.Spin`, :meth:`.dimod.Integer`.
+
+    """
+    @wraps(f)
+    def conditional_unique_label(label = None, *args, **kwargs):
```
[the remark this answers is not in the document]
hbarovertwo: It is used in two files - both for the binary quadratic model and the quadratic model.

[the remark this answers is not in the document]
arcondello: It's not in the `__all__` for this file and the doc string is fairly clear for it's usage. So I think it's probably clear from context. We could always prepend a `_` just to make that even clearer.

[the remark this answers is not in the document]
hbarovertwo: I do agree it's most likely overkill, the idea for filtering arguments with a decorator is laid out as an example in [PEP 318](https://www.python.org/dev/peps/pep-0318/)

releasenotes/notes/unique_variable_label_feature-5e890d6ebd7759a9.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
+  - |
+    Labels for :func:`.dimod.Spin`, :func:`.dimod.Binary`, and :func:`dimod.Integer`
```
arcondello, line 4: We've just been using e.g. ``` ``dimod.SPIN`` ``` rather than the sphinx syntax in the changelog.

releasenotes/notes/unique_variable_label_feature-5e890d6ebd7759a9.yaml
```diff
@@ -0,0 +1,7 @@
...
+    are no longer required positional arguments. By default, unique labels are applied
+    via the uuid module.
```
arcondello, line 6: Might as well make `uuid` a link

arcondello (approved):

PR #939, opened 2021-08-03
Add numpy

vgoliber:
Speed up combinations constraint generator by using NumPy instead of itertools

### DIFF
--- a/dimod/generators/constraints.py
+++ b/dimod/generators/constraints.py
@@ -15,6 +15,8 @@
 import itertools
 import collections.abc as abc
 
+import numpy as np
+
 from dimod.binary_quadratic_model import BinaryQuadraticModel
 from dimod.decorators import graph_argument, vartype_argument
 from dimod.vartypes import BINARY
@@ -91,9 +93,8 @@ def combinations(n, k, strength=1, vartype=BINARY):
     lbias = float(strength*(1 - 2*k))
     qbias = float(2*strength)
 
-    bqm = BinaryQuadraticModel.empty(BINARY)
-    bqm.add_variables_from(((v, lbias) for v in variables))
-    bqm.add_interactions_from(((u, v, qbias) for u, v in itertools.combinations(variables, 2)))
-    bqm.offset += strength*(k**2)
+    Q = np.triu(np.ones((n,n))*qbias, k=1)
+    np.fill_diagonal(Q, np.asarray([lbias]*n))
+    bqm = BinaryQuadraticModel(Q, 'BINARY', offset=strength*(k**2))
 
     return bqm.change_vartype(vartype, inplace=True)
--- a/dimod/generators/constraints.py
+++ b/dimod/generators/constraints.py
@@ -94,8 +94,13 @@ def combinations(n, k, strength=1, vartype=BINARY):
 
     if not isinstance(n, int):
         num_vars = len(n)
+        variables = n
     else:
         num_vars = n
+        try:
+            variables = range(n)
+        except TypeError:
+            raise TypeError('n should be a collection or an integer')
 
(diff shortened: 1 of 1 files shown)


dimod/generators/constraints.py
```diff
@@ -91,9 +92,8 @@ def combinations(n, k, strength=1, vartype=BINARY):
     lbias = float(strength*(1 - 2*k))
     qbias = float(2*strength)
 
-    bqm = BinaryQuadraticModel.empty(BINARY)
-    bqm.add_variables_from(((v, lbias) for v in variables))
-    bqm.add_interactions_from(((u, v, qbias) for u, v in itertools.combinations(variables, 2)))
-    bqm.offset += strength*(k**2)
+    Q = np.triu(np.ones((n,n))*qbias, k=1)
+    np.fill_diagonal(Q, lbias)
+    bqm = BinaryQuadraticModel(Q, 'BINARY', offset=strength*(k**2))
```
randomir, line 97: This doesn't work currently (see [here](https://github.com/dwavesystems/dimod/blob/c69abcf882b2cc78fc124d2733e2a28d02a66e46/dimod/binary/binary_quadratic_model.py#L77-L105)).

Until kwargs/offset is nicely supported, one option is to write it like this:
```
bqm = BinaryQuadraticModel({}, Q, offset, 'BINARY')
```

but much nicer is to use the qubo factory:
```suggestion
    bqm = BinaryQuadraticModel.from_qubo(Q, offset=strength*(k**2))
```

dimod/generators/constraints.py
```diff
@@ -91,9 +92,8 @@ def combinations(n, k, strength=1, vartype=BINARY):
     lbias = float(strength*(1 - 2*k))
     qbias = float(2*strength)
 
-    bqm = BinaryQuadraticModel.empty(BINARY)
-    bqm.add_variables_from(((v, lbias) for v in variables))
-    bqm.add_interactions_from(((u, v, qbias) for u, v in itertools.combinations(variables, 2)))
-    bqm.offset += strength*(k**2)
+    Q = np.triu(np.ones((n,n))*qbias, k=1)
```
randomir, line 95: `n` can also be a list or a set, not just int (see docstring above) -- in which case items define variable names (and `np.ones` above fails).

You could use `len(n)` instead of `n` above, and then relabel the bqm with:
```
bqm.relabel_variables(dict(zip(range(len(n)), n)))
```
(could also skip relabeling when user supplies int for n)

randomir (changes requested):
Nice improvement, just fix the failing tests by addressing comments I made below. Thanks!

dimod/generators/constraints.py
vgoliber: Think I updated this - let me know if more changes are needed for this

dimod/generators/constraints.py
```diff
@@ -91,9 +92,16 @@ def combinations(n, k, strength=1, vartype=BINARY):
     lbias = float(strength*(1 - 2*k))
     qbias = float(2*strength)
 
-    bqm = BinaryQuadraticModel.empty(BINARY)
-    bqm.add_variables_from(((v, lbias) for v in variables))
-    bqm.add_interactions_from(((u, v, qbias) for u, v in itertools.combinations(variables, 2)))
-    bqm.offset += strength*(k**2)
+    if not isinstance(n, int):
```
arcondello, line 95: this likely works for most cases that we care about, but we may as well be consistent with the old behaviour/exceptions. So
```
    if isinstance(n, abc.Collection):
        num_vars = len(n)
        variables = n
    else:
        num_vars = n
        try:
            variables = range(n)
        except TypeError:
            raise TypeError('n should be a collection or an integer')
```

arcondello (approved):
LGTM, thanks!

PR #940, opened 2021-08-03
Some small improvements to QM docs

arcondello:


dimod/quadratic/quadratic_model.py
```diff
@@ -342,16 +354,31 @@ def shape(self) -> Tuple[int, int]:
...
 
     @forwarding_method
     def add_linear(self, v: Variable, bias: Bias):
-        """Add a quadratic term."""
+        """Add a linear bias to an existing variable."""
         return self.data.add_linear
 
     @forwarding_method
     def add_quadratic(self, u: Variable, v: Variable, bias: Bias):
+        """Add quadratic bias to an interaction between two variables."""
```
JoelPasvolsky, line 381: "Add quadratic bias to an interaction between two variables." should be, "Add quadratic bias as an interaction between two variables" or "Add quadratic bias to a pair of variables."

dimod/quadratic/quadratic_model.py
```diff
@@ -597,6 +629,7 @@ def relabel_variables(self, mapping: Mapping[Variable, Variable],
 
     def relabel_variables_as_integers(self, inplace: bool = True
                                       ) -> Tuple['QuadraticModel', Mapping[Variable, Variable]]:
+        """Relabel the variables as `[0, n)` returning the mapping."""
```
JoelPasvolsky, line 632: "returning the mapping" --> "and return the mapping" if you want the return value described in the one-liner

JoelPasvolsky (approved):

docs/reference/quadratic.rst
```diff
@@ -0,0 +1,217 @@
+===============================
+Quadratic Models: Unconstrained
+===============================
+
+.. _qm:
```
JoelPasvolsky, line 5: For consistency: we usually put this tag at the top of the file

arcondello: I put this one and `_bqm` at the relevant section. Seems more useful that way to me?

JoelPasvolsky: We use these to reference the page; for example, 

image

And the text displayed comes from the subsequent header. We'd want that to be "Quadratic Models: Unconstrained", no?

arcondello: I think we'd still want it to say `Binary Quadratic Models`. I think I am going to leave it for now and you can change it back when you work on it :smile:

Issue #918, opened 2021-07-26
`Spins`, `Integers`, `Binaries` or similar


arcondello:
Allow the creation of multiple single-variable BQMs/QMs.
Something like
```
x, y = Binaries(['x', 'y])
```

PR #942, opened 2021-08-05
Add Spins, Binaries and Integers functions

arcondello:
Decided to go with iterators rather than lists, as a user can always make a list from a iterator, but if for whatever reason they want to do something like
```
for x in Binaries(1000):
   ...
```
we don't need to make all 1000 of them at once.

I also stuck with capitalization, to be consistent with `Binary`, `Spin`, and `Integer`, although it's starting to get a bit weird... In a followup PR I might use lower case and deprecate the capitalized ones.

Closes https://github.com/dwavesystems/dimod/issues/918

### DIFF
--- a/dimod/binary/binary_quadratic_model.py
+++ b/dimod/binary/binary_quadratic_model.py
@@ -22,7 +22,11 @@ import tempfile
 import warnings
 
 from numbers import Integral, Number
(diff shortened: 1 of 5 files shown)


randomir (approved):
LGTM.

I agree capitalized generators don't look right. But they are probably as consistent with capitalized spin/binary factories as they can be.

Issue #943, opened 2021-08-06
Adding integer constraints via list of terms results in an error.


alexzucca90:
**Description**
Cannot add the constraint for integer variables via list of terms. Why doing this? Indeed this can be easily done via symbolic addition. Issue is that for large problems adding through list of terms is less expensive.
The issue is that the bounds of the variables are not matching between the quadratic model created for the constraint and for the objective. See below for the code.

**Steps To Reproduce**
```
import dimod
obj = dimod.QuadraticModel()
obj.add_variable('INTEGER', 'i0')
obj.set_lower_bound('i0', 0)
obj.set_upper_bound('i0', 5)
cqm = dimod.CQM()
cqm.set_objective(obj)
const = [('i0', 1), (-8, )]
cqm.add_constraint(const, sense='<=')
```

**Expected Behavior**
The CQM should get a constraint `i0 <= 8`

**Environment**
 - OS: [e.g., Ubuntu 18.04.4 LTS]
 - Python version: [e.g., 3.8.0]

**Additional Context**

arcondello:
Got it, to be clear, the error you're getting is `ValueError: variable 'i0' has already been added with a different lower bound`? Definitely need to fix that error, just double checking that there isn't _another_ error :smile:

alexzucca90:
Yes, that's correct! Sorry forgot to add the error..

PR #944, opened 2021-08-09
Fix adding integer variables with bounds via `CQM.add_constraint_from_iterable`

arcondello:
Closes https://github.com/dwavesystems/dimod/issues/943


alexzucca90:
It works and looks good to me.

alexzucca90 (approved):

PR #946, opened 2021-08-10
Add quicksum function

arcondello:
There is probably still more performance on the table, but this already provides a pretty nice speedup and we can encourage people to use it and improve it more later.

Testing locally,
```python
import time
import dimod

t = time.time()
sum(dimod.Binary(v) for v in range(10000))
print(time.time() - t)

t = time.time()
dimod.quicksum(dimod.Binary(v) for v in range(10000))
print(time.time() - t)
```
returns
```bash
1.0739736557006836
0.1277778148651123 
```

### DIFF
--- a/dimod/binary/binary_quadratic_model.py
+++ b/dimod/binary/binary_quadratic_model.py
@@ -56,6 +56,7 @@ __all__ = ['BinaryQuadraticModel',
            'Float64BQM',
            'as_bqm',
            'Spin', 'Binary', 'Spins', 'Binaries',
+           'quicksum',
            ]
 
 BQM_MAGIC_PREFIX = b'DIMODBQM'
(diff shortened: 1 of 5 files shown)


dimod/binary/binary_quadratic_model.py
```diff
@@ -1980,5 +1981,29 @@ def as_bqm(*args, cls: None = None, copy: bool = False,
     return BinaryQuadraticModel(*args, dtype=dtype)
 
 
+def quicksum(iterable: Iterable[Union[BinaryQuadraticModel, QuadraticModel, Bias]]
+             ) -> Union[BinaryQuadraticModel, QuadraticModel]:
+    r"""Sum the given items.
```
JoelPasvolsky, line 1986: "items" --> Maybe iterable or models?

JoelPasvolsky (approved):

hsadeghidw (approved):
LGTM!

PR #947, opened 2021-08-10
Update doc/reference pages and add some symbolic math content

JoelPasvolsky:
Some first-draft content. Suggestions welcome


dimod/constrained.py
```diff
@@ -93,13 +93,14 @@ def vartype(self, v: Variable) -> Vartype:
 class ConstrainedQuadraticModel:
     r"""A constrained quadratic model.
 
-    Constrained quadratic models are problems of the form:
+    Constrained quadratic models represent problems of the form:
```
arcondello, line 96: I would say ```:class:`ConstrainedQuadraticModel`\s represent```, but `Constrained quadratic models are`

docs/reference/symbolic.rst
```diff
@@ -0,0 +1,65 @@
...
+.. _symbolic_math:
+
+=============
+Symbolic Math
+=============
+
+You can construct a model, for example a CQM, from symbolic math, which is especially
+useful for learning and testing with small problems.
+
+dimod supports binary and integer variables:
```
arcondello, line 11: You kind of say this below with `Such variables are constructed as either...` but IMO it's an extremely key point that these are one-variable models, not variables.

docs/reference/symbolic.rst
```diff
@@ -0,0 +1,65 @@
...
+>>> # Create binary variables
+>>> x = Binary('x')
+>>> y = Binary('y')
+>>> # Create integer variables
+>>> i = Integer('i')
+>>> j = Integer('j')
+
+Such variables are constructed as either BQMs or QMS, depending on the type of
+variable:
+
```
arcondello, line 23: I would break each of these steps out and explain exactly what's happening with text.

I would also include two more:

```
>> x
BinaryQuadraticModel({'x': 1.0}, {}, 0.0, 'BINARY')
```
emphasizes the "one-variable models, not variables" point.

```
>>> Spin('s') + Binary('x')
QuadraticModel({'s': 1.0, 'x': 1.0}, {}, -0.0, {'s': 'SPIN', 'x': 'BINARY'}, dtype='float64')
```
is important because the type of the object is being "promoted" to QM.

docs/reference/symbolic.rst
```diff
@@ -0,0 +1,65 @@
...
+>>> 2*x
+BinaryQuadraticModel({'x': 2.0}, {}, 0.0, 'BINARY')
+>>> 3*i - x
+QuadraticModel({'i': 3.0, 'x': -1.0}, {}, -0.0, {'i': 'INTEGER', 'x': 'BINARY'}, dtype='float64')
+
+You can express mathematical functions on these variables using Python functions such
+as :func:`sum`:
+
+>>> sum([3*i, 2*i])
```
arcondello, line 32: this should use `quicksum` #946 once it's merged.

arcondello:
Also the commit message for https://github.com/dwavesystems/dimod/pull/947/commits/0b6915cdb713dd04f2cbcd5aa06760255933b358 is a hilarious typo

dimod/constrained.py
JoelPasvolsky: I tend to separate the user-centric perspective in which these models represent (model) real-world user problems such as a knapsack problem from the solver-centric perspective in which these mathematical formulae are the problems submitted to a solver.
Not crucial.

arcondello: Sure, but in this case "constrained quadratic model" is the label that we are applying to that mathematical definition. It's not representing it. The class is arguably a representation because it's not an abstract mathematical definition.

docs/reference/symbolic.rst
JoelPasvolsky: Done

docs/reference/symbolic.rst
JoelPasvolsky: I want one for Python functions and then point this one out for performance

docs/reference/symbolic.rst
```diff
@@ -0,0 +1,71 @@
+
+.. _symbolic_math:
+
+=============
+Symbolic Math
+=============
+
+You can construct a model, for example a CQM, from symbolic math, which is especially
```
arcondello, line 8: do we want to write out constrained quadratic model rather than use the abbreviation here? Or alternatively maybe just say more generally `You can construct problems for quantum and hybrid solvers with symbolic math...`

docs/reference/symbolic.rst
JoelPasvolsky: Given that it is you who will be making the `quicksum` PR, it will also be you updating this as part of that PR :-)

arcondello: Bad news, it's already merged! Beat you to it

docs/reference/symbolic.rst
JoelPasvolsky: Changed to "a constrained quadratic model (CQM)," (rather than bring in solver here for now)

docs/reference/quadratic.rst
```diff
@@ -215,3 +223,81 @@ Converting to and from other data structures:
...
+Methods
+~~~~~~~
+
+.. autosummary::
+   :toctree: generated/
+
+   ~DiscreteQuadraticModel.add_linear_equality_constraint
+   ~DiscreteQuadraticModel.add_variable
+   ~DiscreteQuadraticModel.copy
+   ~DiscreteQuadraticModel.from_file
```
arcondello, line 260: While we're at it, this is missing `energy` and `energies`...

docs/reference/symbolic.rst
JoelPasvolsky: Added

docs/reference/quadratic.rst
JoelPasvolsky: Added

arcondello (approved):
I have some more thoughts, but I think this is at a point we can merge and make changes in a follow up.

PR #948, opened 2021-08-10
Add `CQM.substitute_self_loops` method

arcondello:
Also some other misc bug fixes.

### DIFF
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -649,16 +649,16 @@ class ConstrainedQuadraticModel:
     def substitute_self_loops(self) -> Dict[Variable, Variable]:
         """Replace any integer self-loops in the objective or constraints.
 
-        A self-loop :math:`i^2` is removed by introducing a new
-        variable :math:`j`, adding a constraint :math:`j == i`, and then
-        substitute :math:`i^2 == i*j`.
+        Self-loop :math:`i^2` is removed by introducing a new variable
+        :math:`j` with interaction :math:`i*j` and adding constraint
+        :math:`j == i`.
 
         Acts on the objective and constraints in-place.
 
         Returns:
             A mapping from the integer variable labels to their introduced
-            counterparts. The same label is used for the constraint that
-            enforces the relationship.
+            counterparts. The constraint enforcing :math:`j == i` uses
+            the same label.
 
         """
         mapping: Dict[Variable, Variable] = dict()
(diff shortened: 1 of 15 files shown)


dimod/constrained.py
```diff
@@ -599,6 +600,80 @@ def set_objective(self, objective: Union[BinaryQuadraticModel, QuadraticModel]):
...
+
+            qm.add_quadratic(u, new, bias)
+            qm.remove_interaction(u, u)
+
+    def substitute_self_loops(self) -> Dict[Variable, Variable]:
+        """Replace any integer self-loops in the objective or constraints.
+
+        A self-loop :math:`i^2` is removed by introducing a new
+        variable :math:`j`, adding a constraint :math:`j == i`, and then
+        substitute :math:`i^2 == i*j`.
```
JoelPasvolsky, line 654: Maybe more accurate (and without the switch in tense in the last clause): "Self-loop :math:`i^2` is removed by introducing a new variable :math:`j` with interaction :math:`i*j` and adding constraint :math:`j == i`."

dimod/constrained.py
```diff
@@ -599,6 +600,80 @@ def set_objective(self, objective: Union[BinaryQuadraticModel, QuadraticModel]):
...
+
+        Acts on the objective and constraints in-place.
+
+        Returns:
+            A mapping from the integer variable labels to their introduced
+            counterparts. The same label is used for the constraint that
+            enforces the relationship.
```
JoelPasvolsky, line 661: "The same label is used for the constraint that enforces the relationship." Unclear.

releasenotes/notes/add-QM.remove_interaction-e97a7ac91db2e2b8.yaml
```diff
@@ -0,0 +1,9 @@
+---
+features:
+  - |
+    Implement ``QuadraticModel.remove_interaction`` method, previously it
+    raised a ``NotImplementedError`` error.
```
JoelPasvolsky, line 5: If you're splicing with a comma, then "method, which previously raised"

arcondello: Too late. It's in the git history like that forever. Haunting you.

JoelPasvolsky: Well, you're stuck forever with https://github.com/dwavesystems/dimod/pull/947#issuecomment-896239578

PR #950, opened 2021-08-12
Complete BQM one-liners

JoelPasvolsky:
First draft just to complete the one liners and make some fixes, will expand in another PR.

### DIFF
--- a/dimod/binary/binary_quadratic_model.py
+++ b/dimod/binary/binary_quadratic_model.py
@@ -69,11 +69,11 @@ class BinaryQuadraticModel(QuadraticViewsMixin):
 
     .. math::
 
-    E(\bf{v})
-    = \sum_{i=1} a_i v_i
-    + \sum_{i<j} b_{i,j} v_i v_j
-    + c
-    \qquad\qquad v_i \in\{-1,+1\} \text{  or } \{0,1\}
+        E(\bf{v})
+        = \sum_{i=1} a_i v_i
+        + \sum_{i<j} b_{i,j} v_i v_j
+        + c
+        \qquad\qquad v_i \in\{-1,+1\} \text{  or } \{0,1\}
 
     where :math:`a_{i}, b_{ij}, c` are real values.
 
@@ -494,7 +494,7 @@ class BinaryQuadraticModel(QuadraticViewsMixin):
 
     @forwarding_method
     def add_linear(self, v: Variable, bias: Bias):
-        """Add a quadratic term."""
+        """Add a linear term."""
         return self.data.add_linear
 
     def add_linear_equality_constraint(
@@ -672,6 +672,12 @@ class BinaryQuadraticModel(QuadraticViewsMixin):
             self.add_linear(v, bias)
 
     def add_linear_from_array(self, linear: Sequence):
+        """Add linear biases from an array-like to a binary quadratic model.
+
+        Args:
+            linear:
+                A one-dimensional `array_like`_ of linear biases.
+        """
         ldata = np.asarray(linear)
 
         # cython has trouble with readonly buffers as of 0.29.22, in the
@@ -695,6 +701,7 @@ class BinaryQuadraticModel(QuadraticViewsMixin):
 
     @forwarding_method
     def add_quadratic(self, u: Variable, v: Variable, bias: Bias):
+        """Add a quadratic bias between two variables."""
         return self.data.add_quadratic
 
     def add_interaction(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


dimod/binary/binary_quadratic_model.py
```diff
@@ -690,16 +714,16 @@ def add_quadratic_from(self, quadratic: Union[Mapping, Iterable]):
         Args:
             quadratic:
                 Collection of interactions and their associated quadratic
-                bias. If a dict, should be of the form `{(u, v): bias, ...}`
-                where `u` and `v` are variables in the model and `bias` is
+                bias. If a dict, should be of the form ``{(u, v): bias, ...}``
```
arcondello, line 717: We should either use ``` `` ``` and ``` ` ``` or `:math:` and `:code:`.

dimod/binary/binary_quadratic_model.py
```diff
@@ -736,6 +760,16 @@ def add_quadratic_from_dense(self, quadratic: ArrayLike):
 
     @forwarding_method
     def add_variable(self, v: Optional[Variable] = None, bias: Bias = 0):
+        """Add a variable and its linear bias to a binary quadratic model.
```
arcondello, line 763: I would not say "linear bias" here

dimod/binary/binary_quadratic_model.py
```diff
@@ -736,6 +760,16 @@ def add_quadratic_from_dense(self, quadratic: ArrayLike):
...
+
+        Args:
+            v:
```
arcondello, line 766: can just put this on the same line
```
        v: Variable label. ...
```

dimod/binary/binary_quadratic_model.py
```diff
@@ -1637,9 +1694,9 @@ def to_ising(self):
         values are converted.
 
         Returns:
-            tuple: 3-tuple of form (`linear`, `quadratic`, `offset`), where
-            `linear` is a dict of linear biases, `quadratic` is a dict of
-            quadratic biases, and `offset` is a number that represents the
+            tuple: 3-tuple of form (``linear``, ``quadratic``, ``offset``), where
```
arcondello, line 1697: ``` :code:`(linear, quadratic, offset)` ``` or ` ``(linear, quadratic, offset)`` `

dimod/binary/binary_quadratic_model.py
```diff
@@ -1789,11 +1846,22 @@ def to_numpy_vectors(self, variable_order=None, *,
         return tuple(ret)
 
     def to_qubo(self) -> Tuple[Mapping[Tuple[Variable, Variable], Bias], Bias]:
+        """Convert a binary quadratic model to QUBO format.
+
+        If the binary quadratic model's vartype is not :class:`.Vartype.BINARY`,
+        values are converted.
+
+        Returns:
+            tuple: 2-tuple of form ({(u, v): bias, ...}, ``offset``), where
```
arcondello, line 1855: ``` :code:`({(u, v): bias, ...}, offset)` ``` or ` ``({(u, v): bias, ...}, offset)`` `

dimod/binary/binary_quadratic_model.py
JoelPasvolsky: Removed

dimod/binary/binary_quadratic_model.py
JoelPasvolsky: Sure

dimod/binary/binary_quadratic_model.py
JoelPasvolsky: Yes

dimod/binary/binary_quadratic_model.py
JoelPasvolsky: We made the decision a while ago to go for ``\:code:`` but we still have legacy "`" (and when people copy & paste that can sometimes spread to newer content. I convert whenever I'm working on a file.

arcondello: I may take this back. Since the [shinx docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup) say pretty clearly that we should use ` `` ` for code samples.

arcondello: I do hate it though. Looks so much worse. :shrug:

arcondello (approved):

dimod/binary/binary_quadratic_model.py
JoelPasvolsky: We'll make it part of our CSS update :-)

randomir: FWIW, backtick syntax is just plain ReST. Nothing to do with sphinx per se. `:code:` and `:math:` as sphinx extensions to the ReST syntax. I prefer the shorter form, but also think it's fine to mix and use whatever seems more appropriate in context.

JoelPasvolsky: Yes, backtick falls under inline markup. It makes sense to me to usually use it because it's shorter but occasionally it's nice to emphasis by using :code:

PR #952, opened 2021-08-13
Add CQM methods examples

JoelPasvolsky:


dimod/constrained.py
```diff
@@ -244,6 +244,12 @@ def _add_variables_from(self, model: Union[BinaryQuadraticModel, QuadraticModel]
     def add_constraint(self, data, *args, **kwargs) -> Hashable:
         """A convenience wrapper for other methods that add constraints.
 
+        Examples:
+            >>> from dimod import ConstrainedQuadraticModel, Integers
+            >>> x, y = Integers(['x', 'y'])
```
arcondello, line 249: I have tried to follow a convention of `x, y, z` for binary variables, `r, s, t` for spin variables and `i, j, k` for integer. I think it will help in the long run if we're consistent.

dimod/constrained.py
```diff
@@ -287,6 +293,12 @@ def add_constraint_from_model(self,
         Returns:
             The label of the added constraint.
 
+        Examples:
+            >>> from dimod import ConstrainedQuadraticModel, Binary
+            >>> cqm = ConstrainedQuadraticModel()
+            >>> x = Binary('x')
+            >>> cqm.add_constraint_from_model(x, '>=', 0, "Min x")   # doctest: +IGNORE_RESULT
```
arcondello, line 300: Why ignore the result? I think a key point is that it returns the constraint label for future reference. If you don't want an output, I would do
```
>>> label = cqm.add_constraint_from_model(x, '>=', 0, label="Min x")
```
to make that clear.

Applies in a bunch of places.

dimod/constrained.py
```diff
@@ -522,7 +566,18 @@ def from_discrete_quadratic_model(cls, dqm: DiscreteQuadraticModel, *,
     def from_quadratic_model(cls, qm: Union[QuadraticModel, BinaryQuadraticModel]
                              ) -> 'ConstrainedQuadraticModel':
         """Construct a constrained quadratic model from a quadratic model or
-        binary quadratic model."""
+        binary quadratic model.
+
```
arcondello, line 570: I think some text here explaining why it just sets the objective might be useful.

dimod/constrained.py
```diff
@@ -657,10 +723,19 @@ def substitute_self_loops(self) -> Dict[Variable, Variable]:
...
-            A mapping from the integer variable labels to their introduced
+            Mapping from the integer variable labels to their introduced
             counterparts. The constraint enforcing :math:`j == i` uses
             the same label.
 
+        Examples:
+            >>> from dimod import Integer, ConstrainedQuadraticModel
+            >>> i = Integer('i')
+            >>> cqm = ConstrainedQuadraticModel()
+            >>> cqm.add_constraint(i*i <=3, label="i squared")   # doctest: +IGNORE_RESULT
```
arcondello, line 734: Likewise, I think showing how to use the mapping would be useful here.

dimod/constrained.py
JoelPasvolsky: Would it not make more sense to do `label = ...` only if one is not setting the label oneself? Otherwise one gets back exactly the string one set

arcondello: Sure, though this time as a variable. I still think it's an important point to make. In any case I wouldn't ignore the output.

dimod/constrained.py
JoelPasvolsky: Updated

dimod/constrained.py
JoelPasvolsky: Done

dimod/constrained.py
JoelPasvolsky: Done. Also added a constraint to the example

dimod/constrained.py
JoelPasvolsky: Shown, just the line has a `# doctest: +IGNORE_RESULT` because the created variable label is random

dimod/constrained.py
```diff
@@ -657,10 +733,20 @@ def substitute_self_loops(self) -> Dict[Variable, Variable]:
...
             counterparts. The constraint enforcing :math:`j == i` uses
             the same label.
 
+        Examples:
+            >>> from dimod import Integer, ConstrainedQuadraticModel
+            >>> i = Integer('i')
+            >>> cqm = ConstrainedQuadraticModel()
+            >>> cqm.add_constraint(i*i <=3, label='i squared')
+            'i squared'
+            >>> cqm.substitute_self_loops()                      # doctest: +IGNORE_RESULT
```
arcondello, line 746: it's not really important, but why is this `# doctest: +IGNORE_RESULT` so far away?

JoelPasvolsky: Often the code snippet has a few of these and I like them aligned on the right. In this case there's just one so it sticks out.

Issue #949, opened 2021-08-12
Allow samples to have a superset of variables in energy calculations


arcondello:
Especially with the inclusion of CQM, it would be good to allow samples provided to `.energy()` and `.energies()` functions to contain a superset of the variables in the relevant model.

An example of why this is useful:
Given a CQM
```python
x, y, z = Binaries('xyz')

cqm = ConstrainedQuadraticModel()

cqm.set_objective(y + z)
cqm.add_constraint(3*x >= 2)
cqm.add_constraint(-2*y + x == 0)
```
and proposed solution `{'x': 1, 'y': 0, 'z': 1}`, we cannot currently check the energy of any of the individual components. If we allowed energy checks to be more loose, we could do
```python
objective_energy = cqm.objective.energy(sample)
for constraint in cqm.constraints.values():
    constraint_lhs_energy = constraint.lhs.energy(sample)
```
*Additional Considerations*
We could toggle this behavior with a keyword argument. In which case my preference would be something like `strict` defaulting to `False`.

*Additional Context*
This contradicts part of https://github.com/dwavesystems/dimod/issues/244

randomir:
Sounds intuitive. I would expect the `strict=False` behavior to work (if I was a new Ocean user).

And `strict=False` wouldn't be a backwards compat break, so :+1:

pau557:
Nice, LGTM

Issue #954, opened 2021-08-16
Add keyword argument(s) to `QM.energies` and `BQM.energies` to force stricter input checking


arcondello:
A user may wish to test whether the given sample(s):
1) Have exactly matching variables
2) Have the correct values for the given variable vartype(s)

Something like
```
bqm.energies(samples, check_vartypes=True, check_variables=True)
```
or more simply
```
bqm.energies(samples, strict=True)
```

**Additional Context**
Originally discussed as part of https://github.com/dwavesystems/dimod/issues/949
(1) is covered by https://github.com/dwavesystems/dimod/issues/244

arcondello:
Closed by #1010

PR #955, opened 2021-08-16
Support a superset of variables in BQM/QM.energies

arcondello:
Closes https://github.com/dwavesystems/dimod/issues/949

I decided not to go with the `strict` keyword argument. It occurs to me that we may want `strict` to refer to either the set of variables, the vartype, or both. For now, I am going to leave it alone and we can add that feature when it's requested (see https://github.com/dwavesystems/dimod/issues/954)


randomir (approved):
LGTM

pau557 (approved):
LGTM

PR #958, opened 2021-08-16
CQM from LP file 

alexzucca90:
Adding method that reads LP file and converts to CQM instance.


arcondello:
Tests are failing because you'd need to add `pyparsing` to the `setup.py` as a requirement.

arcondello:
You made the right fix, but we do a `--no-deps` install in the CI. You'll also need to add it [here](https://github.com/dwavesystems/dimod/blob/96b175b2859c9500300aba172e611dae29682bee/.circleci/config.yml#L121). Not obvious I know, sorry about that.

arcondello (commented):
Ok, first up I think we should do a little restructuring. IMO we should move almost all of functionality stuff into a `dimod/serialization/lp.py` file. In that file, I would put all of the contents of `parsing_utilities.py`, as well as the `cqm_model_from_parse_output` and `get_variables`.

I would then make a new function (or rework `cqm_model_from_parse_output`) called `cqm_from_lp` which takes an LP file and returns a CQM. The `CQM.from_lp_file` method should then just call that function.

arcondello (commented):
Second thing. It's better to work with the file objects rather than bytes, strings. So IMO `from_lp` should accept a file-like object (like `CQM.from_file`), and then you can later use [`grammar.parseFile`](https://pyparsing-docs.readthedocs.io/en/latest/pyparsing.html?highlight=parseString#pyparsing.ParserElement.parseFile) rather than `grammar.parseString`. It's more flexible this way and we don't need to make an extra copy in memory.

dimod/constrained.py
```diff
@@ -773,9 +774,119 @@ def vartype(self, v: Variable) -> Vartype:
...
+        return cqm_model_from_lp_file(fp, lower_bound_default, upper_bound_default)
 
 
+CQM = ConstrainedQuadraticModel
+
 # register fileview loader
 load.register(CQM_MAGIC_PREFIX, ConstrainedQuadraticModel.from_file)
+
+
+def cqm_model_from_lp_file(fp: Union[BinaryIO, ByteString],
```
arcondello, line 802: If this function is going to live here (which makes sense to avoid circular import), I would just move the contents into the method directly. No need for the separate function.

dimod/constrained.py
```diff
@@ -773,9 +774,119 @@ def vartype(self, v: Variable) -> Vartype:
...
+                           lower_bound_default: Optional[int] = 0,
+                           upper_bound_default: Optional[int] = 1000) -> "ConstrainedQuadraticModel":
```
arcondello, line 804: Is 1000 chosen for a specific reason? Otherwise I would set to `None` and use the existing defaults (as determined by dtype).

dimod/constrained.py
```diff
@@ -773,9 +774,119 @@ def vartype(self, v: Variable) -> Vartype:
...
 
 
+CQM = ConstrainedQuadraticModel
+
 # register fileview loader
 load.register(CQM_MAGIC_PREFIX, ConstrainedQuadraticModel.from_file)
+
+
+def cqm_model_from_lp_file(fp: Union[BinaryIO, ByteString],
+                           lower_bound_default: Optional[int] = 0,
```
arcondello, line 803: You only need to use `Optional` if you're setting the default value to `None`, otherwise you can just do `lower_bound_default: int = 0`.

requirements.txt
```diff
@@ -5,3 +5,9 @@ dwave-preprocessing==0.3.0
 reno==3.3.0  # for changelog
 
 setuptools>=46.4.0  # to support setup.cfg getting __version__
+
+dimod~=0.10.5
```
arcondello, line 9: Only `pyparsing` should be in here. The others are in `tests/requirements.txt`.

dimod/serialization/lp.py
```diff
@@ -0,0 +1,268 @@
...
+    if upper_bound_default is None:
+        ub_def = infinity
+    else:
+        ub_def = upper_bound_default
+
+    all_vars = set()
+    # default variable type for LP file
+    # should be continuous, even though
+    # they are not supported in CQMs
+    variables_info = defaultdict(lambda: {"type": "c", "lb": lb_def, "ub": ub_def})
```
arcondello, line 204: you might consider using a [namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple) rather than a dict. Doesn't really matter but uses a bit less memory and is a bit (IMO) cleaner to access.

dimod/serialization/lp.py
```diff
@@ -0,0 +1,268 @@
...
+        parse_output: the parse results encoding the LP file
+        lower_bound_default: the lower bound of the integer variables, in case they are not specified
+        upper_bound_default: the upper bound of the integer variables in case they are not specified
+
+    Returns:
+        a dictionary for each variable containing type, upper and lower bounds
+    """
+    # handle the default bounds
+    if lower_bound_default is None:
+        lb_def = -infinity
```
arcondello, line 191: You might just leave it as `None` to signal unspecified. That would allow later functions to use their own preferred defaults.

dimod/serialization/lp.py
```diff
@@ -0,0 +1,268 @@
...
+        if isinstance(oe, str):
+            continue
+
+        else:
+            if len(oe) == 2:
+                if oe.name != "":
+                    all_vars.add(oe.name[0])
+                else:
+                    # not supported yet.. although they can be replaced by CQM function
+                    raise ValueError("pure quadratic terms are not supported, found {}^2".format(oe.squared_name[0]))
```
arcondello, line 217: I believe this is saying we don't support `x^2` or `i^2` etc. But we actually do in the QM. Specifically, you can use the `x*x == x` trick for binary, and we do support integer squares.

dimod/serialization/lp.py
```diff
@@ -0,0 +1,268 @@
...
+                    all_vars.update([qe.name[0] for qe in c.quad_expr])
+                    all_vars.update([qe.second_var_name[0] for qe in c.quad_expr])
+                elif len(qe) == 2:
+                    # not supported yet
+                    raise ValueError("pure quadratic terms are not supported, found {}^2".format(qe.squared_name[0]))
+
+    # check the binary variables:
+    for n in parse_output.binaries:
+
+        variables_info[n]["ub"] = 1
```
arcondello, line 241: again, I think we could just allow this to be `None` and then allow the later `add_variable` method(s) to infer them from the vartype.

requirements.txt
alexzucca90: whoops.. I let pycharm do its job... badly

dimod/constrained.py
alexzucca90: gonna keep it to `None`

dimod/constrained.py
alexzucca90: not really, `None` sounds good to me

dimod/constrained.py
```diff
@@ -773,9 +774,112 @@ def vartype(self, v: Variable) -> Vartype:
...
 
+        # adding constraints
+        for c in parse_output.constraints:
+
+            cname = c.name[0]
+            csense = constraint_symbols[c.sense]
+            ccoef = c.rhs
+
+            # add the constraint as a list of terms. It is
+            c_model = []
```
arcondello, line 852: I suspect that you'll get a bit of a performance boost if you just construct the `QuadraticModel` here rather than a list of terms. Like you do with the objective.

dimod/constrained.py
```diff
@@ -773,9 +774,112 @@ def vartype(self, v: Variable) -> Vartype:
...
+        Returns:
+            The model encoded in the LP file, converted to ConstrainedQuadraticModel dimod object
+        """
+
+        grammar = make_lp_grammar()
+        parse_output = grammar.parseFile(fp)
+        variables_info = get_variables(parse_output, lower_bound_default, upper_bound_default)
+
+        cqm = ConstrainedQuadraticModel()
+        x = {}
```
arcondello, line 798: what is `x`? The list of variable names?

dimod/constrained.py
```diff
@@ -773,9 +774,112 @@ def vartype(self, v: Variable) -> Vartype:
...
+            upper_bound_default: in case upper bounds for integer are not given, this will be the default
 
-CQM = ConstrainedQuadraticModel
+        Returns:
+            The model encoded in the LP file, converted to ConstrainedQuadraticModel dimod object
+        """
+
+        grammar = make_lp_grammar()
+        parse_output = grammar.parseFile(fp)
+        variables_info = get_variables(parse_output, lower_bound_default, upper_bound_default)
```
arcondello, line 795: It looks like with relatively few changes you could make `get_variables` an iterator rather than returning a dict. Then do something like
```
for v, vartype, lb, ub in get_variables(parse_output):
    if vartype is Vartype.INTEGER:
        obj.add_variable(vartype, v, lb, ub)
    elif vartype is Vartype.BINARY:
        obj.add_variable(vartype, v)
    else:
        RuntimeError("unexpected vartype")  # or maybe ValueError if this hasn't been checked before
```

Since `obj` will then have all of the variables, you can use it's `.variables` `.lower_bound()` and `.upper_bound` as the source of truth in later steps.

dimod/constrained.py
```diff
@@ -773,9 +774,112 @@ def vartype(self, v: Variable) -> Vartype:
...
 
+    @classmethod
+    def from_lp_file(cls,
+                     fp: Union[BinaryIO, ByteString],
+                     lower_bound_default: Optional[int] = None,
+                     upper_bound_default: Optional[int] = None) -> "ConstrainedQuadraticModel":
+        """
+        Create a CQM model from a LP file.
+        Args:
+            fp: file-like object in the LP format
```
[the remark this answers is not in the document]
alexzucca90: there are many explanations that we can find online, from various solvers: Gurobi. FICO and CPLEX.. no Wikipedia page at the moment.

alexzucca90:
needed to rebase to include last changes in the `add_variable` method.

dimod/constrained.py
```diff
@@ -27,6 +27,9 @@
 from numbers import Number
 from typing import Hashable, Optional, Union, BinaryIO, ByteString, Iterable, Collection, Dict
 from typing import Callable, MutableMapping, Iterator, Tuple, Mapping, Any
+from os import PathLike
```
arcondello, line 30: looks like these imports are not used any more.

dimod/constrained.py
```diff
@@ -852,6 +856,136 @@ def vartype(self, v: Variable) -> Vartype:
...
+                     upper_bound_default: Optional[int] = None) -> "ConstrainedQuadraticModel":
+        """Create a CQM model from a LP file.
+
+        Args:
+            fp: file-like object in the LP format
+            lower_bound_default: in case lower bounds for integer are not given, this will be the default
+            upper_bound_default: in case upper bounds for integer are not given, this will be the default
+
+        Returns:
+            The model encoded in the LP file, converted to ConstrainedQuadraticModel dimod object
```
arcondello, line 872: ```The model encoded in the LP file as a :class:`ConstrainedQuadraticModel`.```

dimod/constrained.py
```diff
@@ -852,6 +856,136 @@ def vartype(self, v: Variable) -> Vartype:
...
+
+        cqm = ConstrainedQuadraticModel()
+
+        # parse and set the objective
+        obj_coefficient = 1
+        for oe in parse_output.objective:
+
+            if isinstance(oe, str):
+                if oe in obj_senses.keys():
+                    obj_coefficient = obj_senses[oe]
```
arcondello, line 886: where is `obj_senses` defined? I may be just missing it but I think this code block is broken (the unittests should probably explore this if condition).

arcondello: BTW, you can see what your unittests cover by running `coverage run -m unittest` followed by `coverage html`. Good coverage is necessary but not sufficient for good unittests :smile:

dimod/serialization/lp.py
```diff
@@ -0,0 +1,267 @@
...
+
+from pyparsing import *
+from collections import defaultdict, namedtuple
+from dimod.vartypes import Vartype
+from dimod.quadratic import QuadraticModel
+
+MINIMIZE = 1
+MAXIMIZE = -1
+
+obj_senses = {"max": MAXIMIZE, "maximum": MAXIMIZE, "maximize": MAXIMIZE,  "min": MINIMIZE, "minimum": MINIMIZE,
```
alexzucca90, line 25: @arcondello `obj_senses` defined here

arcondello: got it. thanks!

arcondello (approved):
LGTM, thanks!

releasenotes/notes/cqm-from-lp-file-152138c1892af7de.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    Added ``from_lp_file`` classmethod to ``ConstrainedQuadraticModel`` that
+    reads input LP file anc converts it to CQM instance.
```
arcondello, line 5: `and` not `anc`

PR #963, opened 2021-08-18
Fix SampleSet.to_pandas_dataframe for empty data vector

arcondello:


randomir (approved):
:+1:

arcondello:
This fails when the size of the data vector is `(n, 2)` or greater. The problem is that you cannot easily create a multiindex on the fly. Closing for now.

PR #965, opened 2021-08-20
Fix ``as_samples`` and ``parse_initial_states``

### LINKED ISSUE #861: Initial state parser modifies input states inplace
The root cause of https://github.com/dwavesystems/dwave-tabu/issues/86 is the behavior of `Initialized.parse_initial_states`, where `as_samples` is used with `copy=False` default, and then vartype converted in place, modifying the input `initial_states` -- but only partially. Raw sample data is converted, but SampleSet's vartype is not.

https://github.com/dwavesystems/dimod/blob/30272a7e47ec71988413ce8f8339d397ceff8d17/dimod/core/initialized.py#L126-L131

hhtong:
Fixes #861 and https://github.com/dwavesystems/dwave-tabu/issues/86

- Ensure that ``copy`` argument is not ignored in ``as_samples``
- Work on a copy of the input data in ``parse_initial_states`` when needed, so that input is not modified

### DIFF
--- a/dimod/core/initialized.py
+++ b/dimod/core/initialized.py
@@ -50,7 +50,7 @@ class Initialized(abc.ABC):
     def parse_initial_states(self, bqm,
                              initial_states=None,
                              initial_states_generator='random',
-                             num_reads=None, seed=None):
+                             num_reads=None, seed=None, copy_always=False):
         """Parses/generates initial states for an initialized sampler.
 
         Args:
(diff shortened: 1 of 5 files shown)


arcondello (approved):

dimod/core/initialized.py
```diff
@@ -106,28 +106,31 @@ def parse_initial_states(self, bqm,
...
             # confirm that the vartype matches and/or make it match
             if isinstance(initial_states, SampleSet):
                 initial_states_vartype = initial_states.vartype
             else:
                 # check based on values, defaulting to match the current bqm
-                initial_states_vartype = infer_vartype(initial_states_array) or bqm.vartype
+                initial_states_vartype = infer_vartype(initial_states) or bqm.vartype
+
+            # only create a copy of the initial states if vartype mismatch
+            copy = initial_states_vartype != bqm.vartype
```
randomir, line 117: So, the sampler will sometimes receive a copy, sometimes the original sampleset. It will never know if it's safe to modify it in place.

I would rather:
- always copy (bad for perf), or
- **provide an explicit control argument**, something along the lines of `copy_only_if_necessary=True` (`False` meaning copy always).

hhtong: Yes, that's true. I'm leaning towards the second option you listed, but maybe `copy_always` instead, or just `copy` (matching the `copy` arg in `as_samples`, which always copies when `copy=True` and copies only if necessary when `False`).

I'm not sure what should be the default behaviour though -> seems like it would be less confusing if the default were `copy_always/copy=True`.

randomir: I would default to the more performant option (copy only if needed), but would enable the developer to choose (sampler developer, so not a huge crowd).

hhtong: Added a `copy_always` parameter that defaults to `False`.

randomir (approved):
LGTM!

Issue #971, opened 2021-08-24
Method to get the violation of each constraint given a sampleset


arcondello:
Something like `cqm.violations() -> Mapping[Hashable, float]`. Can be done now with
```
violations = {}
for label, constraint in cqm.constraints.items():
    violations[label] = constraint.lhs.energy(sampleset) - constraint.rhs
```

hsadeghidw:
When we solve a satisfiability problem, we can we will have objective or energy = 0, and there is no distinction between different solutions. It would be nice to be able to easily assess which solution violates less.

I think there should be two functions. `cqm.violations()` and `cqm.constraint_energies()` or something like that.

Here is the difference. 

for ==, the violation is the absolute of energy (or square of)
for <=, the violation is zero, if energy <= 0, otherwise, it's equal.

arcondello:
The latter seems like the right "default" definition. We could potentially combine them into one method with something like
```python
class CQM:
    ...
    def iter_violations(self, clip=True) -> Iterator[Tuple[Hashable, float]]:
        if clip:
            yield from ((label, max(en, 0)) for label, en in self.iter_violations(clip=False))
        else:
            for label, constraint in cqm.constraints.items():
                yield label, constraint.lhs.energy(sampleset) - constraint.rhs


    def violations(self, clip=True) -> Dict[Hashable, float]:
        return dict(self.iter_violations(clip=clip))

```

arcondello:
Trying that again:
```python
class CQM:
    ...
    def iter_violations(self) -> Iterator[Tuple[Hashable, float]]:
        for label, constraint in cqm.constraints.items():
            if constraint.sense is Sense.Le:
                yield label, constraint.lhs.energy(sampleset) - constraint.rhs
            elif constraint.sense is Sense.Ge:
                yield label, -min(constraint.lhs.energy(sampleset) - constraint.rhs)
            elif constraint.sense is Sense.Eq:
                yield label, abs(constraint.lhs.energy(sampleset) - constraint.rhs)
            else:
                raise RuntimeError("unexpected Sense")


    def violations(self) -> Dict[Hashable, float]:
        return dict(self.iter_violations(clip=clip))
```

hsadeghidw:
Maybe I misunderstood the `clip` argument or maybe that's the wrong argument. I was expecting to see negative values for equality constraint when `clip=False`

```import dimod
import itertools
cqm = dimod.ConstrainedQuadraticModel()
x = dimod.Binary('x')
[...]

PR #975, opened 2021-08-30
Add CQM.violations and CQM.iter_violations methods

arcondello:
Closes https://github.com/dwavesystems/dimod/issues/971

### DIFF
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -619,13 +619,14 @@ class ConstrainedQuadraticModel:
 
         return cqm
 
-    def iter_violations(self, sample_like, *, skip_satisfied: bool = False,
+    def iter_violations(self, sample_like, *, skip_satisfied: bool = False, clip: bool = True,
                         ) -> Iterator[Tuple[Hashable, Bias]]:
         """Yield constraint labels and the amount the constraints are violated.
 
         Args:
             sample_like: A sample over the CQM variables.
             skip_satisfied: If True, samples that are not violated are skipped.
+            clip: If False, then negative violations will be returned.
 
         Yields:
             A 2-tuple containing the constraint label and the amount of
@@ -669,6 +670,19 @@ class ConstrainedQuadraticModel:
             greater equal 3.0
 
         """
+        if skip_satisfied:
+            # clip doesn't matter in this case
+            for label, violation in self.iter_violations(sample_like, skip_satisfied=False):
+                if violation > 0:
+                    yield label, violation
+            return
+
+        if clip:
+            for label, violation in self.iter_violations(sample_like,
+                                                         clip=False, skip_satisfied=False):
+                yield label, max(violation, 0.)
+            return
+
         sample, labels = as_samples(sample_like)
 
         if sample.shape[0] != 1:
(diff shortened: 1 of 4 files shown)


hsadeghidw (changes requested):
Make sure clip=False returns the actual energy/evaluation of both equality and inequality constraints.

dimod/constrained.py
```diff
@@ -618,6 +619,88 @@ def from_file(cls, fp: Union[BinaryIO, ByteString]) -> "ConstrainedQuadraticMode
 
         return cqm
 
+    def iter_violations(self, sample_like, *, skip_satisfied: bool = False, clip: bool = True,
+                        ) -> Iterator[Tuple[Hashable, Bias]]:
+        """Yield constraint labels and the amount the constraints are violated.
```
JoelPasvolsky, line 624: "amount the constraints are violated" --> can we rephrase this? Something like perhaps, "Yield constraint labels and the size of any violations" to be colloquial or else "Yield constraint labels and the amplitude of any violations".
BTW, for the one liner I don't think you really need "label", this could just be "Yield violations for all constraints" and then more details in the subsequent paragraph.

dimod/constrained.py
```diff
@@ -618,6 +619,88 @@ def from_file(cls, fp: Union[BinaryIO, ByteString]) -> "ConstrainedQuadraticMode
...
+
+        Args:
+            sample_like: A sample over the CQM variables.
```
JoelPasvolsky, line 627: Why does this one specify "over the CQM variables"? Typically we'd just say "A raw sample. sample_like is an extension of NumPy’s array_like structure. See as_samples()."

dimod/constrained.py
```diff
@@ -618,6 +619,88 @@ def from_file(cls, fp: Union[BinaryIO, ByteString]) -> "ConstrainedQuadraticMode
...
+            skip_satisfied: If True, samples that are not violated are skipped.
```
JoelPasvolsky, line 628: "samples that are not violated are skipped" --> "If True, does not yield constraints that are satisfied"

dimod/constrained.py
```diff
@@ -618,6 +619,88 @@ def from_file(cls, fp: Union[BinaryIO, ByteString]) -> "ConstrainedQuadraticMode
...
+            clip: If False, then negative violations will be returned.
```
JoelPasvolsky, line 629: "If False, also returns negative violations."

dimod/constrained.py
```diff
@@ -618,6 +619,88 @@ def from_file(cls, fp: Union[BinaryIO, ByteString]) -> "ConstrainedQuadraticMode
...
+
+        Yields:
+            A 2-tuple containing the constraint label and the amount of
+            constraints violation.
```
JoelPasvolsky, line 633: See above. One problem with this phrasing is it is confusing whether what is returned is the number of violations or the amplitude of each violation.

dimod/constrained.py
```diff
@@ -618,6 +619,88 @@ def from_file(cls, fp: Union[BinaryIO, ByteString]) -> "ConstrainedQuadraticMode
...
+            >>> i, j, k = dimod.Binaries(['i', 'j', 'k'])
+            >>> cqm = dimod.ConstrainedQuadraticModel()
+            >>> cqm.add_constraint(i + j + k == 10, label='equal')
+            'equal'
+            >>> cqm.add_constraint(i + j <= 15, label='less equal')
+            'less equal'
+            >>> cqm.add_constraint(j - k >= 0, label='greater equal')
+            'greater equal'
+
+            Check the violations of a satisfying sample.
```
JoelPasvolsky, line 648: "of a satisfying sample" --> "of a sample that satisfies all constraints" (all our samples are satisfying)

dimod/constrained.py
```diff
@@ -618,6 +619,88 @@ def from_file(cls, fp: Union[BinaryIO, ByteString]) -> "ConstrainedQuadraticMode
...
+
+        Example:
+
+            Construct a constrained quadratic model.
+
+            >>> i, j, k = dimod.Binaries(['i', 'j', 'k'])
+            >>> cqm = dimod.ConstrainedQuadraticModel()
+            >>> cqm.add_constraint(i + j + k == 10, label='equal')
+            'equal'
+            >>> cqm.add_constraint(i + j <= 15, label='less equal')
```
JoelPasvolsky, line 643: "label='less equal'" can we give a nicer label? Perhaps "less or equal to 15"? Makes the printed yield clearer below

PR #981, opened 2021-09-03
Backport dependence on dwave-preprocessing

randomir:
This is to enable dwave-system (>1.7.0) to support both dimod 0.9 and
0.10.

Although now we have duplicated composites (available in both dimod
and dwave.preprocessing namespaces), this is necessary to resolve
unsatisfiable dependencies when dwave-system 1.7.x is installed with
dimod 0.9.x (due to dwave-preprocessing introduced as extras dependency
in dimod 0.10.0 and then removed in 0.10.5). This makes specifying dimod
as a dependency in dwave-system impossible, if we want to allow both 0.9
and 0.10.

See also #829, #894, #896, and #924.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,6 +118,7 @@ jobs:
             . env/bin/activate
             pip install -r tests/requirements.txt
             pip install numpy=='<< parameters.numpy-version >>'
+            pip install 'dwave-preprocessing~=0.2.0' --no-deps
             pip install dimod --no-index -f dist/ --no-deps --force-reinstall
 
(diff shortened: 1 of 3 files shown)


randomir:
We could also backport #829, #894, #896, and #924 to actually replace composites in dimod with dwave-preprocessing, but that might be an overkill for this short-term release (0.9.16).

Also, py35 tests are failing because dwave-preprocessing is py36+... So I'm thinking we could drop py35 in 0.9.16?

arcondello (approved):
Makes sense to me

Issue #988, opened 2021-09-06
Symbolic higher order binary polynomials


mhlr:
**Application**
Currently dimod has a BinaryPolynomial class that can be constructed from higher order mappings like {(1,2,3): 1}
which can be passed to `make_quadratic` to create BQMs. We also have symbolic experssions for constructing BQMs.
However it is not possible to construct higher order BinomialPolynomials symbolically
```
dimod.Binary(1)*dimod.Binary(2)*dimod.Binary(3)
```
throws
```
TypeError                                 Traceback (most recent call last)
/tmp/ipykernel_1301/2702588800.py in <module>
----> 1 dimod.Binary(1)*dimod.Binary(2)*dimod.Binary(3)

~/anaconda3/lib/python3.8/site-packages/dimod/binary/binary_quadratic_model.py in __mul__(self, other)
    258         if isinstance(other, BinaryQuadraticModel):
    259             if not (self.is_linear() and other.is_linear()):
--> 260                 raise TypeError(
    261                     "cannot multiply BQMs with interactions")
    262             elif other.num_variables and other.vartype != self.vartype:

TypeError: cannot multiply BQMs with interactions
```
**Proposed Solution**
Extend symbolic expressions to BQMs with interactions and BinaryPolynomials:

BQM * BQM --> BinaryPolynomial,    if either argument contains interactions
BQM <binop> BinaryPolynomial --> BinaryPolynomial
BinaryPolynomial <binop> BQM --> BinaryPolynomial
BinaryPolynomial <binop> BinaryPolynomial --> BinaryPolynomial

Also support non-negative integer powers of BQMs and BinaryPolynomials (see https://github.com/dwavesystems/dimod/issues/864)

arcondello:
I love this suggestion in principal but there are a few things that we need to do before we should allow this:
1) We need a performant implementation of `BinaryPolynomial`. Right not it's pure Python and not all that performant an implementation at that. In practice that means that we would see a big performance cliff as soon as the degree rises above two. 2) You cannot use a graph structure to "natively" encode a binary polynomial, so a lot of the core complexity assumptions start to fall apart. There are some encoding you can do (like a bipartite graph from terms to variables) but we couldn't re-use a lot of the existing infrastructure.
3) We need the implementation of `BinaryPolynomial` to be a closer match to `QuadraticModel` and `BinaryQuadraticModel` in terms of API. `QM` and `BQM` are deliberately similar in their methods/access, but `BinaryPolynomial` less so.
[...]

PR #991, opened 2021-09-09
refactor & optimize BinaryPolynomial.__eq__

mhlr:
An example of what I have in mind for refactoring BinaryPolynomial as discussed with @arcondello  in #988
With https://toolz.readthedocs.io/en/latest/api.html#toolz.functoolz.excepts, we could include the initial part of BinaryPolynomial  #.__eq__in this treatment

### DIFF
--- a/dimod/higherorder/polynomial.py
+++ b/dimod/higherorder/polynomial.py
@@ -128,16 +128,20 @@ class BinaryPolynomial(abc.MutableMapping):
                 # not a polynomial
                 return False
 
-        if self.vartype != other.vartype:
-            return False
+        self_terms = self._terms
+        other_terms = other._terms
 
-        for term, bias in self.items():
-            if bias and other[term] != bias:
-                return False
-        for term, bias in other.items():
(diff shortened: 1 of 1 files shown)


dimod/higherorder/polynomial.py
```diff
@@ -128,16 +128,20 @@ def __eq__(self, other):
...
-        for term, bias in other.items():
-            if bias and self[term] != bias:
-                return False
-        return True
+        self_terms = self._terms
+        other_terms = other._terms
+
+        return (
+            self.vartype == other.vartype
+            and all(
```
arcondello, line 136: could fail even faster with a quick `len(terms)` check. E.g. https://github.com/dwavesystems/dimod/blob/252dca32105186a08bd83883bfc6cdc27aeacd55/dimod/binary/binary_quadratic_model.py#L1223

arcondello (approved):

mhlr:
@arcondello  I see you approved this, but i do not have write privileges to merge :)

PR #992, opened 2021-09-10
Refactoring & optimizing large BinaryPolynomial construction

mhlr:
Continuing #988 explorations.
This produces a >6x speedup in the construction of large BinaryPolynomial instances, see [gist](https://gist.github.com/mhlr/6990be27d0e36699d672353e16bfcaef)

### DIFF
--- a/dimod/higherorder/polynomial.py
+++ b/dimod/higherorder/polynomial.py
@@ -16,10 +16,13 @@ from __future__ import division
 
 import itertools
 import collections.abc as abc
+from collections import defaultdict
 
 from numbers import Number
 
 import numpy as np
+
+import cytoolz as tl
 
 from dimod.decorators import vartype_argument
 from dimod.sampleset import as_samples
@@ -33,6 +36,25 @@ def asfrozenset(term):
     """Convert to frozenset if it is not already"""
     return term if isinstance(term, frozenset) else frozenset(term)
 
+
+is_odd = tl.flip(tl.curried.operator.mod)(2)
+second_is_odd = tl.compose(is_odd, tl.second)
+
+def freeze_term(term, vartype):
+    return (
+        term
+        if isinstance(term, frozenset)
+        else (
+                frozenset(term)
+                if vartype is Vartype.BINARY
+                else frozenset(
+                        map(
+                            tl.first,
+                            filter(second_is_odd, tl.frequencies(term).items())
+                        )
+                )
+        )
+    )
 
 class BinaryPolynomial(abc.MutableMapping):
     """A polynomial with binary variables and real-valued coefficients.
@@ -92,26 +114,11 @@ class BinaryPolynomial(abc.MutableMapping):
             poly = poly.items()
 
         # we need to aggregate the repeated terms
-        self._terms = terms = {}
+
+        self._terms = terms = defaultdict(int)
         for term, bias in poly:
-
-            fsterm = asfrozenset(term)
-
-            # when SPIN-valued, s^2 == 1, so we need to handle that case
-            # in BINARY, x^2 == x
-            if len(fsterm) < len(term) and vartype is Vartype.SPIN:
-                new = set()
-                term = tuple(term)  # make sure it has .count
-                for v in fsterm:
-                    if term.count(v) % 2:
-                        new.add(v)
-                fsterm = frozenset(new)
-
-            if fsterm in terms:
-                terms[fsterm] += bias
-            else:
-                terms[fsterm] = bias
-
+            terms[freeze_term(term, vartype)] += bias
+        
         self.vartype = vartype
 
     def __contains__(self, term):
@@ -127,10 +134,10 @@ class BinaryPolynomial(abc.MutableMapping):
             except Exception:
                 # not a polynomial
                 return False
-
+        
         self_terms = self._terms
         other_terms = other._terms
-
+        
         return (
             self.vartype == other.vartype
             and all(
--- a/polynomial-perf.ipynb
+++ b/polynomial-perf.ipynb
@@ -1,0 +1,469 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "182a16c4-4777-4ec3-99b1-476914c2ec1b",
+   "metadata": {},
+   "source": [
+    "# Current"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "11b1cb71-4f1b-441f-a6bf-93d430ac5356",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "## Create test example"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "61901771-5a93-4cc9-a180-0cc2cc580a53",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%load_ext autoreload\n",
+    "%autoreload 2"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "id": "b0d92f0d-3549-4619-a272-a324ef475104",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import string, random, toolz, pandas, importlib, sys, os"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "id": "d2c51a56-7617-427e-8ec5-12e4a0fc0a08",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "terms = [(term, 1) for term in str(random.choices(string.ascii_letters, k=10**7)).split('Z')]"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "id": "1f8a40cc-df42-4f45-9220-afd1f212239f",
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "count    191557.000000\n",
+       "mean        260.018919\n",
+       "std         258.235286\n",
+       "min           4.000000\n",
+       "25%          74.000000\n",
+       "50%         179.000000\n",
+       "75%         359.000000\n",
+       "max        3084.000000\n",
+       "dtype: float64"
+      ]
+     },
+     "execution_count": 4,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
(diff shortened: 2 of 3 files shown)


dimod/higherorder/polynomial.py
```diff
@@ -16,11 +16,14 @@
 
 import itertools
 import collections.abc as abc
+from collections import defaultdict
 
 from numbers import Number
 
 import numpy as np
 
+import cytoolz as tl
```
arcondello, line 25: while `cytoolz` is a cool package, I don't think we can add it as a core ocean dependency. It seems supported-ish, but, for instance, [it only started releasing wheels in august/september](https://pypi.org/project/cytoolz/0.11.0/#files).

dimod/higherorder/polynomial.py
```diff
@@ -127,10 +134,10 @@ def __eq__(self, other):
             except Exception:
                 # not a polynomial
                 return False
-
+
```
arcondello, line 137: not sure why these lines needed changes?

dimod/higherorder/polynomial.py
mhlr: Fair enough. They have been conda since 2016 though. Many science/ai/ml package are conda only/first.

I am also using these PRs to explore ideas even if they do get merged. Hope that is ok

dimod/higherorder/polynomial.py
mhlr: Looks like an eccidental whitespace change. Will fix. I am not seeing any semantic changes

mhlr: Looks like an eccidental whitespace change. Will fix. I am not seeing any semantic changes

mhlr: Do we normally have whitespace in blank lines up to indentation level? Seems like that is what I removed

arcondello: No, in fact removing the whitespace is correct. It's more about keeping the blame/commit history clean. No big deal either way.

dimod/higherorder/polynomial.py
arcondello: More than ok! It's much easier to have these conversations on a PR.

For conda, we actually have an issue about this https://github.com/dwavesystems/dwave-ocean-sdk/issues/144. I do think conda support in addition would be nice, but we also need to support pypi installs.

mhlr: Great! I would love to have dwave code in conda. Outside dwave I use conda almost exclusively. One adwantage is that the offical repos are curated so the quality tends to be higher than pypi. Also the conda (& now also mamba) package manager maintains global environment consistecy so one install does not break previously installed packages. Conda will (up/down) grade previouly installed packages if necessary. Conda is also language agnostic so you can distribute C, R, Julia, Rust packages with it. Main drawback is it can not install directly from git.

The reason I brought cond up though was more that cytoolz is more active/mature/maintained than it's pypi history would suggest. I believe dask was also long time in conda before it got released on pypi.

mhlr: Also conda gives you MKL compiled numerical libraries by default, which in turn means that native vectorised numpy/scipy operation run multithreaded by default (without GIL)

mhlr: The original pure python pytoolz package has a [PyPi release history](https://pypi.org/project/toolz/#history) going back to 2013.
It is a drop in replacement for cytoolz (or rather the other way around) and for this use case it is less than 2x slower than cytoolz so we still get >5x instead of >8x speedup and retains the nice code. At the moment this PR only uses tl.frequencies,
the other uses were replaced by generator expressions.

mhlr: It also looks like cytool zhas been in PiPy since 2014 https://pypi.org/project/cytoolz/#history****

arcondello: Yes, though they only released an sdist until september. Which means that users without python-dev installed would not be able to install Ocean.
We have to be extremely careful about introducing new dependencies, especially ones with compiled code. Because they need to work across all of our supported OS/python versions.

dimod/higherorder/polynomial.py
```diff
@@ -34,6 +37,22 @@ def asfrozenset(term):
...
+def freeze_term(term, vartype):
+    return (
+        term
+        if isinstance(term, frozenset)
+        else (
+                frozenset(term)
+                if vartype is Vartype.BINARY
+                else frozenset(
+                        var
+                        for var, power in tl.frequencies(term).items()
```
arcondello, line 49: Seems like a big dependency to bring in for a [4 line function](https://github.com/pytoolz/toolz/blob/e9bc1e150b7b1ea1bec300a55382ed564b215f41/toolz/itertoolz.py#L548).

arcondello: I also suspect that in typical use cases, i.e. relatively low degree terms, that just doing `tuple.count()` will be faster than constructing a new `defaultdict`.

arcondello:
```python
In [7]: term = tuple('abcde')

In [8]: %timeit for _ in frequencies(term): pass
1.31 µs ± 9.02 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [9]: %timeit for _ in ((v, term.count(v)) for v in term): pass
979 ns ± 7.79 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

arcondello: I also wonder if [Counter](https://docs.python.org/3/library/collections.html#collections.Counter) would do the same thing.

mhlr: That result seems to be from the interpreted `toolz` version of frequencies. The compiled cytools version is >2x faster than the `.count` version. The count version need to iterate over `set(terms)` rather than term which slows it down more, making it closer to 3x slower than compiled cytoolz.
Counter is around the same speed as the interpreted version of `frequencies`.

```python
%timeit for _ in toolz.frequencies(term).items(): pass
    1.16 µs ± 23.5 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

%timeit for _ in ((v, term.count(v)) for v in term): pass
    826 ns ± 46.9 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

%timeit for _ in cytoolz.frequencies(term).items(): pass
    347 ns ± 10.5 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

%timeit for _ in ((v, term.count(v)) for v in set(term)): pass
    1.01 µs ± 49.5 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

%timeit for _ in ((v, count) for v, count in Counter(term).items()): pass
    1.46 µs ± 15.2 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

%timeit for _ in Counter(term).items(): pass
    1.08 µs ± 20.2 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

mhlr: > 
> 
> Seems like a big dependency to bring in for a [4 line function](https://github.com/pytoolz/toolz/blob/e9bc1e150b7b1ea1bec300a55382ed564b215f41/toolz/itertoolz.py#L548).

Cytoolz offers 2 benefits: clean code and speed.
The speed it brings is often sufficient to avoid needing to cythonize or port code to C leading to further code quality benefits as well
Depending on it brings in a moderately large number of small utility functions only a small number of which is likely to be used in any given PR. In each case it is easy to either reimplement an equivalent function or find a work around. In the long run though that means either having most of the library reimplemented possibly multiple times throughout our codebase and/or a lot of workarounds. Individual reimplementations are unlikely to get cythonized, so the codebase misses out on the performance benefit, in some cases leading to unnecessary C ports.

This could by demonstrated by refactoring a large piece of  code that is performance bottleneck. I will do more refactoring and optimization in this file for now to see how many more cytoolz uses turn up. So far i have been doing it method by method to keep the PRs small but a combine PR would demonstrate the cumulative effect

dimod/higherorder/polynomial.py
mhlr: Ok, I did not realize that the wheel availability was the issue. I thought the main concern was general maturity & support. If they are releasing wheels now what is the criterion for acceptance? Some period of time or some number of wheel releases?

arcondello: There isn't a hard algorithm. But experience over the last few years has taught us to be extremely cautious when introducing new dependencies to Ocean. Especially compiled dependencies. A significant fraction of our support requests come from dependency management. Another place this comes up, each time Python releases a new version we can only release a supporting version of Ocean once all of our dependencies also update. So for some packages like NumPy, they are so fundamental that we cannot possibly avoid using them. But otherwise we definitely err, pretty far, on keeping our dependency tree as small as possible.

mhlr:
I am thinking about making all the operations keep the polynomials zero free

arcondello:
> I am thinking about making all the operations keep the polynomials zero free

That would be a backwards compatibility break. I am not opposed philosophically, but it would need to wait until dimod 0.11.

dimod/higherorder/polynomial.py
mhlr: Fair enough. I do think that (cy)toolz is the kind universal nuts&bolts library linke numpy, scipy & pandas that is worth considering since it can be used everywhere and can reduces the need for resorting to C & cython

mhlr:
Fair enough. I think that could significantly improve scalability, both memory and time.

mhlr:
> 
> 
> > I am thinking about making all the operations keep the polynomials zero free
> 
> That would be a backwards compatibility break. I am not opposed philosophically, but it would need to wait until dimod 0.11.

Does anything really rely on the 0 weighted terms being stored explicitly?

arcondello:
Strictly speaking
```python
In [3]: p = dimod.BinaryPolynomial({}, 'BINARY')

In [4]: p['abc'] = 0

In [5]: p
Out[5]: BinaryPolynomial({frozenset({'b', 'c', 'a'}): 0}, 'BINARY')

In [6]: list(p)
Out[6]: [frozenset({'a', 'b', 'c'})]
```

whether that "break" actually matters to anyone? :shrug: 

FWIW, it is consistent with the BQM/QM behavior where we do explicitly keep 0 bias variables/interactions.

mhlr:
Should we make a backlog entry for 0.11?

mhlr:
@arcondello The last set of commits are the make_quadratic refactoring

PR #993, opened 2021-09-10
Generators refresh

arcondello:
Refresh the generators docs and some of the docstrings/typo annotations.

### DIFF
--- a/dimod/generators/anti_crossing.py
+++ b/dimod/generators/anti_crossing.py
@@ -19,12 +19,12 @@ __all__ = ['anti_crossing_clique', 'anti_crossing_loops']
 
 
 def anti_crossing_clique(num_variables: int) -> BinaryQuadraticModel:
-    """Generate an anti crossing problem with a single clique.
+    """Generate an anti-crossing problem with a single clique.
 
-    Let ``N = num_variables // 2``. This function returns a binary
-    quadratic model with variables `[0, N)` forming a ferromagnetic clique.
-    Each variable `v` in `[N, 2*N)` ferromagnetically interacts with
-    variable `v-N`.
+    Let ``N = num_variables // 2``. This function returns a binary quadratic
+    model where half the variables, `[0, N)`, form a ferromagnetic clique, with
+    each variable, `v`, also ferromagnetically interacting with one variable,
+    `v+N`, of the remaining half of the variables, `[N, 2*N)`.
 
     All of the variables in the clique except variable `1` have a linear bias
     of `+1`, and all of the variables attached to the clique have a linear bias
@@ -63,9 +63,9 @@ def anti_crossing_clique(num_variables: int) -> BinaryQuadraticModel:
 
 
 def anti_crossing_loops(num_variables: int) -> BinaryQuadraticModel:
-    """Generate an anti crossing problem with two loops.
+    """Generate an anti-crossing problem with two loops.
 
-    These instances are copies of the instance studied in [DJA]_.
+    This is the problem studied in [DJA]_.
 
     Note that for small values of ``num_variables``, the loops can be as small
     as a single edge.
--- a/dimod/generators/anti_crossing.py
+++ b/dimod/generators/anti_crossing.py
@@ -12,74 +12,87 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
-from dimod import BinaryQuadraticModel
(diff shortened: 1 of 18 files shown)


dimod/generators/anti_crossing.py
```diff
@@ -12,89 +12,102 @@
...
+from dimod.binary.binary_quadratic_model import BinaryQuadraticModel
+from dimod.vartypes import Vartype
 
 __all__ = ['anti_crossing_clique', 'anti_crossing_loops']
 
 
-def anti_crossing_clique(num_variables):
-    """Anti crossing problems with a single clique.
+def anti_crossing_clique(num_variables: int) -> BinaryQuadraticModel:
+    """Generate an anti crossing problem with a single clique.
```
JoelPasvolsky, line 22: I find this description misleading, it creates an impression that the entire problem is a clique. Perhaps something like, "Generate an anti-crossing problem of a clique with attached nodes." The single liners doesn't fully describe the problem, of course, but it doesn't misdescribe it.

dimod/generators/anti_crossing.py
```diff
@@ -12,89 +12,102 @@
...
 
-    Given the number of variables, the code will generate a clique of size
-    num_variables/2, each variable ferromagnetically interacting with a partner 
-    variable with opposite bias. A single variable in the cluster will have 
-    no bias applied.
+    Let ``N = num_variables // 2``. This function returns a binary
+    quadratic model with variables `[0, N)` forming a ferromagnetic clique.
```
JoelPasvolsky, line 25: On first read it seems that this consists of the entire BQM. Perhaps, "This function returns a binary quadratic model where half the variables, ``[0, N)``,  form a ferromagnetic clique, with each variable, ``v``, also ferromagnetically interacting with one variable, ``v+N``, of the remaining half of the variables, ``[N, 2*N)``."

dimod/generators/anti_crossing.py
```diff
@@ -12,89 +12,102 @@
...
-    
+
     return bqm
 
-def anti_crossing_loops(num_variables):
-    """Anti crossing problems with two loops. These instances are copies of the
-    instance studied in [DJA]_.
+
+def anti_crossing_loops(num_variables: int) -> BinaryQuadraticModel:
+    """Generate an anti crossing problem with two loops.
```
JoelPasvolsky, line 66: Generate an anti-crossing problem with two loops

dimod/generators/anti_crossing.py
```diff
@@ -12,89 +12,102 @@
...
+
+    These instances are copies of the instance studied in [DJA]_.
```
JoelPasvolsky, line 68: Why "instances" with its SW connotations? Maybe, "This is the problem studied ...."?

docs/reference/generators.rst
```diff
@@ -0,0 +1,61 @@
...
+===================================
+Generators and Application Modeling
+===================================
+
+.. currentmodule:: dimod.generators
+
+.. automodule:: dimod.generators
+
+Benchmarking
+============
```
JoelPasvolsky, line 12: Physics maybe?

docs/reference/generators.rst
```diff
@@ -0,0 +1,61 @@
...
+   :toctree: generated/
+
+   anti_crossing_clique
+   anti_crossing_loops
+   chimera_anticluster
+   doped
+   frustrated_loop
+
+Constraints
+===========
```
JoelPasvolsky, line 24: Consider having a category for Boolean Gates, which is useful for many problems

JoelPasvolsky (approved):

docs/reference/generators.rst
arcondello: Trying to strike a balance between the number of categories and the specificity. IMO we overestimate the value of the gate problems because they are used so much in some of our intro examples. I don't see them very much in any practical applications.

docs/reference/generators.rst
arcondello: They are not really physics applications, rather than benchmarking problems developed by physicists. Put another way,tThey use physics language, but their purpose is benchmarking.

dimod/generators/anti_crossing.py
arcondello: The problem is really with the function name IMO, but we are locked in. Blame @conta877 :laughing:

conta877: "from cliques"

PR #994, opened 2021-09-11
fix BinaryPolynomial.__eq__ when args contain different terms

mhlr:
Just noticed this. It was in the original version too, but it stands out more in the new version

### DIFF
--- a/dimod/higherorder/polynomial.py
+++ b/dimod/higherorder/polynomial.py
@@ -134,11 +134,11 @@ class BinaryPolynomial(abc.MutableMapping):
         return (
             self.vartype == other.vartype
             and all(
-                (not bias or other_terms.get(term) == bias)
+                (not bias or other_terms.get(term, 0.) == bias)
                 for term, bias in self.items()
             )
             and all(
-                (not bias or self_terms.get(term) == bias)
+                (not bias or self_terms.get(term, 0.) == bias)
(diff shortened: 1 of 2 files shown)


dimod/higherorder/polynomial.py
```diff
@@ -134,11 +134,11 @@ def __eq__(self, other):
         return (
             self.vartype == other.vartype
             and all(
-                (not bias or other_terms[term] == bias)
+                (not bias or other_terms.get(term) == bias)
```
arcondello, line 137: While it's not really an intended use case, I suppose users may want to specify `None` as a bias. Probably better to have a special signal object for this.

mhlr: I am not sure I follow, `.get` returns None if no bias is specified. We could make that explicit with `.get(term, None)` or specify a different value, as long is it is false, so that `other_terms.get(term) == bias` is still guaranteed to be false. Either way the actual result of `.get` is invisible to the user though.

mhlr: `.get(term, 0.)` seems the most correct mathematically

arcondello: That makes sense to me

mhlr: Done

PR #998, opened 2021-09-14
set_objective from iterables

### LINKED ISSUE #919: set_objective from iterables
**Application**
`add_constraint` and `set_objective` should work in a consistent way.

**Proposed Solution**
Support the `Iterable` type for `.set_objective`

```
import dimod

n = 10
cqm = dimod.CQM()
for i in range(n):
    cqm.add_variable(i, vartype='BINARY')

cqm.add_constraint([(i, 1.0) for i in range(n)],
                   sense='<=', rhs=5)
cqm.set_objective([(i, 1.0) for i in range(n)])
```

The first method `.add_constraint` works, but there is no support for the second method using iterables.

hsadeghidw:
Set objective from an iterable similar to `add_constraint` method.

CQM._iterable_to_qm(iterable) -> QuadraticModel that takes that functionality, then just use that in CQM.add_constraint_from_iterable and CQM.set_objective

Closes #919

### DIFF
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -627,7 +627,7 @@ class ConstrainedQuadraticModel:
             >>> cqm.set_objective(2*i - 0.5*i*j + 10)
 
         """
-        if not isinstance(objective, [BinaryQuadraticModel, QuadraticModel]):
+        if isinstance(objective, Iterable):
             objective = self._iterable_to_qm(objective)
         # clear out current objective, keeping only the variables
         self.objective.quadratic.clear()  # there may be a more performant way...
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -370,34 +370,7 @@ class ConstrainedQuadraticModel:
             ...                                           ('i', 'j', 1)], '<=', rhs=1)
 
         """
-        qm = QuadraticModel()
-
-        def _add_variable(v):
-            # handles vartype, and bounds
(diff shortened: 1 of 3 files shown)


dimod/constrained.py
```diff
@@ -642,7 +615,8 @@ def set_objective(self, objective: Union[BinaryQuadraticModel, QuadraticModel]):
         """Set the objective of the constrained quadratic model.
```
arcondello, line 615: Need to add `Iterable` to `objective: Union[BinaryQuadraticModel, QuadraticModel, Iterable]`

dimod/constrained.py
```diff
@@ -652,6 +626,8 @@ def set_objective(self, objective: Union[BinaryQuadraticModel, QuadraticModel]):
             >>> cqm.set_objective(2*i - 0.5*i*j + 10)
 
         """
+        if isinstance(objective, Iterable):
```
arcondello, line 629: I think this works, but `not isinstance(objective, [QuadraticModel, BinaryQuadraticModel)` is likely to be a tiny bit more future-proof...

arcondello (commented):
Also needs a release note. See https://github.com/dwavesystems/dimod#release-notes

Something like
```
``ConstrainedQuadraticModel.set_objective`` now accepts an iterable.
```

arcondello (approved):

Issue #989, opened 2021-09-07
Ocean build from source fails


randomir:
Introducing `dwave-preprocessing` as a full dependency in #924 is causing build from source of dwave-preprocessing (and consequently the whole Ocean) to fail. Note: dwave-preprocessing has dimod as a build dependency.

See https://github.com/dwavesystems/dwave-preprocessing/issues/26.

Proposed solution: revert to optional dependency on dwave-preprocessing until we remove it completely in 0.11.

randomir:
Not sure that's the only way, just my first thought. I don't mind we try/catch it -- although those would have to be at the root level, when importing everything. Feel free to investigate it further. Thanks!

PR #999, opened 2021-09-14
Don't require dwave-preprocessing to be installed to import dimod

arcondello:
Tested locally the `dwave-preprocessing` install. Credit to @hhtong for the nice commit history that made this very simple.

Closes https://github.com/dwavesystems/dimod/issues/989


randomir (approved):
LGTM!

PR #1001, opened 2021-09-15
Update outdated link 

JoelPasvolsky:
Will cause a linkcheck failure on the SDK: 
```
docs_dimod/reference/generated/dimod.SampleSet.from_samples.rst:2: [broken] https://numpy.org/doc/stable/user/basics.creation.html#converting-python-array-like-objects-to-numpy-arrays: Anchor 'converting-python-array-like-objects-to-numpy-arrays' not found
```
An alternative (for several places where `.. _array_like:` is defined) for consideration is 
`https://github.com/numpy/numpy/blob/main/numpy/typing/_array_like.py`
(see https://github.com/numpy/numpy/blob/dc7dafe70a53d6c122091516f34058bd0a6d89e1/numpy/typing/_array_like.py#L60)

### DIFF
--- a/dimod/sampleset.py
+++ b/dimod/sampleset.py
@@ -568,7 +568,7 @@ class SampleSet(abc.Iterable, abc.Sized):
             >>> sampleset.variables
             Variables(['a', 'b', 'c'])
 
-        .. _array_like:  https://numpy.org/doc/stable/user/basics.creation.html#converting-python-array-like-objects-to-numpy-arrays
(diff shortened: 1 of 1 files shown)


arcondello:
If we were going to point at their typing, I would like to [ArrayLike](https://numpy.org/devdocs/reference/typing.html#arraylike) instead of the source code. That said, I still prefer https://numpy.org/doc/stable/user/basics.creation.html

arcondello (approved):

PR #1002, opened 2021-09-15
Add offset keyword agument to BQM

arcondello:


randomir (approved):
LGTM!

dimod/binary/binary_quadratic_model.py
```diff
@@ -79,6 +79,58 @@ class BinaryQuadraticModel(QuadraticViewsMixin):
...
+            Create a BQM from linear and quadratic biases and an offset.
+            ``offset`` must be a number.
+
+    Args:
+        *args: See above
+
+        offset: The offset (see above) may be supplied as a keyword argument.
+
+        vartype: The variable type (see above) may be supplied as a keyword
+            argument.
```
JoelPasvolsky, line 126: Consider putting the definitions here instead of the (see above) - vartype is used in more than one constructor, why make them look through them?

JoelPasvolsky (approved):
Thanks!

dimod/binary/binary_quadratic_model.py
arcondello: I had them all duplicated but it just made it word soup. So I used the same logic as for acronyms, the first time we see it it gets expanded.

dimod/binary/binary_quadratic_model.py
```diff
@@ -79,6 +79,58 @@ class BinaryQuadraticModel(QuadraticViewsMixin):
...
+            Create a BQM from linear and quadratic biases, where ``linear`` is a
+            one-dimensional array_like_ or a dictionary of the form
+            ``{v: b, ...}``.
+
+        ``BinaryQuadraticModel(linear, quadratic, offset, vartype)``
+            Create a BQM from linear and quadratic biases and an offset.
+            ``offset`` must be a number.
+
+    Args:
+        *args: See above
```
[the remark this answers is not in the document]
arcondello: I'll let you fix this one on your next pass through :laughing:

Issue #1006, opened 2021-09-20
is_feasible method


hsadeghidw:
A quick way to find out if a sample is feasible or not based on a dimod.CQM model. Related to #971 

perhaps makes use of `.violations` and summarizes whether the solution is feasible or not

potentially can return a bool or list of bool (overall is_feasible and/or each constraint is_feasible

PR #1010, opened 2021-09-20
Add CQM.iter_constraint_data, CQM.iter_violations and CQM.violations

arcondello:
Closes https://github.com/dwavesystems/dimod/issues/971
Closes https://github.com/dwavesystems/dimod/issues/975 as redundant

### DIFF
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -26,7 +26,7 @@ import zipfile
 
 from numbers import Number
 from typing import Hashable, Optional, Union, BinaryIO, ByteString, Iterable, Collection, Dict
-from typing import Callable, MutableMapping, Iterator, Tuple, Mapping, Any
+from typing import Callable, MutableMapping, Iterator, Tuple, Mapping, Any, NamedTuple
 
 import numpy as np
 
@@ -34,6 +34,7 @@ from dimod.core.bqm import BQM as BQMabc
 from dimod.binary.binary_quadratic_model import BinaryQuadraticModel, Binary, Spin, as_bqm
 from dimod.discrete.discrete_quadratic_model import DiscreteQuadraticModel
 from dimod.quadratic import QuadraticModel
+from dimod.sampleset import as_samples
 from dimod.sym import Comparison, Eq, Le, Ge, Sense
 from dimod.serialization.fileview import SpooledTemporaryFile, _BytesIO
 from dimod.serialization.fileview import load, read_header, write_header
@@ -47,6 +48,15 @@ __all__ = ['ConstrainedQuadraticModel', 'CQM', 'cqm_to_bqm']
 
 
 CQM_MAGIC_PREFIX = b'DIMODCQM'
+
+
+class ConstraintData(NamedTuple):
+    label: Hashable
+    lhs_energy: float
+    rhs_energy: float
+    sense: Sense
+    activity: float
+    violation: float
 
 
 class ConstrainedQuadraticModel:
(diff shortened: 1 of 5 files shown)


dimod/constrained.py
```diff
@@ -861,6 +987,14 @@ def vartype(self, v: Variable) -> Vartype:
         """The vartype of the given variable."""
         return self.objective.vartype(v)
 
+    def violations(self, sample_like, *, skip_satisfied: bool = False, clip: bool = False,
```
hsadeghidw, line 990: I wonder if we should also accept a `tolerance` parameter for `clip`. But a better place for it is in a method such as `is_feasbile`. Can we add a `is_feasible` method in this PR or should I make a new issue for it?

hsadeghidw (approved):
I left a suggestion to add a method and a new parameter, but it's up to you. You can merge this one and we start a new PR.

arcondello:
Hmm, I had planned to do it as a separate PR but might as well combine them I suppose.

arcondello:
@hsadeghidw do you have a test case that's within tolerance?

arcondello:
Now also closes https://github.com/dwavesystems/dimod/issues/1006

hsadeghidw:
There seems to be an issue:

both cases should have about the same amount of violations. but one is zero the other is a small number. They are both correctly marked as feasible given the tolerance.

```
import dimod
from dimod import quicksum, Binary, Integer
cqm = dimod.ConstrainedQuadraticModel()
x = Binary('x')
z = Binary('z')
y = Integer('y', upper_bound=1e6) * 1e-6

cqm.add_variable('a', vartype='BINARY')
cqm.set_objective(x + y + z)
cqm.add_constraint(x + y + z == 3)
sample = {'x': 1, 'y': 999999.999, 'z': 1}
violations = cqm.violations(sample)
print(cqm.check_feasible(sample))
print(violations)
for v in cqm.iter_constraint_data(sample):
    print(v)
sample = {'x': 1, 'y': 1000000.001, 'z': 1}
violations = cqm.violations(sample)
print(cqm.check_feasible(sample))
print(violations)
for v in cqm.iter_constraint_data(sample):
    print(v)

```

arcondello:
Ok, so I am going to say that that error is unrelated to this PR. I can recreate it with
```python
>>> import dimod
>>> x, z = dimod.Binaries('xz')
>>> y = dimod.Integer('y', upper_bound=1e6) * 1e-6
>>> (x + y + z).energy({'x': 1, 'y':1000000.001, 'z': 1})
3.0
```
has to do with the precision of the energy calculations.

hsadeghidw (approved):

arcondello:
Yeah, I'll make an issue. It's not really a bug so much as a behavior we have to explain...

rahmani15:
I am not sure if we should check for the correctness of input samples. For example here I provide sample that are out of bounds for binary variables. But there is no error and it returns True when I check feasibility. 
```
i, j, k = dimod.Binaries(['i', 'j', 'k'])
cqm = dimod.ConstrainedQuadraticModel()
cqm.add_discrete(['i', 'j', 'k'])
sample = {'i': 2, 'j': 0, 'k': -1}
print(cqm.check_feasible(sample))
```

arcondello:
If we did want to do that, it should be in the energy function. See https://github.com/dwavesystems/dimod/issues/954.

rahmani15:
This also returns True for the above example:
`sample = {'i': 1.2, 'j': 0, 'k': -.2}`

arcondello:
In a context of not checking the inputs for bounds/type, I think that's the most sensible thing do to.

releasenotes/notes/cqm-constraint-data-c8217056d499f800.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
+  - |
+    Add ``ConstrainedQuadraticModel.iter_constraint_data()``,
+    ``ConstrainedQuadraticModel.iter_violations()`` and
+    ``ConstrainedQuadraticModel.violations()`` methods.
+    See `971 <https://github.com/dwavesystems/dimod/issues/971>`_.
```
rahmani15, line 7: Do we need a new blank line in the end?

arcondello: Nope! It's a good habit for code files since it means appending to file won't affect the git history for unrelated lines. But since these files are never touched again it doesn't matter. Good eye though!

rahmani15 (approved):
LGTM

PR #1017, opened 2021-09-29
Add `SampleSet.filter()` method

arcondello:
This further allows nice syntax like
```python
sampleset.filter(lambda d: not d.is_feasible).first
```
to get the lowest-energy feasible sample.

### DIFF
--- a/dimod/sampleset.py
+++ b/dimod/sampleset.py
@@ -20,6 +20,7 @@ import numbers
 import collections.abc as abc
 
 from collections import namedtuple
+from typing import Any, Callable
 
 import numpy as np
 from numpy.lib import recfunctions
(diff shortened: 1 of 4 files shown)


dimod/sampleset.py
```diff
@@ -1151,6 +1153,38 @@ def hook(sampleset):
...
 
+    def filter(self, pred: Callable[[Any], bool]) -> 'SampleSet':
+        """Return a new sampleset with rows filtered out by the given predicate.
+
+        Args:
+            pred: A function that accepts a named tuple as returned by
+                :meth:`.data` and returns a :class:`bool`.
+
+        Returns:
+            A sample set with only the data rows for which ``pred`` returns
```
JoelPasvolsky, line 1164: "Return a new sampleset" vs "A sample set"

JoelPasvolsky (approved):
LGTM, thank you!
Do we want an `in_place` option for it?

arcondello:
We generally treat sample sets as immutable, so no in-place option.

hsadeghidw:
Is this a typo or the expected behavior?

`sampleset.filter(lambda d: not d.is_feasible).first`

The above should get the infeasible solution.

arcondello:
My confusion about `filter` vs `filterfalse`, oops... See #1018

PR #1018, opened 2021-09-29
Fix SampleSet.filter to filter rather than filter out

arcondello:


randomir (approved):
LGTM

JoelPasvolsky:
Should we also add
"""
Raises:
    AttributeError:: If the given sampleset does not contain the specified predicate as a field.
"""

arcondello:
That's actually a property of the predicate that you're passing in.

PR #1028, opened 2021-10-19
Fixes misspelled `upper_boun` in the cqm from LP file.

alexzucca90:
Embarassing bug in the cqm from lp file code...

### DIFF
--- a/releasenotes/notes/fix-integer-quadratic-lp-bug-ad4faf5379ef7bba.yaml
+++ b/releasenotes/notes/fix-integer-quadratic-lp-bug-ad4faf5379ef7bba.yaml
@@ -2,6 +2,7 @@
 
 fixes:
   - |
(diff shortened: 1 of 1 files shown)


releasenotes/notes/fix-integer-quadratic-lp-bug-ad4faf5379ef7bba.yaml
```diff
@@ -0,0 +1,7 @@
+---
+
+fixes:
+  - |
+    Fix misspelled `upper_bound` in `from_lp_file` constructor method when
```
arcondello, line 5: I think something like
```
No longer throw a ``TypeError`` from ``ConstrainedQuadraticModel.from_lp_file`` when adding a integer quadratic term from constraints.
```
would be clearer, or at least describes the broken behaviour

arcondello (approved):

PR #1037, opened 2021-11-02
Add ExactCQMSolver and tests

necaisej:


dimod/reference/samplers/exact_solver.py
```diff
@@ -199,6 +200,76 @@ def sample_dqm(self, dqm, **kwargs):
...
+        
+    """
+    properties = None
+    parameters = None
+
+    def __init__(self):
+        self.properties = {}
+        self.parameters = {}
+        
+    def sample_cqm(self, cqm, rtol: float = 1e-6, atol: float = 1e-8, **kwargs):
```
JoelPasvolsky, line 242: should this have a `-> SampleSet` ?

dimod/reference/samplers/exact_solver.py
```diff
@@ -199,6 +200,76 @@ def sample_dqm(self, dqm, **kwargs):
...
+        """Sample from a constrained quadratic model.
+
+        Args:
+            cqm (:obj:`~dimod.ConstrainedQuadraticModel`):
+                Constrained quadratic model to be sampled from.
+
```
JoelPasvolsky, line 248: What about args `rtol` and `atol`?

docs/reference/sampler_composites/samplers.rst
```diff
@@ -67,6 +67,29 @@ Methods
 
    ExactDQMSolver.sample_dqm
 
+Exact CQM Solver
+------------
```
JoelPasvolsky, line 71: Needs to match length of header

dimod/reference/samplers/exact_solver.py
necaisej: The other exact solvers don't have this present; would it be preferable for them to?

dimod/reference/samplers/exact_solver.py
```diff
@@ -228,4 +303,59 @@ def _all_cases_dqm(dqm):
...
+            break
+            
+        l = []
+        for i in range(len(indexes)):
+            s = np.zeros(d_cases[i])
+            s[indexes[i]] = 1
+            l = np.concatenate([l, s])
+        
+        if len(c1):
+            [combinations.append(np.concatenate([l, row])) for row in c1]
```
randomir, line 340: Avoid list comprehensions with side-effects.

randomir (commented):
Needs better test coverage.

dimod/reference/samplers/exact_solver.py
necaisej: Added `-> SampleSet` to each of the `sample_{}` methods

dimod/reference/samplers/exact_solver.py
```diff
@@ -199,6 +200,80 @@ def sample_dqm(self, dqm, **kwargs):
...
+    Examples:
+        This example solves a CQM with 3 variables and 1 constraint.
+        
+        >>> from dimod import ConstrainedQuadraticModel, Binary
+        >>> cqm = ConstrainedQuadraticModel()
+        >>> x, y, z = Binary('x'), Binary('y'), Binary('z')
+        >>> cqm.set_objective(x*y + 2*y*z)
+        >>> cqm.add_constraint(x*y == 1)      # doctest: +SKIP
+        >>> sampleset = ExactCQMSolver().sample_cqm(cqm)
+        >>> print(sampleset)      # doctest: +SKIP
```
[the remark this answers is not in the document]
necaisej: This was a mistake, added

dimod/reference/samplers/exact_solver.py
```diff
@@ -199,6 +200,80 @@ def sample_dqm(self, dqm, **kwargs):
...
+
+    def __init__(self):
+        self.properties = {}
+        self.parameters = {}
+        
+    def sample_cqm(self, cqm, rtol: float = 1e-6, atol: float = 1e-8, **kwargs):
+        """Sample from a constrained quadratic model.
+
+        Args:
+            cqm (:obj:`~dimod.ConstrainedQuadraticModel`):
```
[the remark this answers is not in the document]
necaisej: Replaced `:obj:` with `:class:`, let me know if it is consistent with the standard or if I made a mistake?

dimod/reference/samplers/exact_solver.py
```diff
@@ -199,6 +200,80 @@ def sample_dqm(self, dqm, **kwargs):
...
+        
+        possible_samples = as_samples(_all_cases_cqm(cqm))
+        energies = cqm.objective.energies(possible_samples)
+        
+        
+        # from_samples requires a single vartype argument, but QuadraticModel
+        # and therefore CQM allow mixed vartypes. For now, only passing 'INTEGER'
+        
+        response = SampleSet.from_samples(possible_samples, 'INTEGER', energies)
+        is_satisfied_list = [[violation <= atol + rtol*abs(datum.energy) for l, violation in cqm.iter_violations(datum.sample)] for datum in response.data(sorted_by=None)]
```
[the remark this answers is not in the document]
necaisej: Turns out it isn't necessary to cast to `np.array`, fixed

dimod/reference/samplers/exact_solver.py
```diff
@@ -228,4 +303,59 @@ def _all_cases_dqm(dqm):
...
+    """Get a numpy array containing all possible samples as lists of valid values"""
+    # developer note: the following is a catch all method, and it may be possible
+    # to streamline or refactor. It is fine for its intended use
+    
+    var_list = list(cqm.variables)
+    
+    # Set aside logical discrete variables, and the Binary variables which make them up
+    d_cases = [len(cqm.constraints[d].lhs) for d in cqm.discrete]
+    d_vars = [x for l in [[v for v in cqm.constraints[d].lhs.variables] for d in cqm.discrete] for x in l]
+    [var_list.remove(d) for d in d_vars]
```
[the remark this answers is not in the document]
necaisej: Ah... of course.

`d_vars` are guaranteed to not have overlapping variables because they are only pulled from `cqm.discrete`. The `cqm.add_discrete` method checks for collisions in the one-hot constraints that enforce logical discrete variables

dimod/reference/samplers/exact_solver.py
```diff
@@ -228,4 +303,59 @@ def _all_cases_dqm(dqm):
...
+    
+    # Construct all combinations of spin, integer, and binary variables not included in discrete
+    cases = [(_iterator_by_vartype(cqm, v)) for v in var_list]
```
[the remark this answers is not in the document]
necaisej: No, just unnecessary. Removed `()`

dimod/reference/samplers/exact_solver.py
```diff
@@ -226,6 +290,59 @@ def _all_cases_dqm(dqm):
...
+    """Get a numpy array containing all possible samples as lists of valid values"""
+    # developer note: the following is a catch all method, and it may be possible
+    # to streamline or refactor. It is fine for its intended use
+    
+    var_list = list(cqm.variables)
+    
+    # Set aside logical discrete variables, and the Binary variables which make them up
+    d_cases = [len(cqm.constraints[d].lhs) for d in cqm.discrete]
+    d_vars = [x for l in [list(cqm.constraints[d].lhs.variables) for d in cqm.discrete] for x in l]
+    var_list = [v for v in var_list if v not in set(d_vars)]
```
randomir, line 306: This will create a new set for _each_ variable in `var_list`!

dimod/reference/samplers/exact_solver.py
```diff
@@ -73,30 +73,28 @@ def __init__(self):
         self.properties = {}
         self.parameters = {}
 
-    def sample(self, bqm, **kwargs):
+    def sample(self, bqm, **kwargs) -> SampleSet:
```
arcondello, line 76: If we're going to be touching other classes/methods to add annotations, we should also add the input types. E.g. `bqm: BinaryQuadraticModel`. You can avoid the circular imports that are likely to occur with
```python
if typing.TYPE_CHECKING:
    from dimod import BinaryQuadraticModel
```
see https://stackoverflow.com/questions/39740632/python-type-hinting-without-cyclic-imports

dimod/reference/samplers/exact_solver.py
```diff
@@ -177,28 +176,93 @@ def __init__(self):
...
+        combination of variable cases. It becomes slow very quickly
+        
+    Examples:
+        This example solves a CQM with 3 variables and 1 constraint.
+        
+        >>> from dimod import ConstrainedQuadraticModel, Binary
+        >>> cqm = ConstrainedQuadraticModel()
+        >>> x, y, z = Binary('x'), Binary('y'), Binary('z')
+        >>> cqm.set_objective(x*y + 2*y*z)
+        >>> cqm.add_constraint(x*y == 1)      # doctest: +SKIP
```
arcondello, line 215: is this solver not deterministic? I think if you use the `label=` keyword argument when adding the constraint you shouldn't need to do any of the `# doctest: +SKIP`

dimod/reference/samplers/exact_solver.py
```diff
@@ -177,28 +176,93 @@ def __init__(self):
...
+        2 1 0 0    0.0       1 arra...   False
+        4 0 0 1    0.0       1 arra...   False
+        6 1 0 1    0.0       1 arra...   False
+        3 1 1 0    1.0       1 arra...    True
+        5 0 1 1    2.0       1 arra...   False
+        7 1 1 1    3.0       1 arra...    True
+        ['INTEGER', 8 rows, 8 samples, 3 variables]
+        
+    """
+    properties = None
```
arcondello, line 230: These can be removed. We're not inheriting from `Sampler` so you don't need these class attributes to satisfy the ABC.

necaisej: So something like
```def __init__(self):
        pass```
and removing ```Sampler.remove_unknown_kwargs(self, **kwargs)``` would be fine?

Comment (arcondello) [reply dimod/reference/samplers/exact_solver.py:234]: No, leave the `__init__` the way it is, just remove the
```python
properties = None
parameters = None
```

dimod/reference/samplers/exact_solver.py
```diff
@@ -20,14 +20,18 @@
...
 
 from dimod.core.sampler import Sampler
-from dimod.sampleset import SampleSet, as_samples
+from dimod.sampleset import SampleSet
 from dimod.core.polysampler import PolySampler
 from dimod.vartypes import Vartype
-from dimod.exceptions import SamplerUnknownArgWarning
 
-__all__ = ['ExactSolver', 'ExactPolySolver', 'ExactDQMSolver']
+from typing import TYPE_CHECKING
```
arcondello, line 30: Not to be terribly pedantic, but let's keep our imports clean by following [Pep8](https://www.python.org/dev/peps/pep-0008/#imports). Specifically
1) Standard library imports.
2) Related third party imports.
3) Local application/library specific imports.

so should be
```python
from itertools import product
from typing import TYPE_CHECKING

import numpy as np

from dimod.core.sampler import ...

File: dimod/reference/samplers/exact_solver.py [hunk 15]
```diff
@@ -20,14 +20,18 @@
...
+if TYPE_CHECKING:
```
arcondello, line 31: btw, you might not actually need this `if` branch, I was assuming you would but worth checking explicitly

dimod/reference/samplers/exact_solver.py
```diff
@@ -73,30 +77,28 @@ def __init__(self):
         self.properties = {}
         self.parameters = {}
 
-    def sample(self, bqm, **kwargs):
+    def sample(self, bqm: BinaryQuadraticModel, **kwargs) -> SampleSet:
```
arcondello, line 80: To resolve the CI failures, you can using strings. e.g. `bqm: 'BinaryQuadraticModel'`

dimod/reference/samplers/exact_solver.py
necaisej: Double checked, the `if` is necessary - thanks for the tip!

dimod/reference/samplers/exact_solver.py
```diff
@@ -22,12 +22,16 @@
...
 
 from dimod.core.sampler import Sampler
-from dimod.sampleset import SampleSet, as_samples
+from dimod.sampleset import SampleSet
 from dimod.core.polysampler import PolySampler
 from dimod.vartypes import Vartype
-from dimod.exceptions import SamplerUnknownArgWarning
 
-__all__ = ['ExactSolver', 'ExactPolySolver', 'ExactDQMSolver']
+from itertools import product
```
arcondello, line 29: Still want these above the others since they are standard library imports

necaisej: Apologies, I need to get my eyes checked apparently

arcondello (commented):
I just noticed we don't have any tests with integer variables. Definitely need those.

arcondello:
I think the only thing left is it needs a release note

arcondello (approved):

arcondello:
@JoelPasvolsky and @randomir let me know if you want to finish reviewing

dimod/reference/samplers/exact_solver.py
```diff
@@ -170,35 +173,96 @@ class ExactDQMSolver():
...
+        self.parameters = {}
+        
+    def sample_cqm(self, cqm: 'ConstrainedQuadraticModel', rtol: float = 1e-6, atol: float = 1e-8, **kwargs) -> SampleSet:
+        """Sample from a constrained quadratic model.
+
+        Args:
+            cqm (:class:`~dimod.ConstrainedQuadraticModel`):
+                Constrained quadratic model to be sampled from.
+            rtol (float):
+                The relative tolerance for constraint violation, scales sample energy
```
JoelPasvolsky, line 244:
```suggestion
                Relative tolerance for constraint violation. A fraction of sample energy any violation must be smaller than for the constraint to be satisfied. 
```

dimod/reference/samplers/exact_solver.py
```diff
@@ -170,35 +173,96 @@ class ExactDQMSolver():
...
+    def sample_cqm(self, cqm: 'ConstrainedQuadraticModel', rtol: float = 1e-6, atol: float = 1e-8, **kwargs) -> SampleSet:
+        """Sample from a constrained quadratic model.
+
+        Args:
+            cqm (:class:`~dimod.ConstrainedQuadraticModel`):
+                Constrained quadratic model to be sampled from.
+            rtol (float):
+                The relative tolerance for constraint violation, scales sample energy 
+            atol (float):
+                The absolute tolerance for constraint violations, constant for sample energy
```
JoelPasvolsky, line 246:
```suggestion
                Absolute tolerance for constraint violations. A constant any violation must be smaller than for the constraint to be satisfied. 
```

dimod/reference/samplers/exact_solver.py
```diff
@@ -170,35 +173,96 @@ class ExactDQMSolver():
...
+
+        Returns:
+            :class:`~dimod.SampleSet`
```
JoelPasvolsky, line 249: Do we keep a Returns field when it repeats the type annotation ("-> SampleSet")? It would make more sense to me if we added some description here about the sampleset containing the non-standard `is_feasible` and `is_satisfied` fields.

JoelPasvolsky (approved):

dimod/reference/samplers/exact_solver.py
arcondello: You're right, we keep only the description not the type. Generally we would do
```
    Returns:
        All possible samples.
```
or something like that.

randomir (approved):
LGTM!

PR #1040, opened 2021-11-11
Asymmetric multiplication circuits

mhlr:
enable `multiplication_circuit()` to create circuits with different size inputs


arcondello:
Is this ready for review or are you still working on it?

mhlr:
> Is this ready for review or are you still working on it?

Thhe unit test needs fixing, but the code is code. Needed to work on something else for a bit. Should  finish today

mhlr:
FYI there is a bug in the current unt tests. It tests equality of `.sort()` whixh returns None. Fixed in my unpushed commits

dimod/generators/gates.py
```diff
@@ -13,10 +13,12 @@
 #    limitations under the License.
 
 from collections import defaultdict
-
+from typing import Optional
+import itertools as it
```
arcondello, line 17: To be consistent with the rest of dimod, I think it's better to use `import itertools` or `from itertools import product, starmap`.

Also while we're at it we should follow pep8's [import ordering](https://www.python.org/dev/peps/pep-0008/#imports) specifically
1) Standard library imports.
2) Related third party imports.
3) Local application/library specific imports.

dimod/generators/gates.py
```diff
@@ -13,10 +13,12 @@
...
 from dimod.binary.binary_quadratic_model import BinaryQuadraticModel
 from dimod.typing import Variable
 from dimod.vartypes import Vartype
+from dimod import quicksum
```
arcondello, line 21: Better to use the full path. Avoids some circular dependency issues as well as making it easier to find for other developers later if needed.

tests/test_generators.py
```diff
@@ -617,11 +617,39 @@ def test_xor_gate_strength(self):
...
+        mc3_vars = {
+            'a0', 'a1', 'a2',
+            'b0', 'b1', 'b2',
+            'and0,1', 'and0,2', 'and1,0', 'and1,1', 'and1,2', 'and2,0', 'and2,1', 'and2,2',
+            'sum1,1', 'sum1,2',
+            'carry1,0', 'carry1,1', 'carry1,2', 'carry2,0', 'carry2,1',
+            'p0', 'p1', 'p2', 'p3', 'p4', 'p5',
+        }
         bqm = dimod.generators.multiplication_circuit(3)
-        self.assertEqual((list(bqm.variables)).sort(), mc3_vars)
```
arcondello, line 624: Good catch, though I am confused about how this was passing before?

dimod/generators/gates.py
```diff
@@ -249,109 +252,66 @@ def multiplication_circuit(nbit: int) -> BinaryQuadraticModel:
...
-        else:
-            carryout = p[2 * nbit - 1]
-
-        gate = fulladder_gate(inputs[0], inputs[1], inputs[2], sumout, carryout)
-        bqm.update(gate)
-
-    return bqm
+    a = {i: 'a%d' % i for i in range(num_multiplier_bits)}
+    b = {j: 'b%d' % j for j in range(num_multiplicand_bits)}
+    p = {k: 'p%d' % k for k in  range(num_multiplier_bits + num_multiplicand_bits)}
```
arcondello, line 277: Looks like you no longer use `p`. Honestly we probably don't need to make these in advance at all

dimod/generators/gates.py
```diff
@@ -249,109 +252,66 @@ def multiplication_circuit(nbit: int) -> BinaryQuadraticModel:
...
+        bqm = and_gate(a[i], b[j], inputs[0])
+        if i > 0:
+            if j < num_multiplicand_bits - 1:
+                inputs.append(SUM(i-1, j+1) if i > 1 else AND(0, j+1))
+            if j > 0:
+                inputs.append(CARRY(i, j-1))
+        l = len(inputs)
+        if l > 1:
+            outputs = SUM(i,j), CARRY(i, j)
+            if l==2:
```
arcondello, line 310: pep8: need whitespace around operator

tests/test_generators.py
mhlr: The definition mc3_vars had the same problem. The this was testing for None == None regardless of input so it could never fail.

dimod/generators/gates.py
mhlr: I can make them functions like AND, SUM & CARRY. This his how they were to start of with. I have no strong feelings about these. unlike AND, SUM & CARRY which were adding unnecessary complexity as mutating nested dicts.

mhlr: ... or icould just inline them as string interpolations

arcondello: I was imagining the inline, but whatever works.

dimod/generators/gates.py
mhlr: Done

dimod/generators/gates.py
mhlr: Done

dimod/generators/gates.py
mhlr: Done

arcondello:
CI failure doesn't seem to be your fault, I've triggered a rerun

dimod/generators/gates.py
```diff
@@ -210,13 +213,13 @@ def xor_gate(in0: Variable, in1: Variable, out: Variable, aux: Variable,
...
 
-def multiplication_circuit(nbit: int) -> BinaryQuadraticModel:
+def multiplication_circuit(num_arg1_bits: int, num_arg2_bits: Optional[int] = None) -> BinaryQuadraticModel:
     """Return a binary quadratic model with ground states corresponding to
     a multiplication circuit.
 
     The generated BQM represents the binary multiplication :math:`ab=p`,
-    where the multiplicands are binary variables of length `nbit`; for example,
-    :math:`2^ma_{nbit} + ... + 4a_2 + 2a_1 + a0`.
+    where the args are binary variables of length `num_arg1_bits` and 'num_arg2_bits;
```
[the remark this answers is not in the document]
mhlr: done

arcondello (approved):

PR #1044, opened 2021-11-16
Fix a collection of small issues

arcondello:


dimod/decorators.py
```diff
@@ -197,28 +182,21 @@ def bqm_structured(f):
...
-            # todo: better error message
-            raise BinaryQuadraticModelStructureError("given bqm does not match the sampler's structure")
-        if not all(u in adjacency[v] for u, v in bqm.quadratic):
-            # todo: better error message
-            raise BinaryQuadraticModelStructureError("given bqm does not match the sampler's structure")
-
+    def structured_sample(sampler, bqm, **kwargs):
+        adjacency = sampler.adjacency
+
+        if not adjacency.keys() >= bqm.variables:
```
hhtong, line 189: Would `adjacency.keys() < bqm.variables` be better for readability?

dimod/decorators.py
```diff
@@ -197,28 +182,21 @@ def bqm_structured(f):
...
+            raise BinaryQuadraticModelStructureError(
+                f"given bqm constains a variable, {(adjacency.keys() - bqm.variables).pop()!r}, "
+                "not supported by the structured solver")
+        if any(u not in adjacency[v] for u, v, _ in bqm.iter_quadratic()):
+            u, v = next((u, v) for u, v, _ in bqm.iter_quadratic() if u not in adjacency[v])
+            raise BinaryQuadraticModelStructureError(
+                f"given bqm constains an interaction, {(u, v)!r}, "
```
hhtong, line 196:
```suggestion
                f"given bqm contains an interaction, {(u, v)!r}, "
```

hhtong (approved):
LGTM

dimod/decorators.py
arcondello: In this case they have a different meaning. The current syntax says "not superset", yours says "is subset".

PR #1047, opened 2021-11-17
Fix pyBQM.energies

arcondello:
Duplicates #1046, possibly avoiding CI failures


jackraymond (commented):
Looks fine, tried failing test from yesterday and worked fine

PR #1051, opened 2021-11-19
Backport modern quadratic views to deprecated BQMs

arcondello:


dimod/core/bqm.py
```diff
@@ -897,6 +668,27 @@ def min_and_max(iterable):
         else:
             return 1.0
 
+    def reduce_linear(self, function, initializer=None):
+        gen = (self.get_linear(v) for v in self.variables)
+        if initializer is None:
+            return functools.reduce(function, gen)
+        else:
+            return functools.reduce(function, gen, initializer)
```
randomir, line 676:
```suggestion
    def reduce_linear(self, function, *args):
        gen = (self.get_linear(v) for v in self.variables)
        return functools.reduce(function, gen, *args)
```
Shorter, with initializer undocumented, but it also accepts `None` for initial value.

randomir: Or
```suggestion
    def reduce_linear(self, function, initializer=_value_missing):
        gen = (self.get_linear(v) for v in self.variables)
        if initializer is _value_missing:
            return functools.reduce(function, gen)
        else:
            return functools.reduce(function, gen, initializer)
```
with
```
_value_missing = object()
```
in module scope.

randomir (approved):
LGTM.

dimod/core/bqm.py
arcondello: In this case the biases are known to be numbers, which is why I was using `None`. But suppose `DictBQM` supports arbitrary biases so may as well use the `_value_missing` version.

arcondello: Once I started pulling on this thread I found some other stuff I wanted to do, so  I am going to make this change in a followup PR

PR #1052, opened 2021-11-23
Add CQM.to_lp_file

alexzucca90:
added function to serialize CQM as LP file

### DIFF
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -1056,6 +1056,140 @@ class ConstrainedQuadraticModel:
         file.seek(0)
         return file
 
+    def to_lp_file(self, *, spool_size: int = int(1e9)) -> tempfile.SpooledTemporaryFile:
+        """Serialize to LP file format.
+        Args:
+            spool_size: Defines the `max_size` passed to the constructor of
+                :class:`tempfile.SpooledTemporaryFile`. Determines whether
+                the returned file-like's contents will be kept on disk or in
+                memory.
+
+        Format Specifications:
+            This format follows the directions in
+            https://www.gurobi.com/documentation/9.5/refman/lp_format.html#format:LP
+
+        """
+
+        sign_dict = {1: '+', -1: '-', 0: '+'}
+        sense_dict = {'==': '=', '<=': '<=', '>=': '>='}
+
+        # forbidden chars in the LP format
+        not_allowed_chars = ['+', '-', '*', '^', ':']
+        not_first_char = ['<', '>', '=', '(', ')', '[', ']', ',']
+        not_first_char += [str(i) for i in range(10)]
+
+        new_line_every = 20
+
+        f = SpooledTemporaryFile(max_size=spool_size, mode='wt')
+
+        if self.objective.linear or self.objective.quadratic:
+            f.write("Minimize\n ")
+            obj_string = 'obj: '
+
+            i_term = 0
+            if self.objective.linear:
+                for var, bias in self.objective.linear.items():
+                    obj_string += '{} {} {} '.format(sign_dict[int(np.sign(bias))], abs(bias), var)
+                    i_term += 1
+
+                    if i_term % new_line_every == 0:
+                        obj_string += '\n '
+
+            if self.objective.quadratic:
+                obj_string += '+ [ '
+                for edge, bias in self.objective.quadratic.items():
+                    var1, var2 = edge
+                    obj_string += '{} {} {} * {} '.format(sign_dict[np.sign(bias)], abs(2 * bias), var1, var2)
+                    i_term += 1
+                    if i_term % new_line_every == 0:
+                        obj_string += '\n '
+                obj_string += ']/2 '
+
+            f.write(obj_string)
+            f.write("\n \n")
+
+        if self.constraints:
+            f.write("Subject To \n")
+
+            for label, constraint in self.constraints.items():
+
+                if label[0] in not_first_char:
+                    raise ValueError('Cannot start variable name with {}'.format(label[0]))
+
+                if any([l in not_allowed_chars for l in label]):
+                    raise ValueError('Cannot use characters {} in names'.format(not_allowed_chars))
+
+                c_string = '{}: '.format(label)
+
+                i_term = 0
+                if constraint.lhs.linear:
+                    for var, bias in constraint.lhs.linear.items():
+                        c_string += '{} {} {} '.format(sign_dict[int(np.sign(bias))], abs(bias), var)
+                        i_term += 1
+
+                        if i_term % new_line_every == 0:
+                            c_string += '\n '
+
+                if constraint.lhs.quadratic:
+                    c_string += '+ [ '
+                    for edge, bias in constraint.lhs.quadratic.items():
+                        var1, var2 = edge
+                        c_string += '{} {} {} * {} '.format(sign_dict[np.sign(bias)], abs(bias), var1, var2)
+                        i_term += 1
+                        if i_term % new_line_every == 0:
+                            c_string += '\n '
+                    c_string += '] '
+
+                rhs = constraint.rhs - constraint.lhs.offset
+                c_string += ' {} {} \n'.format(sense_dict[constraint.sense.value], rhs)
+                f.write(c_string)
+
+        vartypes = {var: self.vartype(var) for var in self.variables}
+
+        # write the bounds
+        f.write('\n')
+        f.write('Bounds \n')
+        for v, vartype in vartypes.items():
+            if any([vv in not_allowed_chars for vv in v]):
+                raise ValueError('Cannot use chars {} in variable name'.format(not_allowed_chars))
+            if v[0] in not_first_char:
+                raise ValueError('Cannot start variable name with {}'.format(v[0]))
+            if vartype == Vartype.INTEGER:
+                bound_string = ' {} <= {} <= {}\n'.format(self.lower_bound(v), v, self.upper_bound(v))
+                f.write(bound_string)
+            elif vartype == Vartype.SPIN:
+                raise ValueError('SPIN variables not supported in LP files, convert them to BINARY beforehand.')
+
+        # Write the Binary variables
+        f.write('\n')
+        f.write('Binary \n')
+        var_string = ''
+        i_term = 0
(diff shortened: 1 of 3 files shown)


dimod/constrained.py
```diff
@@ -1056,6 +1056,140 @@ def to_file(self, *, spool_size: int = int(1e9)) -> tempfile.SpooledTemporaryFil
...
+
+        vartypes = {var: self.vartype(var) for var in self.variables}
+
+        # write the bounds
+        f.write('\n')
+        f.write('Bounds \n')
+        for v, vartype in vartypes.items():
+            if any([vv in not_allowed_chars for vv in v]):
+                raise ValueError('Cannot use chars {} in variable name'.format(not_allowed_chars))
+            if v[0] in not_first_char:
```
arcondello, line 1155: `""` and other 0-length objects are allowed as variable labels

dimod/constrained.py
```diff
@@ -1056,6 +1056,140 @@ def to_file(self, *, spool_size: int = int(1e9)) -> tempfile.SpooledTemporaryFil
...
+                        i_term += 1
+                        if i_term % new_line_every == 0:
+                            c_string += '\n '
+                    c_string += '] '
+
+                rhs = constraint.rhs - constraint.lhs.offset
+                c_string += ' {} {} \n'.format(sense_dict[constraint.sense.value], rhs)
+                f.write(c_string)
+
+        vartypes = {var: self.vartype(var) for var in self.variables}
```
arcondello, line 1147: I think you want to do
```python
vartypes = {repr(var): self.vartype(var) for var in self.variables}
if len(vartypes) != len(cqm.variables):
    raise ValueError("some variables do not have a unique representation")
```

test case for this:
```python
variables = [(0, 1), "(0, 1)"]
```

releasenotes/notes/add-CQM.to_lp_file-method-2a4a5fa0fc9156eb.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    Added ``to_lp_file`` method to dump ``ConstrainedQuadraticModel``
```
arcondello, line 4:
```
Add ``ConstrainedQuadraticModel.to_lp_file()`` method to encode constrained quadratic models as an LP file.
```

tests/test_constrained.py
```diff
@@ -949,3 +949,26 @@ def test_variable_multiple_times(self):
                 self.assertAlmostEqual(cmodel.lhs.get_linear('x0'), 3.5,
                                        msg='constraint c1, linear(x0) should be 3.5')
 
+    def test_functional(self):
```
arcondello, line 952: Need tests for:
- different variable labels sets, e.g. strings, tuples, frozensets
- empty CQM
- no constraints

arcondello: Also needs a test for the `a + b + 3 <= 0` and `a + b <= -3` cases.

dimod/constrained.py
```diff
@@ -1056,6 +1056,140 @@ def to_file(self, *, spool_size: int = int(1e9)) -> tempfile.SpooledTemporaryFil
...
+
+        f = SpooledTemporaryFile(max_size=spool_size, mode='wt')
+
+        if self.objective.linear or self.objective.quadratic:
+            f.write("Minimize\n ")
+            obj_string = 'obj: '
+
+            i_term = 0
+            if self.objective.linear:
+                for var, bias in self.objective.linear.items():
```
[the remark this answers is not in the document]
alexzucca90: yes, I can skip the variable so good point

dimod/constrained.py
```diff
@@ -1056,6 +1056,140 @@ def to_file(self, *, spool_size: int = int(1e9)) -> tempfile.SpooledTemporaryFil
...
+                    i_term += 1
+
+                    if i_term % new_line_every == 0:
+                        obj_string += '\n '
+
+            if self.objective.quadratic:
+                obj_string += '+ [ '
+                for edge, bias in self.objective.quadratic.items():
+                    var1, var2 = edge
+                    obj_string += '{} {} {} * {} '.format(sign_dict[np.sign(bias)], abs(2 * bias), var1, var2)
```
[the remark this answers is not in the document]
alexzucca90: because the LP file requires that the quadratic terms are within the parenthesis `[...]/2` and the `/2` is the required (only for the objective, not the constraints).

dimod/constrained.py
alexzucca90: so since the LP file doesn't allow that how do I check that?

arcondello: Ah, then just raise an informative error in that case.

arcondello:
By the way, depending on what LP files require in terms of variable labels, it might just be reasonable to throw an error if the CQM is not labelled in a way that makes sense.

dimod/constrained.py
alexzucca90: the only thing is that this raises an error when I look for the variable bounds because I'm calling the upper (or lower) bound method with a string

tests/test_constrained.py
alexzucca90: ok, tests for empty CQM, no constraints, only quadratic objective and constraint are in. The problem with tuples and forzensets as variables is that their representation contains characters that are not allowed by the LP format. In that case an error is raised, pointing to the variable that has bad characters.

dimod/constrained.py
alexzucca90: also,  the test case above, `variables = [(0, 1), "(0, 1)"]` does not apply to LP files because those variables contain forbidden names. But still that would give me two different representations of the tuples..

arcondello: Sure, if it's handled by the "invalid name" error then we're all good.

alexzucca90:
There is still work to do, especially in how the variable names are handled..

dimod/constrained.py
```diff
@@ -1056,6 +1056,153 @@ def to_file(self, *, spool_size: int = int(1e9)) -> tempfile.SpooledTemporaryFil
...
+        def string(bias: float) -> str:
+            return repr(abs(int(bias))) if bias.is_integer() else repr(abs(bias))
+
+        def sense(s: Sense) -> str:
+            return '=' if s.value == "==" else s.value
+
+        # forbidden chars for names in the LP format
+        not_allowed_chars = {'+', '-', '*', '^', ':'}
+        not_first_char = {'<', '>', '=', '(', ')', '[', ']', ','}
+        not_first_char.update({str(i) for i in range(10)})
```
arcondello, line 1085: doesn't really matter since it's small, but this creates an unnecessary set. You can do
```python
not_first_char.update((str(i) for i in range(10)))
```
or
```python
not_first_char.update(map(str, range(10))
```
or even
```python
not_first_char = {'<', '>', '=', '(', ')', '[', ']', ',', *map(str, range(10))}
```

arcondello:
Re variable handling, IMO the best approach is something like
```python
def to_lp_file(self, *,
               spool_size: int = int(1e9), 
               encoder: Optional[Callable[[Variable], str]] = None,
               ) -> tempfile.SpooledTemporaryFile:
    if encoder is None:
        if not all(isinstance(v, str) for v in self.variables):
            raise ValueError("unless an encoder is provided, all variables must be string-labelled, "
                             "If the variables are integer-labelled, using `encoder=str` will encode them as strings")

        def encoder(v: Variable) -> str:
            return v

    ...
```
and then add the `encoder=str` as an example in the docstring.

We could do something more automagical, but IMO it'll just create issues in the long run.

dimod/constrained.py
alexzucca90: third option is the best, one liner :)

arcondello: one more option that is nice and explicit and uses no fancy syntax is 
```python
not_first_char = set('<>=()[],0123456789')
```

alexzucca90: yes, that's correct, but the way I check the characters (first all chars and then the first) is equivalent to have them included.

dimod/constrained.py
```diff
@@ -1056,6 +1056,153 @@ def to_file(self, *, spool_size: int = int(1e9)) -> tempfile.SpooledTemporaryFil
...
+            return bias < 0
+
+        def string(bias: float) -> str:
+            return repr(abs(int(bias))) if bias.is_integer() else repr(abs(bias))
+
+        def sense(s: Sense) -> str:
+            return '=' if s.value == "==" else s.value
+
+        # forbidden chars for names in the LP format
+        not_allowed_chars = {'+', '-', '*', '^', ':'}
```
randomir, line 1083: These checks are not consistent with your LP file grammar: https://github.com/dwavesystems/dimod/blob/d36fbfe1d4cd6e1bdc1aa7ef9767f5ef500d19f5/dimod/serialization/lp.py#L40-L41

Also, since variable/constraint names/labels are unicode strings, I bet the lists provided here are not exhaustive (think: `π`, [nbsp](https://en.wikipedia.org/wiki/Non-breaking_space), or any other of 1M+ codepoints).

dimod/constrained.py
randomir: In combination with my comment above, I'd opt for regex-based filtering (so, no need to expand numbers).

arcondello:
Putting this here so I don't forget, the LP file format actually is pretty strict about variable labels, see http://www-eio.upc.edu/lceio/manuals/cplex-11/pdf/reffileformatscplex.pdf (page 9, section 4)
```
Variables can be named anything provided that the name does not exceed 255 characters,
all of which must be alphanumeric (a-z, A-Z, 0-9) or one of these symbols: ! " # $ % &
( ) , . ; ? @ _ ‘ ’ { } ~. Longer names are truncated to 255 characters. A variable name can
not begin with a number or a period.
The letter E or e, alone or followed by other valid symbols, or followed by another E or
e, should be avoided as this notation is reserved for exponential entries. Thus, variables
can not be named e9, E-24, E8cats, or other names that could be interpreted as an
exponent. Even variable names such as eels or example can cause a read error,
depending on their placement in an input line.
```

arcondello:
Closed in favor of https://github.com/dwavesystems/dimod/pull/1202

PR #1053, opened 2021-11-23
Fix coverage report in CI

arcondello:


.circleci/config.yml
```diff
@@ -237,13 +266,7 @@ jobs:
...
             cd tests/
             pip install -r requirements.txt
-            coverage run --source=dimod --append -m unittest
-      - run: &unix-codecov
-          name: codecov
-          command: |
-            . env/bin/activate
-            cd tests/
-            codecov
+            python -m unittest
```
randomir, line 269:
```suggestion
            python -m unittest discover
```

arcondello: Why?

randomir: Is `discover` implied? I see you use it below.

randomir: Oh, looks like it is implied. Ignore the suggestion then please.

arcondello: It is indeed. Agree I should be consistent.

PR #1054, opened 2021-11-24
Add ability to print quadratic models as a polynomial

arcondello:
~Todo: Quadratic model~

In the future we might consider using this as the default printing. Or at least allowing a global configuration to set that.

### DIFF
--- a/dimod/views/quadratic.py
+++ b/dimod/views/quadratic.py
@@ -16,8 +16,8 @@ import abc
 import io
 import operator
 
-from collections.abc import Callable, ItemsView, Iterable, Mapping, MutableMapping
-from typing import Any, Collection, Iterator, Optional, Tuple
+from collections.abc import ItemsView, Iterable, Mapping, MutableMapping
+from typing import Any, Callable, Collection, Iterator, Optional, Tuple
 
 from dimod.typing import Bias, Variable
 
@@ -339,6 +339,11 @@ class QuadraticViewsMixin(abc.ABC):
 
     @property
     @abc.abstractmethod
+    def offset(self) -> Bias:
+        raise NotImplementedError
+
+    @property
+    @abc.abstractmethod
     def variables(self) -> Collection[Variable]:
         raise NotImplementedError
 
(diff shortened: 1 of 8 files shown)


arcondello:
Especially interested in feedback on:
- the method name `to_polystring()` seems pretty arbitrary
- the default encoding. It made an arbitrary decision to leave strings as-is and prepend other things with `v`. We could instead just always use repr, which would give things like `5*'a'*(0, 'c')` or always prepend `5*va*v(0, 'c')` or both  `5*v'a'*v(0, 'c')`. Currently this will look like `5*a*v(0, 'c')` which IMO seems cleanest/clearest

dimod/views/quadratic.py
```diff
@@ -390,3 +395,90 @@ def set_linear(self, v: Variable, bias: Bias):
...
+    def to_polystring(self, encoder: Optional[Callable[[Variable], str]] = None) -> str:
+        """Return a string representing the model as a polynomial.
+
+        Args:
+            encoder: A function mapping variables to a string. By default
+                string variables are mapped directly whereas all other types
+                are mapped to a string :code:`f"v{variable!r}"`.
+
+        Returns:
+            A string representing the binary quadratic model.
```
JoelPasvolsky, line 414:
```suggestion
            String representing the quadratic model.
```

dimod/views/quadratic.py
```diff
@@ -390,3 +395,90 @@ def set_linear(self, v: Variable, bias: Bias):
...
+        def string(bias: float) -> str:
+            return repr(abs(int(bias))) if bias.is_integer() else repr(abs(bias))
+
+        def coefficient(bias: float) -> str:
+            return '' if abs(bias) == 1 else f"{string(bias)}*"
+
+        # linear variables that have a positive bias or are not going to be
+        # included implicitly in quadratic
+        linear = ((v, float(bias)) for v, bias in self.iter_linear()
+                  if bias or not any(bias for _, bias in self.iter_neighborhood(v)))
```
JoelPasvolsky, line 443: I guess you want to show all variables in the model but (1) typical polynomial representations do not include variables with zero coefficients and (2) if the string is used as input to some parsing code, it complicates it by having linear terms appear/disappear dependent on a quadratic bias (e.g., the linear z term in `'-x + 0*y + 0*z'` versus `'-x + 0*y - x*z'`)

arcondello: (1) I am not sure there is a "typical"
(2) We already handle this case in BQM/QM construction. E.g. `BQM({'x': -1, 'y': 0}, {'xz': -1}, 0.0, 'BINARY')` implicitly adds 'z'. That's why we only ignore linear variables when they appear in a quadratic term.

dimod/views/quadratic.py
```diff
@@ -390,3 +395,90 @@ def set_linear(self, v: Variable, bias: Bias):
     @abc.abstractmethod
     def set_quadratic(self, u: Variable, v: Variable, bias: Bias):
         raise NotImplementedError
+
+    def iter_linear(self) -> Iterator[Tuple[Variable, Bias]]:
+        """Iterate over the variables and their biases."""
+        get = self.get_linear
+        for v in self.variables:
+            yield v, get(v)
```
JoelPasvolsky, line 403: Just curious: what's the advantage of defining this `get` and using `.get_linear` over using `key, val in self.linear.items()` ?

arcondello: very minor performance boosts. This method is pretty low level so it makes sense to squeeze our the optimizations but not in general worth doing.
```python
In [1]: import dimod

In [2]: bqm = dimod.generators.gnp_random_bqm(500, 1, 'BINARY')

In [3]: def f():
   ...:     for k, v in bqm.linear.items(): pass
   ...: 

In [4]: def g():
   ...:     for v in bqm.variables:
   ...:         bqm.get_linear(v)
   ...: 

In [5]: def h():
   ...:     get = bqm.get_linear
   ...:     for v in bqm.variables:
   ...:         get(v)
   ...: 

In [6]: %timeit f()
204 µs ± 5.5 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [7]: %timeit g()
101 µs ± 717 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [8]: %timeit h()
84.9 µs ± 483 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

dimod/views/quadratic.py
JoelPasvolsky: I mean a mathematical representation rather than a printout of the QM data structure. My impression is that the intention here is to print the math format of a polynomial rather than a string with tided-up values of qm.linear  + qm.quadratic. Is that a mistaken understanding?

dimod/views/quadratic.py
JoelPasvolsky: perfect, thanks!

dimod/views/quadratic.py
arcondello: I am saying that there is no "math format". Writing a polynomial `E(x, y) = 2*x` or `E(x, y) = 2*x + 0*y` are both perfectly valid. I agree the latter looks a bit more redundant, but IMO is is better to err on the side of [explicitly](https://www.python.org/dev/peps/pep-0020/#id2) including a variable rather than droping it implicitly.

arcondello: To be fair - this case only arises with isolated variables with 0 linear bias. So really shouldn't show up in most use cases unless something is going wrong.

JoelPasvolsky: The usecase that comes to my mind is someone preloading variable labels, then in iterations adding biases/interactions and printing out the polynomial for verification, as part of self-checking or frontal training.
But that might not be the kind of usecase this was added for. No worries.

arcondello: I think that case it's even more important. For example:
```python
>>> bqm = BQM.empty('SPIN')
>>> bqm.offset = 1
>>> bqm.to_polystring()
'1'
>>> bqm.add_variable('a')
>>> bqm.to_polystring()
'1'
```
seems confusing to me. Whereas in
```python
>>> bqm.add_variable('a')
>>> bqm.to_polystring()
'1 + 0*a'
```
I can see that a change has happened.

JoelPasvolsky: That's a good usecase.
The one I had in mind is more like this:

```
>>> qm = QuadraticModel()
>>> qm.add_variables_from('INTEGER', ['a', 'b', 'c'])
>>> qm.to_polystring()
''
>>> qm.add_linear('a', -1)
>>> qm.to_polystring()
'-a'
>>> qm.add_linear('b', 2)
'-a + 2*b'
```

dimod/views/quadratic.py
```diff
@@ -390,3 +395,90 @@ def set_linear(self, v: Variable, bias: Bias):
...
+        def neg(bias: float) -> bool: return bias < 0
+
+        def string(bias: float) -> str:
+            return repr(abs(int(bias))) if bias.is_integer() else repr(abs(bias))
+
+        def coefficient(bias: float) -> str:
+            return '' if abs(bias) == 1 else f"{string(bias)}*"
+
+        # linear variables that have a positive bias or are not going to be
+        # included implicitly in quadratic
```
randomir, line 441: This is a nice touch. Output is clear, but not verbose.

randomir (approved):
LGTM!

Issue #1021, opened 2021-10-04
Revisit install dependency on dwave-preprocessing


randomir:
Circular dependency on dwave-preprocessing strikes again!

See https://github.com/dwavesystems/dwave-preprocessing/issues/28.

We need to revisit #999 / the idea from #989.

PR #1056, opened 2021-11-26
Drop dwave-preprocessing as a full dependency

arcondello:
I don't love dropping it with only a patch version increment
but I don't see another way to resolve the circular dependency
issues we're having. The error message at least is pretty
explicit and all of the ocean packages that rely on it have been
updated to also use dwave-preprocessing.

Closes https://github.com/dwavesystems/dimod/issues/1021


randomir (approved):
This should fix it. Thanks.

PR #1058, opened 2021-11-30
Fix bqm_structured BinaryQuadraticModelStructureError error

JoelPasvolsky:
fixes the unit test in https://github.com/dwavesystems/dwave-hybrid/pull/256

```
python -m unittest dwave-ocean-sdk/dwave-hybrid/tests/test_samplers.py 
.........................
----------------------------------------------------------------------
Ran 25 tests in 1.675s

OK
```

### DIFF
--- a/dimod/decorators.py
+++ b/dimod/decorators.py
@@ -188,7 +188,7 @@ def bqm_structured(f):
 
         if not adjacency.keys() >= bqm.variables:
             raise BinaryQuadraticModelStructureError(
-                f"given bqm contains variables, {(list(bqm.variables) - adjacency.keys())!r}, "
+                f"given bqm has at least one variable, {(list(bqm.variables) - adjacency.keys()).pop()!r}, "
                 "not supported by the structured solver")
         if any(u not in adjacency[v] for u, v, _ in bqm.iter_quadratic()):
             u, v = next((u, v) for u, v, _ in bqm.iter_quadratic() if u not in adjacency[v])
--- a/dimod/decorators.py
+++ b/dimod/decorators.py
@@ -188,7 +188,7 @@ def bqm_structured(f):
 
         if not adjacency.keys() >= bqm.variables:
             raise BinaryQuadraticModelStructureError(
-                f"given bqm contains a variable, {(adjacency.keys() - bqm.variables).pop()!r}, "
+                f"given bqm contains a variable, {(list(bqm.variables) - adjacency.keys())!r}, "
                 "not supported by the structured solver")
         if any(u not in adjacency[v] for u, v, _ in bqm.iter_quadratic()):
             u, v = next((u, v) for u, v, _ in bqm.iter_quadratic() if u not in adjacency[v])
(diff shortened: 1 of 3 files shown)


dimod/decorators.py
```diff
@@ -188,7 +188,7 @@ def structured_sample(sampler, bqm, **kwargs):
 
         if not adjacency.keys() >= bqm.variables:
             raise BinaryQuadraticModelStructureError(
-                f"given bqm contains a variable, {(adjacency.keys() - bqm.variables).pop()!r}, "
+                f"given bqm contains a variable, {(list(bqm.variables) - adjacency.keys())!r}, "
```
arcondello, line 191: I think you still want  `(list(bqm.variables) - adjacency.keys()).pop()` since the error message says "variable". Otherwise you want to say "variables".

arcondello:
Needs tests.

dimod/decorators.py
JoelPasvolsky: I think more than one variable is likely to be more common, so I'd go with variables

arcondello: I think that's true, but it's also less likely to be readable. Since it's likely to be quite a long list, potentially thousands. Whereas "there is at least one variable, x" provides enough info and is very easy to read.

arcondello:
I am not actually convinced that this line is tested at all. Though I see we're not sending the right info to codecov to find out. Will fix that...

randomir (approved):
LGTM.

Don't forget to add a release note.

dimod/decorators.py
randomir: Another option is to truncate the set repr to a fixed string length...

arcondello: :biking_man: :house: 

:laughing:

JoelPasvolsky:
> I am not actually convinced that this line is tested at all.

It's tested by dwave-hybrid :-)

releasenotes/notes/fix_bqm_structured_error-98a72320a8759168.yaml
```diff
@@ -0,0 +1,7 @@
+---
+
+fixes:
+  - |
+    Fixes the ``bqm_structured`` decorator's error message for cases where
```
arcondello, line 5: This is misleading. I think it should say something like ` ``bqm_structured`` decorator now correctly raises a ``BinaryQuadraticModelStructureError`` even when there are no shared variables.` or something like that. It's not the message that's fixed but the error type.

tests/test_decorators.py
```diff
@@ -48,6 +48,20 @@ def sample(self, bqm):
 class TestBQMIndexLabels(unittest.TestCase):
     pass
 
+class EmptyStructuredSampler(dimod.Structured):
```
arcondello, line 51: It doesn't really matter but I would either put these above `TestBQMIndexLabels` or down just above `TestBQMStructured`. I have a very weak preference for the latter.

JoelPasvolsky: I had meant to put them above ``TestBQMIndexLabels``. I moved them to above ``TestBQMStructured``, why not make you happy?

PR #1061, opened 2021-12-07
Add support for continuous variables in QuadraticModel, Python and C++

arcondello:
Still need sample set support and has some TODOs. Putting this work on the [continuous](https://github.com/dwavesystems/dimod/tree/continuous) branch for now, so I don't just make an enormous patch.

### DIFF
--- a/dimod/include/dimod/quadratic_model.h
+++ b/dimod/include/dimod/quadratic_model.h
@@ -30,7 +30,8 @@ namespace dimod {
 enum Vartype {
     BINARY,  ///< Variables that are either 0 or 1.
     SPIN,    ///< Variables that are either -1 or 1.
-    INTEGER  ///< Variables that are integer valued.
+    INTEGER,  ///< Variables that are integer valued.
+    REAL  ///< Variables that are real valued.
 };
 
 /**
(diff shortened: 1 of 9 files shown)


dimod/include/dimod/quadratic_model.h
```diff
@@ -937,31 +939,37 @@ class QuadraticModel : public QuadraticModelBase<Bias, Index> {
...
     index_type add_variable(Vartype vartype, bias_type lb, bias_type ub) {
         assert(lb <= ub);
-        assert(lb >= -this->max_integer());
-        assert(ub <= this->max_integer());
+
+        assert(vartype != Vartype::BINARY || lb == 0);
+        assert(vartype != Vartype::BINARY || ub == 1);
+
+        assert(vartype != Vartype::SPIN || lb == -1);
+        assert(vartype != Vartype::SPIN || ub == +1);
```
randomir, line 962: Seems redundant to require type _and_ bounds determined by the type.

randomir (approved):
LGTM

dimod/include/dimod/quadratic_model.h
arcondello: It's true, we could add a default version of this method.

PR #1074, opened 2022-01-10
Implement `BinaryArray`, `SpinArray`, and `IntegerArray` methods.

### LINKED ISSUE #1063: `Integers`, `Binaries`, `Spins` etc should have a flag to return an array
Something like
```python
In [13]: dimod.Binaries(range(5))
Out[13]: <generator object Binaries at 0x7f299a084ac0>

In [14]: dimod.Binaries(range(5), asarray=True)
Out[14]:
array([BinaryQuadraticModel({0: 1.0}, {}, 0.0, 'BINARY'),
       BinaryQuadraticModel({1: 1.0}, {}, 0.0, 'BINARY'),
       BinaryQuadraticModel({2: 1.0}, {}, 0.0, 'BINARY'),
       BinaryQuadraticModel({3: 1.0}, {}, 0.0, 'BINARY'),
       BinaryQuadraticModel({4: 1.0}, {}, 0.0, 'BINARY')], dtype=object)
```

We could also consider allowing the user to specify a shape.

For now, this can be done with
```python
In [14]: np.asarray(list(dimod.Binaries(range(5))))
Out[14]: 
array([BinaryQuadraticModel({0: 1.0}, {}, 0.0, 'BINARY'),
       BinaryQuadraticModel({1: 1.0}, {}, 0.0, 'BINARY'),
       BinaryQuadraticModel({2: 1.0}, {}, 0.0, 'BINARY'),
       BinaryQuadraticModel({3: 1.0}, {}, 0.0, 'BINARY'),
       BinaryQuadraticModel({4: 1.0}, {}, 0.0, 'BINARY')], dtype=object)

In [15]: np.asarray(list(dimod.Binaries(range(9)))).reshape(3, 3)
Out[15]: 
array([[BinaryQuadraticModel({0: 1.0}, {}, 0.0, 'BINARY'),
        BinaryQuadraticModel({1: 1.0}, {}, 0.0, 'BINARY'),
        BinaryQuadraticModel({2: 1.0}, {}, 0.0, 'BINARY')],
       [BinaryQuadraticModel({3: 1.0}, {}, 0.0, 'BINARY'),
        BinaryQuadraticModel({4: 1.0}, {}, 0.0, 'BINARY'),
        BinaryQuadraticModel({5: 1.0}, {}, 0.0, 'BINARY')],
       [BinaryQuadraticModel({6: 1.0}, {}, 0.0, 'BINARY'),
        BinaryQuadraticModel({7: 1.0}, {}, 0.0, 'BINARY'),
        BinaryQuadraticModel({8: 1.0}, {}, 0.0, 'BINARY')]], dtype=object)

```

hbarovertwo:
Implement `BinaryArray`, `SpinArray`, and `IntegerArray` methods.
Each method builds a numpy object array from their respective variable generator methods.

Closes #1063

### DIFF
--- a/dimod/binary/binary_quadratic_model.py
+++ b/dimod/binary/binary_quadratic_model.py
@@ -60,7 +60,7 @@ __all__ = ['BinaryQuadraticModel',
            'Float32BQM',
            'Float64BQM',
            'as_bqm',
-           'Spin', 'Binary', 'Spins', 'Binaries',
+           'Spin', 'Binary', 'Spins', 'Binaries', 'SpinArray', 'BinaryArray',
            'quicksum',
            ]
 
@@ -2106,10 +2106,10 @@ def BinaryArray(labels: Union[int, Iterable[Variable]],
         dtype: Data type for the returned binary quadratic models.
 
     Returns:
-        Binary quadratic models, each with a single binary variable.
+        Array of binary quadratic models, each with a single binary variable.
     
     """
-    return _VariableArray(Spins, labels, dtype)
+    return _VariableArray(Binaries, labels, dtype)
 
 
 @unique_variable_labels
@@ -2160,7 +2160,7 @@ def SpinArray(labels: Union[int, Iterable[Variable]],
         dtype: Data type for the returned binary quadratic models.
 
     Returns:
-        Binary quadratic models, each with a single spin variable.
+        Array of binary quadratic models, each with a single spin variable.
     
     """
     return _VariableArray(Spins, labels, dtype)
--- a/dimod/binary/binary_quadratic_model.py
+++ b/dimod/binary/binary_quadratic_model.py
@@ -2095,7 +2095,7 @@ def Binaries(labels: Union[int, Iterable[Variable]],
 
 def BinaryArray(labels: Union[int, Iterable[Variable]],
                 dtype: Optional[DTypeLike] = None) -> np.ndarray:
-    """Return a numpy array of binary quadratic models, each with a 
+    """Return a NumPy array of binary quadratic models, each with a 
     single binary variable.
 
     Args:
@@ -2149,7 +2149,7 @@ def Spins(labels: Union[int, Iterable[Variable]],
 
 def SpinArray(labels: Union[int, Iterable[Variable]],
               dtype: Optional[DTypeLike] = None) -> np.ndarray:
-    """Return a numpy array of binary quadratic models, each with a 
+    """Return a NumPy array of binary quadratic models, each with a 
     single spin variable.
 
     Args:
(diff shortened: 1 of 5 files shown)


dimod/binary/binary_quadratic_model.py
```diff
@@ -2092,6 +2093,23 @@ def Binaries(labels: Union[int, Iterable[Variable]],
         yield from (Binary(dtype=dtype) for _ in range(labels))
 
 
+def BinaryArray(labels: Union[int, Iterable[Variable]],
+                dtype: Optional[DTypeLike] = None) -> np.ndarray:
+    """Return a numpy array of binary quadratic models, each with a
```
arcondello, line 2098: nit: `NumPy`.

dimod/quadratic/quadratic_model.py
```diff
@@ -1224,5 +1224,37 @@ def Integers(labels: Union[int, Iterable[Variable]],
...
+
+
+def _VariableArray(variable_generator: Callable,
+                   labels: Union[int, Iterable[Variable]],
+                   dtype: Optional[DTypeLike] = None) -> np.ndarray:
+    """Builds numpy array from a variable generator method."""
+    if isinstance(labels, int):
+        number_of_elements = labels
+    else:
+        number_of_elements = len(labels)
```
arcondello, line 1252: This will currently fail for `labels = (v for v in range(5))` which is an iterable, but is not sized. One thing you could do is
```python
if isinstance(labels, int):
    number_of_elements = labels
elif isinstance(labels, Sized):
    number_of_elements = len(labels)
else:
    labels = list(labels)
    number_of_elements = len(labels) 
```
Dumps it to memory which is not ideal but I can't think of a nice way around it.

hbarovertwo: At first glance the edge case looked like tuple comprehension, but now I see that's actually done with `tuple(...)`, and would be sized. Thanks for pointing this out!

dimod/quadratic/quadratic_model.py
```diff
@@ -1224,5 +1224,41 @@ def Integers(labels: Union[int, Iterable[Variable]],
...
+def _VariableArray(variable_generator: Callable,
+                   labels: Union[int, Iterable[Variable]],
+                   dtype: Optional[DTypeLike] = None) -> np.ndarray:
+    """Builds NumPy array from a variable generator method."""
+    if isinstance(labels, int):
+        number_of_elements = labels
+    elif isinstance(labels, Sized):
+        number_of_elements = len(labels)
+    else:
+        labels = list(labels)
```
arcondello, line 1254: can we add a test for this case? Doesn't have to be for all of them, but at least one.

hbarovertwo: Yeah, will add that.

arcondello (approved):

PR #1076, opened 2022-01-13
Raise minimum Python version to 3.7 and minimum NumPy version to 1.19.1

arcondello:
Python 3.6 reached EOL Dec 2021

The NumPy minimum was chosen to still support https://github.com/dwavesystems/penaltymodel/issues/134

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,11 +30,6 @@ jobs:
       - restore_cache: &build-linux-restore-cache
           keys:
             - pip-{{ .Environment.CIRCLE_JOB }}-{{ checksum "pyproject.toml" }}
-      - when:
-          condition:
-            equal: [3.6.8, << parameters.python-version>>]
-          steps:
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -419,38 +401,25 @@ workflows:
             parameters:
               # test the lowest numpy version and the highest minor of each
               # currently deployed
-              numpy-version: [==1.17.3, ~=1.18.0, ~=1.19.0, ~=1.20.0, ~=1.21.0]
+              numpy-version: [==1.19.1, ~=1.20.0, ~=1.21.0, ~=1.22.0]
```
randomir, line 404: Perhaps include `~=1.19.0` as well?

randomir (approved):
LGTM!

.circleci/config.yml
arcondello: My reasoning is that NumPy is pretty trustworthy to not break backwards compatibility between 1.19.1 and ~=1.19.0 and it saves a bit of solve time not to test.

PR #1077, opened 2022-01-14
multiplication_circuit bugfix

mhlr:
Carry bit of the most significant column of a row need to feed add to the most significant column of the next row.


arcondello:
Can you add a test? No good deed goes unpunished. 

Also needs a release note.

mhlr:
> Can you add a test? No good deed goes unpunished.

@arcondello The problem is that the bug only manifests for problems that are too large for the exact solver and and then it is pretty hit or miss on the QPU. Not sure how to make a good test for it

arcondello:
My recommendation would be to generate a big one then feed some correct and incorrect solutions in and test that the energy is what you expect.

mhlr:
Test and note added

arcondello (approved):

PR #1079, opened 2022-01-17
Add C++ coverage to CI

arcondello:


arcondello:
I would like to also add Cython coverage, but I seem to be getting an error related to https://github.com/cython/cython/issues/3636. Probably because of the direct include.

hhtong (approved):

PR #1081, opened 2022-01-18
Fix QM.fix_variable when fixing the last variable in the model

arcondello:


mhlr:
LGTM/ Works on the factoriazation example. Thanks!

mhlr (approved):

PR #1082, opened 2022-01-18
Restore BQM.to_serializable and .from_serializable

arcondello:
We could also create an equivalent for quadratic models, but I am going to leave that alone for now.


randomir (approved):
lgtm

pau557 (approved):
LGTM

PR #1083, opened 2022-01-20
Fix QM.fix_variables to correctly update vartypes and bounds

arcondello:


hsadeghidw:
Here is an issue that I could find. 

`
x, y, z = (dimod.Binary(i) for i in range(3)
cqm = dimod.CQM()
cqm.add_discrete([0, 1, 2])
cqm.fix_variable(1, 1)
`

Now there is a discrete constraint left in cqm: x + z == 1. Given that I fixed y = 1, both x and z must be zero. Generally, constraints may become infeasible when we fix a variable. but in this case, we can trivially avoid that by simply acknowledging that the discrete constraint further fixes other variables.

arcondello:
Ah yes, that was deliberate. For the same reason we don't fix `j` in
```python
i, j = dimod.Integers('ij')
cqm.add_constraint(i + j == 5)
cqm.fix_variable('i', 3)
```
I would like to handle these sorts of cases, but it complicates things a lot, so I'd like to do it in a follow up feature and rely on server-side for now.

hsadeghidw:
I think because the discrete constraint is always so trivial and it behaves like a single variable, it should be handled.

arcondello:
Ok, I can add a `cascade=False` keyword argument to toggle the behavior, and just say in docs that it's not very smart.

hsadeghidw:
Alternatively, we can have an option to remove constraints, if that's not already a part of cqm. I'm not sure about "it's not very smart", I would argue that if a binary is set to 1, it would not be smart to keep the discrete constraint around. 

The use cases that I have in mind, all need the reduction of discrete constraints.

arcondello:
"It" here refers to the cascade algorithm,  "it is not very smart" as in it doesn't catch many other "trivial" cases, like `i+j=7` above. And invites questions for users why some and not others.

Also, it's dangerous because it'll remove variables from the model and therefore the sample set that you didn't ask it to, and it might not be obvious to users why. Whereas now it doesn't do anything that the user didn't explicitly ask for.

That said, I understand why it's convenient and useful, it just requires some thought about how to do it in a way that wont be surprising.

arcondello:
I will add this functionality in a followup PR, it's beyond the scope of this one.

PR #1088, opened 2022-01-28
Add QM.flip_variable() and CQM.flip_variable() methods

arcondello:


dimod/constrained.py
```diff
@@ -611,6 +611,21 @@ def fix_variables(self,
 
         return new
 
+    def flip_variable(self, v: Variable):
+        """Flip the specified variable in the objective and constraints.
+
+        Note that this may remove a constraint as being tracked as discrete.
+        Subsequently flipping the variable again will not restore it.
```
kevinchern, line 618: Should this potential perturbation of the original problem be tracked explicitly when the `flip_variable` method is invoked? Or return something to indicate a constraint was removed? e.g., a var `has_flipped` tracking whether `v` has ever been flipped, so subsequent calls can know a constraint has been removed; or `return True` if constraint was removed.

dimod/constrained.py
```diff
@@ -611,6 +611,21 @@ def fix_variables(self,
 
         return new
 
+    def flip_variable(self, v: Variable):
+        """Flip the specified variable in the objective and constraints.
+
+        Note that this may remove a constraint as being tracked as discrete.
```
jackraymond, line 617: This is a bit ugly, but I understand why its necessary after some thought and can't think of a workaround. Clear enough docstring.

dimod/binary/binary_quadratic_model.py
```diff
@@ -1018,11 +1018,20 @@ def energy(self, sample, dtype: Optional[DTypeLike] = None) -> Bias:
         energy, = energies
         return energy
 
-    def flip_variable(self, v: Hashable):
+    def flip_variable(self, v: Variable):
         """Flip the specified variable in a binary quadratic model."""
-        for u in self.adj[v]:
```
jackraymond, line 1023: Good catch.

kevinchern (commented):
just one question RE `flip_variable` in constrained.py

jackraymond (approved):
Looks good to me.

dimod/binary/binary_quadratic_model.py
arcondello: It actually worked before because we were modifying `bqm.spin` which is a "virtual" BQM of the SPIN vartype. E.g.
```python
In [5]: bqm = dimod.BQM('BINARY')

In [6]: bqm.spin.set_quadratic('a', 'b', -1)

In [7]: bqm
Out[7]: BinaryQuadraticModel({'a': 2.0, 'b': 2.0}, {('b', 'a'): -4.0}, -1.0, 'BINARY')
```
but that machinery is relatively expensive/slow which is why I changed it here.

dimod/constrained.py
arcondello: It's a good question. I thought about it and my instinct is no - simply because I don't really expect this case to happen very often. Further, I would expect any user who cares about that level of detail to also be sophisticated enough to track it themselves. Really this is just covering an edge case.

dimod/constrained.py
arcondello: In a followup PR I will be adding a
```python
>>> cqm.add_KinN_as_discrete(sum(Binary(v) for v in range(5)) == 3)  # returns the flipped variables
{0, 1}
```
because you can turn a KinN constraint into a discrete constraint with some variable flips. That feature is actually why I made this one and #1089

dimod/constrained.py
kevinchern: OK, makes sense.

kevinchern: now how do I add my first "LGTM"...

arcondello: Top right corner of "files changed", there is a "Review changes" button

kevinchern (approved):
And here it is: LGTM!

Issue #921, opened 2021-07-26
add_discrete using dimod.Binary raises error


hsadeghidw:
**Application**
When adding a set of binary variables as a discrete variable, an error is raised that indicates the variable is not hashable. This is not a bug because the function expects variable labels. However, if one is using dimod.Binary or dimod.Integer to build a model, it's convenient to be able to pass a set of such variables to construct a discrete variable.

This raises error,
```
import dimod
n = 10
cqm = dimod.CQM()
x = {}
for i in range(n):
    x[i] = dimod.Binary(i)
cqm.add_discrete([v for v in x.values()])
```

currently I can do:
```
import dimod
n = 10
cqm = dimod.CQM()
x = {}
for i in range(n):
    x[i] = dimod.Binary(i)
cqm.add_discrete([v.variables[0] for v in x.values()])
```
to get around it.

arcondello:
How would you expect 
```
x = Binary('x')
y = Binary('y')
z = Binary('z')

# 1
cqm.add_discrete([x+y, z+1])

# 2
cqm.add_discrete([2*x, y, z])
```
to behave? Would you expect it to fail for anything except zero offset, one variable with linear bias of 1?

hsadeghidw:
oh, that raises very good questions. I had not thought about it. For now, I think it should just raise an error.

arcondello:
FWIW numpy handles a similar case, `bool(np.asarray([0, 1]))`, with `ValueError: The truth value of an array with more than one element is ambiguous`.

arcondello:
At the very least we should probably update the docstring to make the current usage clearer (credit to @pau557 for the suggestion).

arcondello:
https://github.com/dwavesystems/dimod/pull/1089 adds support for
```python
x, y, z = dimod.Binaries('xyz')
cqm.add_discrete(sum((x, y, z)))
cqm.add_discrete(x + y + z)
cqm.add_discrete(x + y + z == 1)
```
and similar.

PR #1089, opened 2022-01-28
Support additional argument types in CQM.add_discrete()

arcondello:
This does not go quite as far as https://github.com/dwavesystems/dimod/issues/921 but it does allow for closer syntax.

### DIFF
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -388,15 +388,118 @@ class ConstrainedQuadraticModel:
         return self.add_constraint_from_model(
             qm, sense, rhs=rhs, label=label, copy=False)
 
-    def add_discrete(self, variables: Iterable[Variable],
-                     label: Optional[Hashable] = None) -> Hashable:
-        """Add an iterable of binary variables as a disjoint one-hot constraint.
+    def add_discrete(self, data, *args, **kwargs) -> Hashable:
+        """Add a one-hot constraint.
 
-        Adds a special kind of one-hot constraint. These one-hot constraints
+        Add a special kind of one-hot constraint. These one-hot constraints
         must be disjoint, that is they must not have any overlapping variables.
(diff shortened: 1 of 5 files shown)


dimod/constrained.py
```diff
@@ -408,34 +511,95 @@ def add_discrete(self, variables: Iterable[Variable],
...
+            if v in self.variables:
+                # it already exists, let's make sure it's not already used
+                if any(v in self.constraints[label].lhs.variables for label in self.discrete):
+                    raise ValueError(f"variable {v!r} is already used in a discrete variable")
+                if self.vartype(v) is not Vartype.BINARY:
+                    raise ValueError(f"variable {v!r} has already been added but is not BINARY")
+
+            bqm.set_linear(v, 1)
+
+        if bqm.num_variables < 2:
```
mhlr, line 542: This is checked in `add_discrete_from_model` which is transitively called from here

dimod/constrained.py
```diff
@@ -408,34 +511,95 @@ def add_discrete(self, variables: Iterable[Variable],
...
+            raise ValueError("discrete constraints must have at least two variables")
+
+        label = self.add_constraint_from_comparison(bqm == 1, label=label, copy=False)
+        self.discrete.add(label)
```
mhlr, line 546: This is checked in add_discrete_from_model which is transitively called from here

dimod/constrained.py
arcondello: Here we're actually calling `add_constraint_from_comparison` rather than `add_discrete_from_comparison`. So we do need these additional checks.

dimod/constrained.py
arcondello: See above

mhlr (approved):

PR #1092, opened 2022-01-31
Refresh binary_quadratic_model class's method's docstrings

JoelPasvolsky:


dimod/binary/binary_quadratic_model.py
```diff
@@ -601,18 +612,20 @@ def add_linear_equality_constraint(
...
+                by this value (which does not appear explicitly in the above
+                equation) when added to the binary quadratic model.
             constant:
-                The constant value of the constraint, :math:`C`, in the equation
-                above.
+                Value of the constant term, :math:`C`, of the linear constraint.
+
+        Examples:
+            >>> bqm = dimod.BinaryQuadraticModel("BINARY")
+            >>> bqm.add_linear_equality_constraint([("x1", 5), ("x2", -2)], 10, -3)
```
arcondello, line 628: this is deterministic, so I think you can safely show the output. Maybe with `.to_polystring()`?

dimod/binary/binary_quadratic_model.py
```diff
@@ -649,42 +662,39 @@ def add_linear_inequality_constraint(
     ) -> Iterable[Tuple[Variable, int]]:
         """Add a linear inequality constraint as a quadratic objective.
 
-        Adds a linear inequality constraint of the form:
-
-        math:'lb <= \sum_{i,k} a_{i,k} x_{i,k} + constant <= ub'
+        Adds a linear inequality constraint of the form,
```
arcondello, line 665: Should this be consistent with the first line? I.e. `Add` instead of `Adds`? Applies other places as well.

dimod/binary/binary_quadratic_model.py
```diff
@@ -840,8 +854,13 @@ def add_quadratic_from_dense(self, quadratic: ArrayLike):
         """Add quadratic biases from a square 2d array-like.
 
         Args:
-            quadratic:
-                An square 2d `array_like`_ of quadratic biases.
+            quadratic: Quadratic biases as a square 2d `array_like`_.
+
+        Examples:
+            >>> bqm = dimod.BinaryQuadraticModel("BINARY")
```
arcondello, line 860: IMO we should also include a SPIN model to show how the diagonal is handled.

Should also print the whole thing.

dimod/binary/binary_quadratic_model.py
```diff
@@ -899,11 +916,8 @@ def contract_variables(self, u, v):
         corresponding interactions of ``u``.
 
         Args:
-            u (variable):
-                Variable in the binary quadratic model.
-
-            v (variable):
-                Variable in the binary quadratic model.
+            u (variable): Variable in the binary quadratic model.
```
arcondello, line 919: Don't need `(variable)`, should instead update the type annotations.

dimod/binary/binary_quadratic_model.py
```diff
@@ -1007,6 +1033,13 @@ def energy(self, sample, dtype: Optional[DTypeLike] = None) -> Bias:
         Returns:
             The energy.
 
+        Examples:
+            >>> bqm = dimod.BinaryQuadraticModel({}, {("x0", "x1"): -1}, "BINARY")
+            >>> bqm.energy(([1, 1], ["x0", "x1"]))
```
arcondello, line 1038: IMO `{'x0': 1, 'x1': 1}` is clearer. I would leave this syntax for a discussion in `sample_like`.

dimod/binary/binary_quadratic_model.py
```diff
@@ -1232,7 +1265,36 @@ def from_numpy_matrix(cls, mat, variable_order=None, offset=0,
     @classmethod
     def from_numpy_vectors(cls, linear, quadratic, offset, vartype, *,
                            variable_order=None, dtype=np.float64):
-        """Create a binary quadratic model from NumPy vectors."""
+        """Create a binary quadratic model from NumPy vectors.
+
+        Args:
+            linear (NumPy vector): Linear biases.
```
arcondello, line 1271: Types should be added to the annotation, not the docstring.

dimod/binary/binary_quadratic_model.py
```diff
@@ -2145,7 +2356,12 @@ def to_serializable(self,
 
     def update(self, other: 'BinaryQuadraticModel'):
         """Add the variables, interactions, offset and biases from another
-        binary quadratic model."""
+        binary quadratic model.
+
+        Args:
+            other:
```
arcondello, line 2362: Given the type annotation, this one feels especially redundant.

dimod/binary/binary_quadratic_model.py
JoelPasvolsky: Okay, currently the examples show different formats, but they can show always the simplest

dimod/binary/binary_quadratic_model.py
JoelPasvolsky: We usually don't for the "more elaborate description" that follows the one-sentence summary line; e.g., 
https://docs.ocean.dwavesys.com/en/stable/docs_system/reference/generated/dwave.system.samplers.DWaveSampler.sample_ising.html. PEP 257 requires that the summary line prescribes as a command but doesn't say the same for multi-line docstings.
If the second use of the same verb jars, can rephrase

dimod/binary/binary_quadratic_model.py
JoelPasvolsky: @arcondello, have you tested the handling of diagonals for SPIN models? I don't understand the relation between the `pybqm.py` and `cybqm/cybqm_template.pyx.pxi` files but when you set a non-zero diagonal it triggers an exception from line https://github.com/dwavesystems/dimod/blob/3bf1ab194e6875cb7f4ac99ef75d4985f3360a4b/dimod/binary/cybqm/cybqm_template.pyx.pxi#L282 :

```
bqm.add_quadratic_from_dense([[0.2, -0.4],[0.1, -0.2]])

dimod\binary\cybqm\cybqm_template.pyx.pxi in dimod.binary.cybqm.cybqm_float64.cyBQM_template.add_quadratic_from_dense()

ValueError: 0 cannot have an interaction with itself
```

dimod/binary/binary_quadratic_model.py
JoelPasvolsky: This is for consistency: otherwise it looks like we missed descriptions for this method's params

dimod/binary/binary_quadratic_model.py
JoelPasvolsky: Did you mean to add the `.to_polystring()` method to BQMs too? It's now just on QMs.
I hadn't printed the BQM to keep the example short, but I can just print it the standard way.

dimod/binary/binary_quadratic_model.py
arcondello: I don't think we always need to show the simplest, but in this case I would consider the `(array_like, labels)` format to be a bit unintuitive, if a lot more performant.

JoelPasvolsky: Agree, I updated

dimod/binary/binary_quadratic_model.py
arcondello: Ah, I forgot what the exact behaviour is. But in this case we should mention that values on the diagonal raise an exception.

dimod/binary/binary_quadratic_model.py
arcondello: I am not sure anyone is checking them for consistency in that way, just whether they have useful information.

dimod/binary/binary_quadratic_model.py
JoelPasvolsky: Will add a `Raises`

dimod/binary/binary_quadratic_model.py
JoelPasvolsky: Not checking for consistency, but if a users has been through a few methods and then hits one with the parameters undescribed it looks like neglect. 
Another advantage is that the alternative is endless judgment calls on how self-descriptive is every parameter name, and we'll all differ on that
I try to keep them short in such cases to save the code file's real estate

dimod/binary/binary_quadratic_model.py
JoelPasvolsky: Changed definition to use type annotations

dimod/binary/binary_quadratic_model.py
JoelPasvolsky: Changed definition to use type annotations.

arcondello:
Current CI error is unrelated

dimod/binary/binary_quadratic_model.py
```diff
@@ -1230,9 +1263,40 @@ def from_numpy_matrix(cls, mat, variable_order=None, offset=0,
         return bqm
 
     @classmethod
-    def from_numpy_vectors(cls, linear, quadratic, offset, vartype, *,
-                           variable_order=None, dtype=np.float64):
-        """Create a binary quadratic model from NumPy vectors."""
+    def from_numpy_vectors(cls, linear: ArrayLike, quadratic: ArrayLike,
+                           offset: Number, vartype: Vartype, *,
```
arcondello, line 1267: Offset's type should be `float`, [Pep484](see https://www.python.org/dev/peps/pep-0484/#the-numeric-tower)

dimod/binary/binary_quadratic_model.py
```diff
@@ -860,50 +879,45 @@ def add_variable(self, v: Optional[Variable] = None, bias: Bias = 0):
...
-            v: Variable label. If not provided, the next interger label
+            v: Variable label. If not provided, the next integer label
                 is used.
-
             bias: Linear bias for the added variable.
         """
         return self.data.add_variable
 
-    def change_vartype(self, vartype, inplace=True):
+    def change_vartype(self, vartype: Optional[Vartype],
```
arcondello, line 888: `vartype` is not optional (not sure why the old docs claimed it was, probably a copy/paste error).

JoelPasvolsky: Fixed

dimod/binary/binary_quadratic_model.py
JoelPasvolsky: Fixed

dimod/binary/binary_quadratic_model.py
```diff
@@ -1435,51 +1532,59 @@ def iter_variables(self):
...
+            v:  Variable in the binary quadratic model.
+        """
         return self.data.get_quadratic
 
-    def normalize(self, bias_range=1, quadratic_range=None,
-                  ignored_variables=None, ignored_interactions=None,
-                  ignore_offset=False):
+    def normalize(self, bias_range: Union[Number, Iterable] = 1,
+                  quadratic_range: Union[Number, Iterable] = None,
+                  ignored_variables: Optional(Iterable) = None,
```
arcondello, line 1561: You mean to use square brackets here.

dimod/binary/binary_quadratic_model.py
```diff
@@ -1435,51 +1532,59 @@ def iter_variables(self):
...
+        Args:
+            u:  Variable in the binary quadratic model.
+            v:  Variable in the binary quadratic model.
+        """
         return self.data.get_quadratic
 
-    def normalize(self, bias_range=1, quadratic_range=None,
-                  ignored_variables=None, ignored_interactions=None,
-                  ignore_offset=False):
+    def normalize(self, bias_range: Union[Number, Iterable] = 1,
```
arcondello, line 1559: `bias_range: Union[float, Tuple[float, float]]`
`quadratic_range: Union[float, Tuple[float, float], None]`
`ignored_variables: Optional[Iterable[Variable]]`
`ignored_interactions: Optional[Iterable[Tuple[Variable, Variable]]`

JoelPasvolsky: Changed

arcondello (approved):

Issue #1091, opened 2022-01-31
bqm.resize should not return a value


JoelPasvolsky:
**Description**
(https://github.com/dwavesystems/dimod/blob/220a22ae53a422e04476bfeb5654e8c475120895/dimod/binary/binary_quadratic_model.py#L1600) :

```
return self.data.resize
```
returns a zero when this method is run. Why? 

**Steps To Reproduce**

```
bqm.resize(3)
0
```

**Expected Behavior**
Quietly do its work.

**Environment**
 - OS: WIN10
 - Python version: 3.6

**Additional Context**
I'm in that file now, can I delete the `return` ?

arcondello:
We need to return a value in order to be able to raise an exception. See [error return values](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#error-return-values).

JoelPasvolsky:
What about returning the difference in number of variables between current and previous BQM rather than 0?

arcondello:
Also, to be clear, https://github.com/dwavesystems/dimod/blob/220a22ae53a422e04476bfeb5654e8c475120895/dimod/binary/binary_quadratic_model.py#L1600 does not return the value of zero, it is returning the method to be used. See https://github.com/dwavesystems/dimod/blob/220a22ae53a422e04476bfeb5654e8c475120895/dimod/decorators.py#L406

The value is returned by https://github.com/dwavesystems/dimod/blob/220a22ae53a422e04476bfeb5654e8c475120895/dimod/binary/cybqm/cybqm_template.pyx.pxi#L713 which is a `cpdef` function which is why we need that return value.

arcondello:
> What about returning the difference in number of variables between current and previous BQM rather than 0?

Could just be the size of the resulting BQM (since -1 is what indicates an error). It's a bit redundant since that's the value being passed in but at least would be a bit more meaningful :shrug:

arcondello:
> What about returning the difference in number of variables between current and previous BQM rather than 0?

Changed my mind, your idea is better. We can just error on 0 instead since that should be an uncommon case if we're returning the delta.

PR #1093, opened 2022-01-31
Change BQM.resize() to return the size difference

arcondello:
Previously it always returned 0.

Closes https://github.com/dwavesystems/dimod/issues/1091


JoelPasvolsky (approved):
Thank you!

JoelPasvolsky:
Just a thought: if someone sets `n` to `bqm.num_variables` will it give an exception now rather than do nothing?

arcondello:
No, it checks whether an exception has been thrown in that case.

Issue #1080, opened 2022-01-18
Propagate min and max integer algorithmically


arcondello:
Right now this number can be access with
```python
import dimod

dimod.Integer('i').upper_bound('i')
```
but that is quite clunky.

PR #1094, opened 2022-02-01
Expose variable type bounds at the Python level

arcondello:
Closes https://github.com/dwavesystems/dimod/issues/1080

### DIFF
--- a/dimod/include/dimod/quadratic_model.h
+++ b/dimod/include/dimod/quadratic_model.h
@@ -33,6 +33,89 @@ enum Vartype {
     INTEGER  ///< Variables that are integer valued.
 };
 
+/// Compile-time limits by variable type.
+template <class Bias, Vartype vartype>
+class vartype_limits {};
+
+template <class Bias>
+class vartype_limits<Bias, Vartype::BINARY> {
+ public:
+    static constexpr Bias default_max() noexcept { return 1; }
+    static constexpr Bias default_min() noexcept { return 0; }
+    static constexpr Bias max() noexcept { return 1; }
+    static constexpr Bias min() noexcept { return 0; }
+};
+
+template <class Bias>
+class vartype_limits<Bias, Vartype::SPIN> {
+ public:
+    static constexpr Bias default_max() noexcept { return 1; }
+    static constexpr Bias default_min() noexcept { return -1; }
+    static constexpr Bias max() noexcept { return +1; }
+    static constexpr Bias min() noexcept { return -1; }
+};
+
+template <class Bias>
+class vartype_limits<Bias, Vartype::INTEGER> {
+ public:
+    static constexpr Bias default_max() noexcept { return max(); }
+    static constexpr Bias default_min() noexcept { return 0; }
+    static constexpr Bias max() noexcept {
+        return ((std::int64_t)1 << (std::numeric_limits<Bias>::digits)) - 1;
+    }
+    static constexpr Bias min() noexcept { return -max(); }
+};
+
+/// Runtime limits by variable type.
+template <class Bias>
+class vartype_info {
+ public:
+    static Bias default_max(Vartype vartype) {
+        if (vartype == Vartype::BINARY) {
+            return vartype_limits<Bias, Vartype::BINARY>::default_max();
+        } else if (vartype == Vartype::SPIN) {
+            return vartype_limits<Bias, Vartype::SPIN>::default_max();
+        } else if (vartype == Vartype::INTEGER) {
+            return vartype_limits<Bias, Vartype::INTEGER>::default_max();
+        } else {
+            throw std::logic_error("unknown vartype");
+        }
+    }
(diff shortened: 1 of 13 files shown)


dimod/include/dimod/quadratic_model.h
```diff
@@ -33,6 +33,89 @@ enum Vartype {
...
 };
 
+/// Compile-time limits by variable type.
+template <class Bias, Vartype vartype>
+class vartype_limits {};
+
+template <class Bias>
+class vartype_limits<Bias, Vartype::BINARY> {
+ public:
+    static constexpr Bias default_max() noexcept { return 1; }
```
hbarovertwo, line 43: Unlike `Integer` variables, having a `default_min` and `default_max` seems a little off for both `Binary` and `Spin` variables. It seems to imply that there are non-defaults options that are allowed.

arcondello: Agree it's a bit verbose, but it lets us write more generic code. For instance https://github.com/dwavesystems/dimod/pull/1094/files#diff-314bf3065935756ec61ae283725d32f27810d08fb64dd440c4195b17fc2e390eR1076 lower down.
```c++
        return this->add_variable(
                vartype, vartype_info<bias_type>::default_min(vartype),
                vartype_info<bias_type>::default_max(vartype));
```

arcondello: This is overall following the pattern of [numeric_limits](https://en.cppreference.com/w/cpp/types/numeric_limits), which for instance also supports `bool` even though most of the values are just 0/1

dimod/quadratic/quadratic_model.py
```diff
@@ -1226,6 +1226,25 @@ def vartype(self, v: Variable) -> Vartype:
         """The variable type of the given variable."""
         return self.data.vartype
 
+    def vartype_info(self, vartype: VartypeLike):
```
hbarovertwo, line 1229: Overall the code works as intended. I'm not sure if there is a clean way to define the method downstream such that it doesn't have to live in `QuadraticModel`. But maybe we can use the `@staticmethod` decorator here so that we don't have to instantiate a QM just to check on variable type information.

tests/test_quadratic_model.py
```diff
@@ -685,6 +685,39 @@ def test_simple(self):
         self.assertEqual((-i*j - x).to_polystring(), '-x - i*j')
 
 
+class TestVartypeInfo(unittest.TestCase):
+    @parameterized.expand([(np.float32,), (np.float64,)])
```
hbarovertwo, line 689: Looking at the tests, maybe the static method suggestion doesn't hold. Unless it wouldn't be much trouble to optionally pass a `dtype` arg to `vartype_info` itself, mainly for the case of the `Integer` vartypes.

dimod/quadratic/quadratic_model.py
hbarovertwo: The way it is currently implemented, I can do:
```
import dimod
dimod.Integer('i').vartype_info(dimod.INTEGER)
```
Which feels equally as clunky as the original issue, although I get a lot more information packaged together. 
The behavior is quite different as what is demonstrated in the issue itself, since the `lower_bound` and `upper_bound` methods would actually return the set bounds on the integer variable, where `vartype_info` gives static information that doesn't depend on an instantiated variable.

It also feels a little clunky that I can use an instantiated QM, and by extension an instantiated `Integer` to check vartype info, but the same is not true for BQMs/Spin/Binary instances.

tests/test_quadratic_model.py
arcondello: Yes, that was the reasoning. But now that you mention it, perhaps I should be thinking of this differently and just make it a function similar to the C++ level. Something like
```python
def vartype_info(dtype, vartype):
    ...
```
will consider

releasenotes/notes/c++vartype-info-ae0bf7b4ca0a82bd.yaml
```diff
@@ -0,0 +1,10 @@
+---
+features:
+  - |
+    Add ``dimod::vartype_limits`` to C++ code for compile-time information
+    about variable types.
+  - |
+    Add ``dimod::vartype_info`` to C++ code for runtime information
+    about variable types.
+  - |
+    Add ``vartype_info()`` function.
```
hbarovertwo, line 10: no newline at EOF here

hbarovertwo (approved):
This looks great now!

releasenotes/notes/c++vartype-info-ae0bf7b4ca0a82bd.yaml
arcondello: good catch, but since this is a release note file we'll never touch it again (since it's position in the git history is how reno determines when it was released) so it doesn't matter

Issue #1060, opened 2021-12-06
Add documentation for single variable aliases


arcondello:
Right now `Spin`, `Spins`, `Binary`, `Binaries`, `Integer` and `Integers` are not included in the docs.

arcondello:
And now also https://github.com/dwavesystems/dimod/pull/1074

PR #1095, opened 2022-02-01
Refresh binary_quadratic_model single-var models' docstrings

JoelPasvolsky:
This PR continues updating https://github.com/dwavesystems/dimod/tree/main/dimod/bqm. It adds the single-var models, and so closes https://github.com/dwavesystems/dimod/issues/1060, but I'll return to the new *.rst pages later, so those pages currently are minimally updated. We can review those in a following PR. I also have not updated the QM single-var models, just rendered them.


dimod/binary/binary_quadratic_model.py
```diff
@@ -2454,13 +2462,18 @@ def BinaryArray(labels: Union[int, Iterable[Variable]],
...
         dtype: Data type for the returned binary quadratic models.
 
     Returns:
-        Array of binary quadratic models, each with a single binary variable.
+        Array of :class:`.BinaryQuadraticModel`\ s, each representing a single
+        binary variable.
 
+    Examples:
+        >>> x = dimod.BinaryArray(["x0", "x1"])
+        >>> x[1]
```
arcondello, line 2475: Again, I think this should just point to symbolic, but if we must put an example here why not just display `x`? Or better yet
```python
>>> x = dimod.BinaryArray(["x0", "x1"])
>>> Q = [[1, 2], [0, 3]]
>>> x.dot(Q).dot(x)
BinaryQuadraticModel({'x0': 1.0, 'x1': 3.0}, {('x1', 'x0'): 2.0}, 0.0, 'BINARY')
```
I don't think this example conveys what this function is useful for very well.

docs/intro/intro_symbolic_math.rst
```diff
@@ -0,0 +1,85 @@
...
+=============
+
+dimod enables easy incorporation of binary and integer variables as
+:ref:`single-variable models <symbolic_math_generators>`. For example, you can
+represent such binary variables as follows:
+
+>>> from dimod import Binary, Spin
+>>> x = Binary('x')
+>>> s = Spin('s')
+>>> x
```
arcondello, line 14: Overall this seems like a place that we should be using the `to_polystring()` representations rather than printing the models.

docs/intro/intro_symbolic_math.rst
```diff
@@ -0,0 +1,85 @@
...
+
+Example: Adding Models
+======================
+
+This example uses the performant :func:`~dimod.quicksum` to add multiple models.
+
+>>> import random
+>>> from dimod import Binary, quicksum
+>>> vars = ['x'+str(i) for i in range(5)]
+>>> models = [Binary(var, bias=random.randint(0,5)) for var in vars]
```
arcondello, line 84:
```python
>>> models = [v*random.randint(0, 5) for v in dimod.Binaries(variables)]
```

docs/intro/intro_symbolic_math.rst
```diff
@@ -0,0 +1,85 @@
...
+>>> cqm.add_constraint(i + j >= 1)                 # doctest: +IGNORE_RESULT
+
+Example: Adding Models
+======================
+
+This example uses the performant :func:`~dimod.quicksum` to add multiple models.
+
+>>> import random
+>>> from dimod import Binary, quicksum
+>>> vars = ['x'+str(i) for i in range(5)]
```
arcondello, line 83:
```python
>>> variables = [f'x{i}' for i in range(5)]
```
Should use `variables` since [vars](https://docs.python.org/3/library/functions.html#vars) is a built-in function.

docs/intro/intro_symbolic_math.rst
arcondello: Can also just do `dimod.BinaryArray(range(5))*np.random.randint(0, 5, size=5)` which IMO is the flow we actually want to be encouraging.

arcondello: I think showing both preferred patterns:
```python
In [14]: num_variables = 100

In [15]: %timeit (dimod.BinaryArray(range(num_variables))*np.random.randint(0, 5, size=num_variables)).sum()
2.11 ms ± 11.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [16]: %timeit dimod.quicksum(v*np.random.randint(0, 5) for v in dimod.Binaries(range(num_variables)))
2.5 ms ± 33.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
would be good (don't need the timing info, but it's interesting)

arcondello: For posterity, the latter ends up being faster
```
In [17]: num_variables = 5000

In [18]: %timeit (dimod.BinaryArray(range(num_variables))*np.random.randint(0, 5, size=num_variables)).sum()
287 ms ± 1.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [19]: %timeit dimod.quicksum(v*np.random.randint(0, 5) for v in dimod.Binaries(range(num_variables)))
128 ms ± 1.43 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

docs/intro/intro_symbolic_math.rst
```diff
@@ -0,0 +1,85 @@
...
+BinaryQuadraticModel({'x': 1.0}, {}, 0.0, 'BINARY')
+
+Similarly for integers:
+
+>>> from dimod import Integer
+>>> i = Integer('i')
+>>> i
+QuadraticModel({'i': 1.0}, {}, 0.0, {'i': 'INTEGER'}, dtype='float64')
+
+The construction of such variables as either BQMs or QMS depends on the type of
```
arcondello, line 24: Specifically, anything with more than one variable type or with integers will be a quadratic model.

dimod/binary/binary_quadratic_model.py
JoelPasvolsky: I love this idea. 
I wanted to keep the example code as short as possible but do think that because we know this concept is unintuitive it's worth keeping what it does in front of users. If you're okay with the extra bit of space, something like this is much better.

docs/intro/intro_symbolic_math.rst
JoelPasvolsky: Do we still consider these as useful primarily for testing and small problems or have they begun moving into (high) performance now that we have also `BinaryArray`, `quicksum`, maybe the older single-var BQMs are more efficient too?

docs/intro/intro_symbolic_math.rst
JoelPasvolsky: This one is to show the generated BQM but definitely a better output for `x + s` and such that use the variables in some formula

dimod/binary/binary_quadratic_model.py
arcondello: I am ok with the extra space, though still mildly prefer pointing everyone to one larger example in a centralized space.

docs/intro/intro_symbolic_math.rst
arcondello: We still don't encourage any symbolic for true performance, but with quicksum etc the performance can be OK.

docs/reference/constrained.rst
```diff
@@ -73,3 +80,5 @@ Converting constrained quadratic models to other model types:
    :toctree: generated/
 
    cqm_to_bqm
+
+.. currentmodule:: dimod.sym
```
JoelPasvolsky, line 84: Deleted but not pushed

dimod/binary/binary_quadratic_model.py
```diff
@@ -2454,13 +2462,34 @@ def BinaryArray(labels: Union[int, Iterable[Variable]],
...
+            models. If a number, labels are generated using :class:`uuid.UUID`.
         dtype: Data type for the returned binary quadratic models.
 
     Returns:
-        Array of binary quadratic models, each with a single binary variable.
+        Array of :class:`.BinaryQuadraticModel`\ s, each representing a single
+        binary variable.
+
+    Examples:
+        This example generates the QUBO, :math:`{x}^{T} {Q}{x}`, for a Boolean
```
arcondello, line 2474: I like this example, I think `Binaries` and `Binary` should get a similar treatment.

JoelPasvolsky: Thanks, that's the plan for the next push.

JoelPasvolsky: Done

dimod/binary/binary_quadratic_model.py
```diff
@@ -2439,8 +2443,12 @@ def Binaries(labels: Union[int, Iterable[Variable]],
         dtype: Data type for the returned binary quadratic models.
 
     Yields:
-        Binary quadratic models, each with a single binary variable.
+        A :class:`.BinaryQuadraticModel` for each binary variable.
 
+    Examples:
+        >>> x = dimod.Binaries(["x0", "x1"])
```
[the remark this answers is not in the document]
JoelPasvolsky: Replaced with new example

dimod/binary/binary_quadratic_model.py
```diff
@@ -2423,8 +2423,12 @@ def Binary(label: Optional[Variable] = None, bias: Bias = 1,
         dtype: Data type for the returned binary quadratic model.
 
     Returns:
-        Instance of :class:`.BinaryQuadraticModel`.
+        A :class:`.BinaryQuadraticModel` representing a binary variable.
 
+    Examples:
```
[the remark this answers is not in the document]
JoelPasvolsky: Replaced example

JoelPasvolsky:
Link checking is failing on 
```
/home/circleci/project/docs/README.rst.rst: WARNING: broken link: https://codecov.io/gh/dwavesystems/dimod (('Connection aborted.', OSError(0, 'Error')))
/home/circleci/project/docs/README.rst.rst: WARNING: broken link: https://codecov.io/gh/dwavesystems/dimod/branch/main/graph/badge.svg (('Connection aborted.', OSError(0, 'Error')))
make: *** [Makefile:25: linkcheck] Error 1
```
Might work if I rerun

JoelPasvolsky:
yup, passed on second run

dimod/binary/binary_quadratic_model.py
```diff
@@ -2423,8 +2423,17 @@ def Binary(label: Optional[Variable] = None, bias: Bias = 1,
...
+        This example generates a BQM to represent the polynomial,
+        :math:`x^2 + 3x -1.5`, where :math:`x` is a binary variable with values
+        :math:`\{0, 1\}`. Remember that for such variables, :math:`x^2 = x`.
```
arcondello, line 2431: Much better. Though I am surprised that you wanted to introduce the `x^2 == x` thing here. I think just `3*x - 1.5` would be more than sufficient.

dimod/binary/binary_quadratic_model.py
```diff
@@ -2439,8 +2448,25 @@ def Binaries(labels: Union[int, Iterable[Variable]],
...
+        :std:doc:`Ocean documentation's <oceandocs:index>`
+        :ref:`NOT example <oceandocs:not>`. The output of the brute-force solver,
+        :class:`~dimod.reference.samplers.ExactSolver`, shows that best
+        solutions are for assignments of the variables where :math:`z = \overline{x}`.
+
+        >>> x, y = dimod.Binaries(["x", "y"])
+        >>> bqm = 2*x*y - x - y
+        >>> print(bqm.to_polystring())
+        -x - y + 2*x*y
+        >>> print(dimod.ExactSolver().sample(bqm).lowest())
```
arcondello, line 2465: I think just the first three lines are sufficient. But OK with keeping this as well.

dimod/binary/binary_quadratic_model.py
```diff
@@ -2477,8 +2524,17 @@ def Spin(label: Optional[Variable] = None, bias: Bias = 1,
...
-        Instance of :class:`.BinaryQuadraticModel`.
+        A :class:`.BinaryQuadraticModel` representing a spin-valued binary variable.
+
+    Examples:
+        This example generates a BQM to represent the polynomial,
+        :math:`s^2 + 3s -1.5`, where :math:`s` is a binary variable with spin values
+        :math:`\{-1, 1\}`. Remember that for such variables, :math:`s^2 = 1`.
 
+        >>> s = dimod.Spin('s')
+        >>> bqm = s*s + 3*s - 1.5
```
arcondello, line 2535: Similar, I think just `3*s - 1.5` is sufficient here.

arcondello (approved):
A few final comments, but I am OK with it as is if you're feeling sick of looking at it

dimod/binary/binary_quadratic_model.py
arcondello: Also, as a fancy example, I love `(1-x)*(1-y)` as a way to show off everything symbolic in one line.

JoelPasvolsky:
Sick of looking at it? Just getting warmed up.

dimod/binary/binary_quadratic_model.py
JoelPasvolsky: Yes,  lovely, `(1-x)*(1-y)` I want to keep for the intro

dimod/binary/binary_quadratic_model.py
JoelPasvolsky: I deliberated on that but saw a couple of advantages: (1) it makes the example's output more interesting (2) it shows/reminds users of an extremely important property of binary variables that someone who isn't immersed might miss. I figured it might save a few Support questions.
I can remove if you think it distracts in a harmful way.

arcondello: I've thought about it more and I do think it distracts. When I see examples I usually just skip the the code rather than reading the text. I think in this case we want the code to be extremely self-explanatory. The self-loops thing is an excellent topic for the "gotchas" section of the more full symbolic docs

docs/intro/intro_symbolic_math.rst
```diff
@@ -78,18 +86,14 @@ and constraints (:math:`xj <= 3` and :math:`i + j >= 1`) in a simple CQM.
...
+:func:`~dimod.binary.BinaryArray` to add multiple models.
 
->>> import random
->>> from dimod import Binary, quicksum
->>> vars = ['x'+str(i) for i in range(5)]
->>> models = [Binary(var, bias=random.randint(0,5)) for var in vars]
+>>> import numpy as np
+>>> from dimod import BinaryArray, quicksum
+...
+>>> num_vars = 10; max_bias = 5
```
arcondello, line 95: `num_variables`, and pep8 doesn't like single line

docs/intro/intro_symbolic_math.rst
```diff
@@ -78,18 +86,14 @@ and constraints (:math:`xj <= 3` and :math:`i + j >= 1`) in a simple CQM.
...
+>>> var_labels = range(num_vars)
+...
+>>> models = BinaryArray(var_labels)*np.random.randint(0, max_bias, size=num_vars)
```
arcondello, line 98: how about `array = `? or to be consistent with NumPy, `arr`

docs/intro/intro_symbolic_math.rst
arcondello: Or this can be part of the followup pass.

arcondello (approved):

PR #1100, opened 2022-02-07
Refresh core files composite, initialized, polysampler

JoelPasvolsky:


dimod/core/polysampler.py
```diff
@@ -51,55 +54,54 @@ def sample_poly(self, polynomial, **kwargs):
         """Sample from a higher-order polynomial."""
         pass
 
-    def sample_hising(self, h, J, **kwargs):
+    def sample_hising(self, h: dict, J: dict, **kwargs) -> SampleSet:
```
JoelPasvolsky, line 57: Should this be `Dict[Tuple[Variable, Variable], Bias]` ?

dimod/core/initialized.py
```diff
@@ -19,16 +19,21 @@
...
 
 """
+from __future__ import annotations
 
 import abc
 
 from collections import namedtuple
 from numbers import Integral
 
+from typing import Optional, Tuple
```
arcondello, line 29: Ok, I know I do this everywhere so it's very hypocritical of me, but recently I've starting just `import typing` and doing `typing.Tuple` etc everywhere. In some files like `binary_quadratic_model.py` it really makes sense to import at least `Union`, `Tuple`, and `Optional`, but in a file like this I'd rather keep the namespace clean if that makes sense. This is still something that I am trying to build best practices around though so forgive me if I change around.

dimod/core/initialized.py
```diff
@@ -47,31 +52,28 @@ class Initialized(abc.ABC):
...
     # dev note: if this function is updated, make sure to also update
     # IdentitySampler
-    def parse_initial_states(self, bqm,
-                             initial_states=None,
-                             initial_states_generator='random',
-                             num_reads=None, seed=None, copy_always=False):
-        """Parses/generates initial states for an initialized sampler.
+    def parse_initial_states(self, bqm: BinaryQuadraticModel,
+                             initial_states = None,
+                             initial_states_generator: str = 'random',
```
arcondello, line 57: If we want to be fancy, we should use `typing.Literal`, but since that arrived in Python 3.8 we could do at the top
```python
try:
    InitialStateGenerators = typing.Literal['random', ...
except ImportError:
    InitialStateGenerators = str
```

dimod/core/polysampler.py
```diff
@@ -17,11 +17,14 @@
 not constrained to quadratic interactions.
 
 """
+from __future__ import annotations
```
arcondello, line 20: Don't think we need this here.

dimod/core/polysampler.py
arcondello: That would be more specific, so yes.

dimod/core/initialized.py
JoelPasvolsky: @arcondello I think for this to not fail CI Py 3.7 it needs also something like this:
```
from typing import TYPE_CHECKING
if TYPE_CHECKING:             # not entered at runtime
    from typing import Literal
```

arcondello: `from typing import Literal` will fail in 3.7 always. It will just silently fail when mypy runs in `TYPE_CHECKING` mode
the `if TYPE_CHECKING:` guard only affects runtime vs not.

So you can do
```python
if TYPE_CHECKING:
    try:
        InitialStateGenerators = typing.Literal['random', ...
    except ImportError:
        InitialStateGenerators = str
```
if you want to do both but :shrug: IMO that's overkill.

arcondello: Correction, you need `except AttributeError:`.

dimod/core/polysampler.py
```diff
@@ -51,55 +57,44 @@ def sample_poly(self, polynomial, **kwargs):
         """Sample from a higher-order polynomial."""
         pass
 
-    def sample_hising(self, h, J, **kwargs):
+    def sample_hising(self, h: typing.Dict[Variable, Bias],
```
arcondello, line 60: also, strictly speaking this should be `typing.Mapping` rather than `typing.Dict`. E.g. `bqm.linear` is a mapping but not a dict.

arcondello: Though if you return a dict, you _should_ use `typing.Dict`, since it's more specific. Isn't typing fun?

dimod/core/initialized.py
```diff
@@ -19,26 +19,31 @@
...
 import numpy as np
 
 from dimod.sampleset import as_samples, infer_vartype, SampleSet
 from dimod.vartypes import Vartype
+from dimod.binary.binary_quadratic_model import BinaryQuadraticModel
 
-__all__ = ['Initialized']
-
+try:
+    InitialStateGenerators = typing.Literal['none','tile', 'random']
```
arcondello, line 36: Missing a space. Also probably should be `InitialStateGenerator` (I know that I had it plural in my example).

PR #1107, opened 2022-02-09
Add BinaryQuadraticModel.maximum_energy_delta() method

seatim:
### Summary

Add a method to `BinaryQuadraticModel` that may be generally useful for constructing penalty models.

### Background

This method is defined as a function in the [Mutual Information Feature Selection example](https://github.com/dwave-examples/mutual-information-feature-selection/) ([here](https://github.com/dwave-examples/mutual-information-feature-selection/blob/master/titanic.py#L90)).

### Alternatives considered

Defining `maximum_energy_delta` as a free function elsewhere in `dimod` (e.g. in `utilities.py` and in `generators/constraints.py`) was considered.  It seemed out of place elsewhere, and on the other hand, a natural fit in `BinaryQuadraticModel`, despite its specialized nature.

Adding `maximum_energy_delta` to another class (e.g. `QuadraticModel`) was also considered.  There may be some value in extending this functionality to that class in the future (i.e. supporting integer variables in the calculation).  But there is less utility in doing this, because problems with integer variables are naturally modeled using `ConstrainedQuadraticModel`.  And a penalty model is not needed in that case as constraints can be represented natively.

### DIFF
--- a/dimod/binary/binary_quadratic_model.py
+++ b/dimod/binary/binary_quadratic_model.py
@@ -1111,6 +1111,37 @@ class BinaryQuadraticModel(QuadraticViewsMixin):
         energy, = energies
         return energy
 
+    def maximum_energy_delta(self) -> Bias:
+        """Compute a conservative bound on the maximum change in energy that can
+        result from flipping a single variable in a binary quadratic model.
+
+        The bound is useful as a starting point for determining the values of
+        `penalty parameters`_ in a `penalty model`_.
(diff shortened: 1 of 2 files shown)


arcondello (commented):
Love it, can you also add a [release note](https://github.com/dwavesystems/dimod#release-notes) and some tests?

arcondello (approved):
LGTM!

Issue #1108, opened 2022-02-09
generators.frustrated_loop should output an error message when good_cycles < num_cycles


JoelPasvolsky:
**Description**
When the generator fails to find the requested number of cycles it raises the error without a message. One can see the code line that triggered the error, and it's reasonably clear, but nicer to produce a proper message

**Steps To Reproduce**
```
>>> bqm = dimod.generators.frustrated_loop(4, 3, R=1, seed=1)
RuntimeError:
```
**Expected Behavior**
A clear and concise error message.

arcondello:
Closed by https://github.com/dwavesystems/dimod/pull/1109

PR #1109, opened 2022-02-10
Refresh generators docstrings & deprecate random generators' ``cls`` parameter

JoelPasvolsky:


dimod/generators/anti_crossing.py
```diff
@@ -84,6 +90,18 @@ def anti_crossing_loops(num_variables: int) -> BinaryQuadraticModel:
         quantum annealing of a 16-qubit problem. Nat Commun 4, 1903 (2013).
         https://doi.org/10.1038/ncomms2920
 
+    Examples:
+        The graphic lines of this example require the `Matplotlib <https://matplotlib.org>`_
```
arcondello, line 94: again, I don't think this is a very useful example without the visual. Better to have nothing.

dimod/generators/fcl.py
```diff
@@ -79,6 +80,15 @@ def frustrated_loop(graph: GraphLike,
 
         seed: Random seed.
 
+    Examples:
+        This example creates a BQM from four frustrated loops built on a six-node
+        complete graph, with a maximum allowed interaction of :math:`\pm 2` on
+        any edge.
+
+        >>> bqm = dimod.generators.frustrated_loop(6, 4, R=2, seed=1)
+        >>> set(bqm.quadratic.values())
```
arcondello, line 89: I don't think this is a very informative example. Better to have nothing.

dimod/generators/gates.py
```diff
@@ -68,21 +78,27 @@ def and_gate(in0: Variable, in1: Variable, out: Variable,
...
+        strength: Energy of the lowest-energy infeasible state.
 
     Returns:
         A binary quadratic model with ground states corresponding to a full
         adder gate. The model has five variables and ten interactions.
 
+    Examples:
+        >>> bqm = dimod.generators.fulladder_gate('x1', 'x2', 'x3', 'z', 'c')
+        >>> bqm.fix_variables({'x1': 0, 'x2': 1, 'x3': 1})
+        >>> print(dimod.ExactSolver().sample(bqm).first)
```
arcondello, line 99: Why not just do the same thing as with `and_gate`?

dimod/generators/gates.py
```diff
@@ -115,20 +131,26 @@ def fulladder_gate(in0: Variable, in1: Variable, in2: Variable, sum_: Variable,
...
+        strength: Energy of the lowest-energy infeasible state.
 
     Returns:
         A binary quadratic model with ground states corresponding to a half
         adder gate. The model has four variables and six interactions.
 
+    Examples:
+        >>> bqm = dimod.generators.halfadder_gate('x1', 'x2', 'z', 'c')
+        >>> bqm.fix_variables({'x1': 0, 'x2': 1})
+        >>> print(dimod.ExactSolver().sample(bqm).first)
```
arcondello, line 151: Why not just do the same thing as with `and_gate`?

dimod/generators/gates.py
JoelPasvolsky: To save space... I can limit the output lines instead

dimod/generators/anti_crossing.py
arcondello: I think for these more advanced generators, we don't really get a lot of benefit from these examples. Anyway who knows what they are won't need them and other folks won't get much from a short example. Especially one without the visual. I would recommend adding "frustrated loop" to the SDK glossary and put images there and just link from here.

JoelPasvolsky: Can do. Of course the visual will need to be in a SDK release before a link to it can be added ... :-)

arcondello: That's fine I think

arcondello:
Do you also want to just do https://github.com/dwavesystems/dimod/issues/1108 as part of this PR?

dimod/generators/gates.py
```diff
@@ -68,21 +78,30 @@ def and_gate(in0: Variable, in1: Variable, out: Variable,
...
+        carry: Variable label for the carry output.
+        strength: Energy of the lowest-energy infeasible state.
 
     Returns:
         A binary quadratic model with ground states corresponding to a full
         adder gate. The model has five variables and ten interactions.
 
+    Examples:
+        >>> bqm = dimod.generators.fulladder_gate('a1', 'a2', 'a3', 's', 'c')
+        >>> print(dimod.ExactSolver().sample(bqm).lowest().slice(0, 6, 2))
```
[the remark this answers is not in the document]
JoelPasvolsky: Okay. @randomir usually grudges me every docstring line so I've become frugal.

JoelPasvolsky:
> Do you also want to just do #1108 as part of this PR?

Sure.

dimod/generators/gates.py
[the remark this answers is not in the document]
arcondello: Haha, it's not the number of lines, it's the information content. I am much more begrudging about the next around it, but here I think it conveys the point very nicely to see the truth table.

JoelPasvolsky:
Commit https://github.com/dwavesystems/dimod/pull/1109/commits/f92b921391c5063e1c981829ae638691c99061fc closes https://github.com/dwavesystems/dimod/issues/1108

arcondello (approved):

Issue #1110, opened 2022-02-11
Relabeling creates inconsistent BQMs


JoelPasvolsky:
**Description**
Relabeling can create BQMs that are inconsistent with generated BQMs:

```
>>>  bqm = dimod.BinaryQuadraticModel({'a': 1, 'b': 2}, {('a', 'b'): 1.5}, 'BINARY')
>>> bqm.relabel_variables({'a': 'c', 'b': 'c'})
>>> print(bqm.linear, bqm.quadratic)
{'c': 2.0, 'c': 2.0} {('c', 'c'): 1.5}
```

In comparison:

```
>>> bqm = dimod.BinaryQuadraticModel({'c': 1, 'c': 2}, {('c', 'c'): 1.5}, 'BINARY')
>>> print(bqm.linear, bqm.quadratic)
{'c': 3.5} {}
```

**Expected Behavior**
Consistent treatment of a variable set in multiple places. For the given example, it's noticeable, but in relabeling large numbers of variables in a big BQM, this might not end well

**Environment**
dimod version here is 0.10.10

arcondello:
Also `{'c': 1, 'c': 2}` in your second example is just `{'c': 2}`. You can get something similar to what you're looking for with `bqm.contract_variables(''a, 'b')`.

PR #1111, opened 2022-02-11
Fix relabelling two model variables to the same target

arcondello:
Closes https://github.com/dwavesystems/dimod/issues/1110


JoelPasvolsky (approved):
Thanks!

Issue #668, opened 2020-06-30
BQM.nbytes or similar


arcondello:
Something like [ndarray.nbytes](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.nbytes.html).

PR #1113, opened 2022-02-14
Add QuadraticModel.nbytes() and BinaryQuadraticModel.nbytes() methods

arcondello:
Closes https://github.com/dwavesystems/dimod/issues/668

### DIFF
--- a/tests/test_quadratic_model.py
+++ b/tests/test_quadratic_model.py
@@ -606,6 +606,29 @@ class TestIsEqual(unittest.TestCase):
         self.assertTrue(qm.is_equal(bqm))
 
 
+class TestNBytes(unittest.TestCase):
+    @parameterized.expand([(np.float32,), (np.float64,)])
+    def test_small(self, dtype):
+        qm = dimod.QuadraticModel(dtype=dtype)
+        qm.add_variable('INTEGER', 'a')
+        qm.add_variable('BINARY', 'b')
+        qm.add_variable('SPIN', 'c')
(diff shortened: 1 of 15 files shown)


tests/test_quadratic_model.py
```diff
@@ -606,6 +606,29 @@ def test_bqm(self):
...
+
+        size = sum([itemsize,  # offset
+                    qm.num_variables*itemsize,  # linear
+                    2*qm.num_interactions*(itemsize + np.dtype(np.float32).itemsize),  # quadratic
+                    qm.num_variables*3*itemsize,  # vartype info and bounds
+                    ])
+
+        self.assertEqual(qm.nbytes(), size)
+        self.assertEqual(qm.nbytes(), qm.nbytes(False))
+        self.assertGreaterEqual(qm.nbytes(True), qm.nbytes(False))
```
jackraymond, line 629: Is there a stronger test than this on capacity (similar other functions)? I suppose its difficult to control allocations.

jackraymond (approved):
Couldn't see any errors or shortcomings in commentary.
Tests on capacity is only an inequality - perhaps this could be strengthened?
Testing two models might be valuable, removing a variable and retesting - I suppose this might impact capacity = True/False differently. Perhaps capacity is unchanged.

tests/test_quadratic_model.py
arcondello: Yes, especially since we don't know what compiler a user might be using locally.

Issue #1114, opened 2022-02-14
Summation ranges for QM: i<=j versus i<j


JoelPasvolsky:
**Description**
[class QuadraticModel](https://github.com/dwavesystems/dimod/blob/655e330d5813546630191a7b1e594e63d240a33e/dimod/quadratic/quadratic_model.py#L127) defines the summation range for quadratic coefficients as :math:`\sum_{i<j} b_{i, j} x_i x_j`, and that is accurate for binary variables but quadratic models allow self loops for int variables, so it seems to me the summation for QMs should be :math:`\sum_{i \le j} b_{i, j} x_i x_j`

**Steps To Reproduce**

```
>>> from dimod import QuadraticModel
>>> qm = QuadraticModel({}, {(0, 0): 1}, 0, {0: "INTEGER"})
```

**Expected Behavior**
Update definition as needed.

**Additional Context**
https://docs.ocean.dwavesys.com/en/stable/concepts/cqm.html

arcondello:
Feel free to make the change. Probably just going to `<=` and adding a note that self-loops for binary and spin variables are not allowed.

PR #1115, opened 2022-02-15
Replace i<=j for i<j in quadratic model definition

JoelPasvolsky:
Close https://github.com/dwavesystems/dimod/issues/1114

### DIFF
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -72,18 +72,23 @@ class ConstrainedQuadraticModel:
 
         \begin{align}
             \text{Minimize an objective:} & \\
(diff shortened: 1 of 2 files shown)


arcondello (approved):

arcondello:
Actually, can we use `.. tip::` instead? I think it would look way better 
image

arcondello:
The footnotes are just awful, though I suppose they [look better in other themes](https://pydata-sphinx-theme.readthedocs.io/en/latest/demo/kitchen-sink/paragraph-markup.html#footnotes).

arcondello:
I am fine to leave it then I suppose

PR #1116, opened 2022-02-15
Add additional information to the CQM.to_file() header

arcondello:


dimod/constrained.py
```diff
@@ -1090,6 +1090,29 @@ def num_biases(self) -> int:
                           for const in self.constraints.values())
         return num_biases
 
+    def num_quadratic_real_variables(self) -> int:
```
JoelPasvolsky, line 1093: what about ``num_variables_real_quadratic`` so it's the number of variables that are real and have quadratic biases? That way you have the "number of variables" first, then filtered by "real" then filtered by "quadratic". Might provide better order when there are lots of such functions.
Then the other would be ``num_linear_variables_real`` for the same considerations. 
Ugly, in both cases. 
At least it is not "continuous"

arcondello: The problem is I don't want to confuse it with `num_variables` which is unique variables in the CQM, and this which counts them twice if they appear in two constraints.

JoelPasvolsky: right, but the counted object is variable, so you have to rely on the modifiers to distinguish

arcondello: I think I like `num_linear_biases_real` and `num_quadratic_variables_real`, where real is a narrowing down of existing fields `num_linear_biases` and `num_quadratic_variables`.

dimod/constrained.py
```diff
@@ -1083,21 +1083,88 @@ def lower_bound(self, v: Variable) -> Bias:
...
-            lhs = const.lhs
-            count += sum(lhs.degree(v) > 0 for v in lhs.variables)
-        return count
+    def num_biases(self, vartype: Optional[VartypeLike] = None, *,
+                   linear_only: bool = False,
+                   ) -> int:
+        """The number of biases across the objective and constraints.
+
+        Args:
+            vartype: If provided, only variables of the given type are counted.
```
JoelPasvolsky, line 1092:
```suggestion
            vartype: Count only variables of the specified :class:`~dimod.Vartype`.
```

dimod/constrained.py
```diff
@@ -1083,21 +1083,88 @@ def lower_bound(self, v: Variable) -> Bias:
...
+
+            linear_only: Only count the linear biases.
```
JoelPasvolsky, line 1094:
```suggestion
            linear_only: Count only linear biases.
```

dimod/constrained.py
```diff
@@ -1083,21 +1083,88 @@ def lower_bound(self, v: Variable) -> Bias:
...
+                                          for u, v, _ in qm.iter_quadratic())
+                    return num_biases
+
+        return count(self.objective) + sum(count(const.lhs) for const in self.constraints.values())
+
+    def num_quadratic_variables(self, vartype: Optional[VartypeLike] = None, *,
+                                include_objective: Optional[bool] = None,
+                                ) -> int:
+        """The total number of variables with at least one quadratic interaction
+        across the constraints and objective.
```
JoelPasvolsky, line 1124:
```suggestion
        """Number of variables with at least one quadratic interaction in the constrained quadratic model.
        
        Includes interactions in both the objective and any constraints.
```

dimod/constrained.py
```diff
@@ -1083,21 +1083,88 @@ def lower_bound(self, v: Variable) -> Bias:
...
+
+        Args:
+            vartype: If provided, only variables of the given type are counted.
```
JoelPasvolsky, line 1127:
```suggestion
            vartype: Count only variables of the specified :class:`~dimod.Vartype`.
```

dimod/constrained.py
```diff
@@ -1083,21 +1083,88 @@ def lower_bound(self, v: Variable) -> Bias:
...
+
+            include_objective: Whether to include variables in the objective
+                in the count. Currently defaults to false.
```
JoelPasvolsky, line 1130:
```suggestion
            include_objective: Count also variables in the objective. Currently defaults to false.
```

dimod/constrained.py
```diff
@@ -1083,21 +1083,88 @@ def lower_bound(self, v: Variable) -> Bias:
...
+
+        Return:
+            The number of variables.
+
+        .. deprecated:: 0.10.14
+
+            In dimod 0.12.0``include_objective`` will default to true.
```
JoelPasvolsky, line 1137:
```suggestion
            In dimod 0.12.0 ``include_objective`` will default to true.
```

dimod/constrained.py
```diff
@@ -1083,21 +1083,88 @@ def lower_bound(self, v: Variable) -> Bias:
...
+                return sum(qm.degree(v) > 0 for v in qm.variables)
+        else:
+            vartype = as_vartype(vartype, extended=True)
+
+            def count(qm: Union[QuadraticModel, BinaryQuadraticModel]) -> int:
+                if isinstance(qm, BinaryQuadraticModel):
+                    return sum(qm.degree(v) > 0 for v in qm.variables) if qm.vartype is vartype else 0
+                else:
+                    return sum(qm.vartype(v) is vartype and qm.degree(v) > 0
+                               for v in qm.variables)
```
JoelPasvolsky, line 1160: Can you move the ``for`` up? I first parsed this line as ``return sum(...) > 0``, and was confused why it was returning a boolean

releasenotes/notes/cqm-header-8d40959255466603.yaml
```diff
@@ -0,0 +1,13 @@
...
+    ``ConstrainedQuadraticModel.to_file()`` now uses serialization format 1.3.
+    This format is backwards compatibile.
+  - |
+    Add ``vartype`` and ``linear_only`` arguments to ``ConstrainedQuadraticModel.num_biases()``.
+  - |
+    Add ``vartype`` and ``include_objective`` arguments to ``ConstrainedQuadraticModel.num_quadratic_variables()``.
+deprecations:
+  - |
+    In the future, ``ConstrainedQuadraticModel.num_quadratic_variables()``
+    will also count the objective towards the count by default.
```
JoelPasvolsky, line 13: "also count the objective towards the count" --> "also include the objective towards the count"?

JoelPasvolsky (approved):

arcondello:
I believe I have incorporated all of the feedback if you want to look again

JoelPasvolsky (commented):
LGTM

PR #1120, opened 2022-02-16
Put all files generated by sphinx into the same directory

arcondello:
See https://github.com/dwavesystems/dwave-ocean-sdk/pull/191 for an example of why this is useful


arcondello:
Closing as stale

PR #1124, opened 2022-02-24
Refresh higher-order docstrings & add examples

JoelPasvolsky:
Current doctest failure is unrelated (``README.rst.rst: WARNING: broken link: https://img.shields.io/pypi/pyversions/dimod.svg``) and will probably resolve itself later

### DIFF
--- a/dimod/higherorder/utils.py
+++ b/dimod/higherorder/utils.py
@@ -24,7 +24,6 @@ from collections import defaultdict
 from functools import partial
 
 import numpy as np
-from numpy.typing import ArrayLike
 
 import dimod
 from dimod.binary_quadratic_model import BinaryQuadraticModel
@@ -104,7 +103,7 @@ def reduce_binary_polynomial(poly: BinaryPolynomial) -> Tuple[
         List[Tuple[FrozenSet[Hashable], Number]],
         List[Tuple[FrozenSet[Hashable], Hashable]]
     ]:
-    """Reduce a binary polynomial to a linear and quadratic terms, plus constraints.
+    """Reduce a binary polynomial to linear and quadratic terms, plus constraints.
 
     Introduces auxillary variables and constraints to reduce the polynomial
     to linear and quadratic terms.
@@ -364,7 +363,7 @@ def poly_energy(sample_like, poly: Mapping[Tuple[Variable, Variable], Bias]) -> 
 
 
 def poly_energies(samples_like, poly: Mapping[Tuple[Variable, Variable], Bias]
-                  ) -> ArrayLike:
+                  ) -> np.ndarray:
     """Calculates energy of samples from a higher order polynomial.
 
     Args:
@@ -373,11 +372,19 @@ def poly_energies(samples_like, poly: Mapping[Tuple[Variable, Variable], Bias]
             NumPy's array_like structure. See :func:`.as_samples`.
 
         poly:
-            Polynomial as a dict of form `{term: bias, ...}``, where `term` is a
+            Polynomial as a dict of form `{term: bias, ...}`, where `term` is a
             tuple of variables and `bias` the associated bias. Variable
             labeling/indexing here must match that of ``samples_like``.
 
     Returns: Energies of the samples.
+
+    Examples:
+        >>> poly = dimod.BinaryPolynomial({'a': -1, ('a', 'b'): 1, ('a', 'b', 'c'): -1},
+        ...                               dimod.BINARY)
+        >>> samples = [{'a': 1, 'b': 1, 'c': 0},
+        ...            {'a': 1, 'b': 1, 'c': 1}]
+        >>> dimod.poly_energies(samples, poly)
+        array([ 0., -1.])
 
     """
     return BinaryPolynomial(poly, 'SPIN').energies(samples_like)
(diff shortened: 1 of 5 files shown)


dimod/higherorder/utils.py
```diff
@@ -332,41 +336,55 @@ def make_quadratic(poly, strength, vartype=None, bqm=None):
...
+        >>> sample = {'a': 1, 'b': 1, 'c': 0}
+        >>> dimod.poly_energy(sample, poly)
+        0.0
 
     """
     return BinaryPolynomial(poly, 'SPIN').energy(sample_like)
 
 
-def poly_energies(samples_like, poly):
+def poly_energies(samples_like, poly: Mapping[Tuple[Variable, Variable], Bias]
```
arcondello, line 365: we now have `SamplesLike` in `dimod.typing`.

dimod/higherorder/utils.py
```diff
@@ -336,11 +336,12 @@ def make_quadratic(poly: Mapping[Tuple[Variable, Variable], Bias], strength: flo
     return bqm
 
 
-def poly_energy(sample_like, poly: Mapping[Tuple[Variable, Variable], Bias]) -> float:
+def poly_energy(sample_like: SamplesLike,
+                poly: Mapping[Tuple[Variable, Variable], Bias]) -> float:
```
arcondello, line 340: This should be `Mapping[Tuple[Variable, ...]], Bias]` since the tuples are variable length. Also, you can just do
```python
Poly = Mapping[Tuple[Variable, ...]], Bias]
```
somewhere at the top if you like and then use it everywhere.

arcondello: Also, technically I think it's a sequence, so 
```python
Poly = Mapping[Sequence[Variable]], Bias]
```

arcondello: Could also consider adding this new type to `dimod.typing` and using it from there.

dimod/higherorder/utils.py
```diff
@@ -269,7 +269,7 @@ def var(x):
     return cqm
 
 
-def make_quadratic(poly: Mapping[Tuple[Variable, Variable], Bias], strength: float,
+def make_quadratic(poly: Polynomial, strength: float,
```
arcondello, line 272: Isn't this also `Union[Polynomial, BinaryPolynomial]`? Applies other places as well

JoelPasvolsky: Is not currently stated as such: https://docs.ocean.dwavesys.com/en/stable/docs_dimod/reference/generated/dimod.higherorder.utils.make_quadratic.html#dimod.higherorder.utils.make_quadratic 
Might just have been missed. I can check that it works and add, assuming we don't want to only allow the one in some cases

arcondello: Anywhere we do `poly = _init_binary_polynomial(poly, vartype)` it seems to work

JoelPasvolsky: Is this not a worry though?

```
>>> poly = dimod.BinaryPolynomial({"a": -2, ("a", "b"): -0.5, ("b", "c"): 1.5, ("a", "b", "c"): 3}, dimod.SPIN)
>>> dimod.poly_energy({"a": 1, "b": 1, "c": -1}, poly)
-7.0
>>> dimod.poly_energy({"a": 1, "b": 0, "c": 0}, poly)
-2.0
```

For just `poly: Polinomial` it should be the user's responsibility to ensure that sample vartype is correct for the poly but if we allow `BinaryPolynomial` shouldn't the energy calculation balk at a mismatched vartype?

arcondello: Ah, interesting. Probably `_init_binary_polynomial` should handle that case.

JoelPasvolsky: Changing `return BinaryPolynomial(poly, 'SPIN').energies(samples_like)` to use `_init_binary_polynomial` won't help with the static `poly_energy` & `poly_energies` methods. And I think maybe this should be checked in `dimod.sampleset.as_samples`: 
```
>>> poly_energy({"a": 1, "b": -1, "c": 0}, poly)
-1.5
```
I think `as_samples` should barf at a "sample" that has both 0 and -1 in it.

arcondello: We are actually very permissive for samples, deliberately. We've considered adding a `strict=True` (https://github.com/dwavesystems/dimod/issues/954) keyword on occasion, but IMO it's better to assume the user has a reason.

JoelPasvolsky: Okay, I think I added `BinaryPolynomial` to all relevant functions

dimod/typing.py
```diff
@@ -39,15 +39,16 @@
...
            'SampleLike',
            'SamplesLike',
            'Variable',
            'VartypeLike',
            ]
 
-
-Variable = Hashable  # todo: exclude None
 Bias = Union[int, float, np.number]
+Variable = Hashable  # todo: exclude None
```
arcondello, line 50: Why move this? Unless it's to make it alphabetical in which case why put Polynomial after?

Also, adding `Polynomal` needs a release note.

JoelPasvolsky: It was to make it alphabetical but `Variable` is required for `Polynomial`, so best-effort alphabetical. 
Will add RN
(I just rebased too)

arcondello (approved):

Issue #1136, opened 2022-03-02
`QuadraticModel.energy` does not include squared terms


mcfarljm:
Identified by @vgoliber 

``` python
>>> import dimod
>>> i = dimod.Integer('i')
>>> qm = i**2
>>> print(qm.energy({'i': 1}))
0.0
```

Dimod version 0.10.14

PR #1137, opened 2022-03-02
Fix energy calculations for QMs with squared terms

arcondello:
Closes https://github.com/dwavesystems/dimod/issues/1136

### DIFF
--- a/tests/test_quadratic_model.py
+++ b/tests/test_quadratic_model.py
@@ -296,6 +296,11 @@ class TestEnergies(unittest.TestCase):
         self.assertAlmostEqual((i*j).energy({'i': 4294967296, 'j': 4294967296}),
                                1.8446744073709552e+19)
 
+    def test_bug1136(self):
+        # https://github.com/dwavesystems/dimod/issues/1136
+        i = dimod.Integer('i')
+        self.assertEqual((i**2).energy({'i': 1}), 1)
+
     def test_empty(self):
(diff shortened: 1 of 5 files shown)


alexzucca90:
do we need to modify this part too?
https://github.com/dwavesystems/dimod/blob/ffee615fe6b48a829d44e455fd9fe9561a1496ec/dimod/include/dimod/quadratic_model.h#L410

arcondello:
Good catch, we should. Though that's not used from the Python level

tests/test_quadratic_model.py
```diff
@@ -296,6 +296,11 @@ def test_bug982(self):
         self.assertAlmostEqual((i*j).energy({'i': 4294967296, 'j': 4294967296}),
                                1.8446744073709552e+19)
 
+    def test_bug1136(self):
+        # https://github.com/dwavesystems/dimod/issues/1136
+        i = dimod.Integer('i')
+        self.assertEqual((i**2).energy({'i': 1}), 1)
```
alexzucca90, line 302: I think this is better to have with an integer value `!=` 1. If the bug was not fixed, this test would not catch it.

alexzucca90: Actually, I'm wrong... there would be linear term = 0 and the energy would be 0, sorry

alexzucca90 (approved):

PR #1140, opened 2022-03-04
Add SampleSet.from_samples_cqm() method

seatim:


seatim:
One possible addition that seems sensible, would be to use `.from_samples_cqm()` in `ExactCQMSolver.sample_cqm()`.

seatim:
Note: 810b87d introduces a change in behavior that I believe fixes an minor extant bug.

The root cause of the (AFAIK unreported) bug is the missing `abs` call around `info.rhs_energy` in the [list comprehension that is assigned to `is_satisfied`](https://github.com/dwavesystems/dimod/blob/main/dimod/reference/samplers/exact_solver.py#L246).

C.f. what I believe is the correct expression [here in `ConstrainedQuadraticModel.check_feasible`](https://github.com/dwavesystems/dimod/blob/main/dimod/constrained.py#L677).  This is the expression that `SampleSet.from_samples_cqm()` uses.

Should we add a release note to document this change in behavior?  Should it be labeled as a bug fix?

dimod/sampleset.py
```diff
@@ -789,6 +789,75 @@ def from_samples_bqm(cls, samples_like, bqm, **kwargs):
...
+        samples_like = samples, labels = as_samples(samples_like)
+
+        energies = cqm.objective.energies(samples_like)
+
+        is_satisfied = [[datum.violation <= atol + rtol*abs(datum.rhs_energy)
+                         for datum in cqm.iter_constraint_data((s, labels))] for s in samples]
+        is_feasible = [all(satisfied) for satisfied in is_satisfied]
+
+        info = kwargs.get('info', {})
+        info['constraint_labels'] = list(cqm.constraints.keys())
```
arcondello, line 855: this makes me a bit nervous. I know that dicts are ordered in Python 3.7+, and that [.iter_constraint_data() uses that ordering](https://github.com/dwavesystems/dimod/blob/003253f5a28419ed7664042bb7dd3644c78832c0/dimod/constrained.py#L975) so it's safe, but this feels like a place that a bug could easily creep in.

Can we either read the label from `datum`, or document in `iter_constraint_data` that it's guaranteed to use the same order as `.constraints`, or both? Perhaps a test as well.

dimod/sampleset.py
```diff
@@ -789,6 +789,75 @@ def from_samples_bqm(cls, samples_like, bqm, **kwargs):
...
+        # and in cls.from_samples
+        samples_like = samples, labels = as_samples(samples_like)
+
+        energies = cqm.objective.energies(samples_like)
+
+        is_satisfied = [[datum.violation <= atol + rtol*abs(datum.rhs_energy)
+                         for datum in cqm.iter_constraint_data((s, labels))] for s in samples]
+        is_feasible = [all(satisfied) for satisfied in is_satisfied]
+
+        info = kwargs.get('info', {})
```
arcondello, line 854: You could use [`.setdefault()`](https://docs.python.org/3/library/stdtypes.html#dict.setdefault) to save on the `        kwargs['info'] = info` below. Totally aesthetic but :shrug:

arcondello:
> Should we add a release note to document this change in behavior? Should it be labeled as a bug fix?

IMO yes, a bug fix would be appropriate.

dimod/sampleset.py
seatim: Thanks for that feedback, I was thinking about this also.  Here is a question relating to possible implementations of this.  What is the expected behavior for the case of `samples_like` being empty?

`SampleSet.from_samples` supports empty `samples_like`.  But `SampleSet.from_samples_bqm` does not:

```python
>>> bqm = dimod.BinaryQuadraticModel.from_ising({}, {('a', 'b'): -1})
>>> sampleset = dimod.SampleSet.from_samples_bqm([], bqm)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/timdoyle/src/dimod/dimod/sampleset.py", line 788, in from_samples_bqm
    energies = bqm.energies(samples_like)
  File "/Users/timdoyle/src/dimod/dimod/binary/binary_quadratic_model.py", line 1079, in energies
    return self.data.energies(samples_like, dtype=dtype)
  File "dimod/binary/cybqm/cybqm_template.pyx.pxi", line 378, in dimod.binary.cybqm.cybqm_float64.cyBQM_template.energies
    return np.asarray(self._energies[np.float64_t](samples, labels), dtype=dtype)
  File "dimod/binary/cybqm/cybqm_template.pyx.pxi", line 341, in dimod.binary.cybqm.cybqm_float64.cyBQM_template._energies
    raise ValueError(f"missing variable {self.variables[si]!r} in sample(s)")
ValueError: missing variable 'a' in sample(s)
```

So, the question becomes, should `SampleSet.from_samples_cqm` behave like `SampleSet.from_samples_bqm` (effectively not support empty `samples_like`) or should it behave like `SampleSet.from_samples`?

I am leaning toward the latter because the `ValueError` raised by `SampleSet.from_samples_bqm` seems like an accident rather than intention.  That interpretation raises another question: should we change the behavior of `SampleSet.from_samples_bqm` to support empty `samples_like` also?  I would lean toward "yes" to that also.  In any case, having consistent behavior in handling empty `samples_like` among all three methods seems desirable.

arcondello: Yes, IMO the latter makes sense. I think that's another bug :facepalm:

seatim: > Can we either read the label from `datum`, or document in `iter_constraint_data` that it's guaranteed to use the same order as `.constraints`, or both? Perhaps a test as well.

✔️ 
Went with the "document that ``iter_constraint_data`` is gauranteed to use the same order" approach and added a unit test to cover that.

dimod/sampleset.py
seatim: ✔️ 
Nice, thanks

arcondello (approved):

PR #1143, opened 2022-03-07
Refresh constrained

JoelPasvolsky:


dimod/constrained.py
```diff
@@ -95,45 +95,45 @@ class ConstrainedQuadraticModel:
 
     Example:
 
-        Solve a simple `bin packing problem <https://w.wiki/3jz4>`_. In this
-        problem we wish to pack a set of items of different weights into
-        the smallest number of bins possible.
+        This example solves the simple `bin packing problem <https://w.wiki/3jz4>`_
+        problem of packing a set of items of different weights into the smallest
```
arcondello, line 99: Won't this read `This examples solves the simple bin packing problem problem of packing a set...`

dimod/constrained.py
```diff
@@ -457,37 +488,43 @@ def add_discrete_from_comparison(self,
                                      comp: Comparison,
                                      label: Optional[Hashable] = None,
                                      copy: bool = True) -> Hashable:
-        """Add a one-hot constraint.
+        """Add a one-hot constraint from a comparison.
+
+        One-hot constraints can represent discrete variables (for example a
```
arcondello, line 493: I think maybe all of these functions should have have a ``See :meth:`add_constraint()` `` so we're not repeating ourselves too much.

dimod/constrained.py
```diff
@@ -905,7 +1042,22 @@ def from_qm(cls, qm: QuadraticModel) -> 'ConstrainedQuadraticModel':
     def from_file(cls, fp: Union[BinaryIO, ByteString]) -> "ConstrainedQuadraticModel":
         """Construct from a file-like object.
 
+        Args:
+            fp: File pointer. If file-like, should be a readable, seekable
+                file-like object. If bytes-like, it will be wrapped with
+                :class:`io.BytesIO`.
```
arcondello, line 1048: this seems like an internal detail.

dimod/constrained.py
```diff
@@ -997,18 +1165,20 @@ def iter_constraint_data(self, sample_like) -> Iterator[ConstraintData]:
                 label=label,
                 )
 
-    def iter_violations(self, sample_like, *, skip_satisfied: bool = False, clip: bool = False,
-                        ) -> Iterator[Tuple[Hashable, Bias]]:
+    def iter_violations(self, sample_like: SamplesLike, *,
+            skip_satisfied: bool = False,
+            clip: bool = False,) -> Iterator[Tuple[Hashable, Bias]]:
```
arcondello, line 1170: newline after the comma, also indentation is off.

docs/reference/constrained.rst
```diff
@@ -53,6 +53,7 @@ Methods
    ConstrainedQuadraticModel.iter_violations
```
arcondello, line 53: I think these should all be `~ConstrainedQuadraticModel.` to be consistent with `QuadraticModel` and `BinaryQuadraticModel`.

dimod/constrained.py
JoelPasvolsky: I considered doing that but the params slightly differ and I suspect there's a good chance that a user getting it work won't pop over to see the cautions on (1) supported vars and (2) LeapHybridCQMSampler behavior. 
If you're unhappy with repeating those two sentences though, I can have them just in ``add_discrete``.

arcondello (approved):

PR #1144, opened 2022-03-08
add inner_child_properties tool

pau557:
This tool will descend and find the properties of the inner-most children sampler in a composite (or just a sampler).

Useful for extracting QPU parameters when the QPU sampler is wrapped on many composite layers.

### DIFF
--- a/dimod/utilities.py
+++ b/dimod/utilities.py
@@ -12,6 +12,8 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+from __future__ import annotations
+
 import copy
 import itertools
 import os
@@ -22,14 +24,16 @@ from functools import reduce
 
 import numpy as np
 
+import dimod
 from dimod.typing import Variable
+
 
 __all__ = ['ising_energy',
            'qubo_energy',
            'ising_to_qubo',
            'qubo_to_ising',
            'child_structure_dfs',
-           'inner_child_properties',
+           'innermost_child_properties',
            'get_include',
            ]
 
@@ -450,9 +454,8 @@ def child_structure_dfs(sampler, seen=None):
 
     raise ValueError("no structured sampler found")
 
-def inner_child_properties(sampler):
-    """
-    Returns the properties of inner-most child sampler in a composite.
+def innermost_child_properties(sampler: dimod.Sampler) -> typing.Dict[str, typing.Any]:
+    """Returns the properties of the inner-most child sampler in a composite.
 
     Args:
         sampler: A dimod sampler
@@ -462,16 +465,15 @@ def inner_child_properties(sampler):
 
     Example:
     >>> sampler = EmbeddingComposite(ScaleComposite(qpu_sampler))
-    >>> inner_child_properties(sampler)["annealing_time_range]
+    >>> inner_child_properties(sampler)["annealing_time_range"]
     [0.5, 2000.0]
     """
 
-    def get_properties(properties):
-        if "child_properties" in properties:
-            return get_properties(properties["child_properties"])
-        else:
-            return properties
-    return get_properties(sampler.properties)
+    try:
+        return innermost_child_properties(sampler.child)
+    except AttributeError:
+        return sampler.properties
+
 
 def get_include():
     """Return the directory with dimod's header files."""
--- a/dimod/utilities.py
+++ b/dimod/utilities.py
@@ -29,6 +29,7 @@ __all__ = ['ising_energy',
            'ising_to_qubo',
            'qubo_to_ising',
            'child_structure_dfs',
+           'inner_child_properties',
            'get_include',
            ]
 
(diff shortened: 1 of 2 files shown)


dimod/utilities.py
```diff
@@ -449,6 +450,28 @@ def child_structure_dfs(sampler, seen=None):
...
+
+    Returns:
+        properties (dict): The properties of the inner-most sampler
+
+    Example:
+    >>> sampler = EmbeddingComposite(ScaleComposite(qpu_sampler))
+    >>> inner_child_properties(sampler)["annealing_time_range]
+    [0.5, 2000.0]
+    """
+
```
arcondello, line 468: maybe
```python
try:
    return inner_child_properties(sampler.child)
except AttributeError:
    return sampler.properties
```

dimod/utilities.py
```diff
@@ -449,6 +450,28 @@ def child_structure_dfs(sampler, seen=None):
 
     raise ValueError("no structured sampler found")
 
+def inner_child_properties(sampler):
+    """
+    Returns the properties of inner-most child sampler in a composite.
```
arcondello, line 455: This should be on the first line

dimod/utilities.py
```diff
@@ -449,6 +450,28 @@ def child_structure_dfs(sampler, seen=None):
 
     raise ValueError("no structured sampler found")
 
+def inner_child_properties(sampler):
```
arcondello, line 453:
```python
def inner_child_properties(sampler: dimod.Sampler) -> typing.Dict[str, typing.Any]:
```

arcondello (commented):
Also, maybe `innermost_child_properties` would be more explicit?

Or, another pattern,
```python
sampler = ScaleComposite(ExactSolver())

sampler.innermost_child().properties

```

dimod/utilities.py
```diff
@@ -449,6 +450,28 @@ def child_structure_dfs(sampler, seen=None):
...
+    Returns the properties of inner-most child sampler in a composite.
+
+    Args:
+        sampler: A dimod sampler
+
+    Returns:
+        properties (dict): The properties of the inner-most sampler
+
+    Example:
+    >>> sampler = EmbeddingComposite(ScaleComposite(qpu_sampler))
```
JoelPasvolsky, line 464: Having `EmbeddingComposite` means a dependency on dwave-system, so you'd need to skip the doctest

dimod/utilities.py
```diff
@@ -449,6 +450,28 @@ def child_structure_dfs(sampler, seen=None):
...
+    >>> inner_child_properties(sampler)["annealing_time_range]
```
JoelPasvolsky, line 465: Should be inner_child_properties(sampler)["annealing_time_range"]

dimod/utilities.py
```diff
@@ -449,6 +450,28 @@ def child_structure_dfs(sampler, seen=None):
...
+    [0.5, 2000.0]
+    """
+
+    def get_properties(properties):
+        if "child_properties" in properties:
+            return get_properties(properties["child_properties"])
+        else:
+            return properties
+    return get_properties(sampler.properties)
```
JoelPasvolsky, line 474: Some of our composites are tricky:
```
>>> sampler = EmbeddingComposite(TilingComposite(DWaveSampler(), 1, 1, 4))
>>> inner_child_properties(sampler["annealing_time_range"])
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-e57179dc0149> in <module>
----> 1 inner_child_properties(sampler["annealing_time_range"])

TypeError: 'EmbeddingComposite' object is not subscriptable
```

dimod/utilities.py
```diff
@@ -449,6 +450,28 @@ def child_structure_dfs(sampler, seen=None):
...
+    >>> sampler = EmbeddingComposite(ScaleComposite(qpu_sampler))
+    >>> inner_child_properties(sampler)["annealing_time_range]
+    [0.5, 2000.0]
+    """
+
+    def get_properties(properties):
+        if "child_properties" in properties:
+            return get_properties(properties["child_properties"])
+        else:
+            return properties
```
JoelPasvolsky, line 473: I'd add a space line here

dimod/utilities.py
JoelPasvolsky, line 455: "of the inner-most child"

dimod/utilities.py
pau557: Try:

```inner_child_properties(sampler)["annealing_time_range"]```

Works for me.

JoelPasvolsky: Yup, user error, sorry

dimod/utilities.py
pau557: is that a problem if it's in the example docstring? It just seems like the most obvious usage to me.

I wasn't aware that docstrings get tested, does that mean that they should include imports and be well self-contained?

dimod/utilities.py
pau557: ugh

pau557:
Your suggestion makes sense @arcondello, worth just closing this PR and making a new one instead of making a complete overhaul

PR #1145, opened 2022-03-09
expose innermost_child sampler

pau557:
Allow composite samplers to expose the innermost child sampler, this can be used to gather properties. For example, one can get QPU properties from a composite made of several layers

### DIFF
--- a/dimod/core/composite.py
+++ b/dimod/core/composite.py
@@ -45,6 +45,8 @@ Examples:
     For more examples, see the source code for the composed
     documented in :ref:`quadratic_composites`.
 """
+from __future__ import annotations
+
 import abc
 
 from dimod.core.sampler import Sampler
@@ -75,10 +77,14 @@ class Composite(abc.ABC):
             raise RuntimeError("A Composite must have at least one child Sampler")
 
 
+
 class ComposedSampler(Sampler, Composite):
     """Abstract base class for dimod composed samplers.
 
     Inherits from :class:`.Sampler` and :class:`.Composite`.
 
     """
-    pass
+
+    def innermost_child(self) -> Sampler:
+        """Returns the inner-most child sampler"""
+        return self.child.innermost_child()    
--- a/dimod/core/sampler.py
+++ b/dimod/core/sampler.py
@@ -105,6 +105,7 @@ non-empty dicts.
             return self._parameters
 
 """
+from __future__ import annotations
 
 import abc
 import typing
@@ -317,3 +318,6 @@ class Sampler(metaclass=SamplerABCMeta):
             kwargs.pop(kw)
 
         return kwargs
+    
+    def innermost_child(self) -> Sampler:
+            return self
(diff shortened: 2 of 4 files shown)


dimod/core/composite.py
```diff
@@ -45,6 +45,8 @@
     For more examples, see the source code for the composed
     documented in :ref:`quadratic_composites`.
 """
+from __future__ import annotations
```
arcondello, line 48: don't need this in this case

dimod/core/composite.py
```diff
@@ -75,10 +77,14 @@ def child(self):
             raise RuntimeError("A Composite must have at least one child Sampler")
 
 
+
 class ComposedSampler(Sampler, Composite):
```
arcondello, line 81: I think if you change this to `class ComposedSampler(Composite, Sampler):` then you don't need to implement the method on this class.

dimod/core/composite.py
```diff
@@ -74,8 +75,12 @@ def child(self):
         except IndexError:
             raise RuntimeError("A Composite must have at least one child Sampler")
 
+    def innermost_child(self) -> Sampler:
+        """Returns the inner-most child sampler"""
```
JoelPasvolsky, line 79:
```suggestion
        """Returns the innermost child sampler."""
```

JoelPasvolsky (commented):
Shouldn't this new feature be release noted?

dimod/core/composite.py
arcondello: Also shouldn't it be `Return`?

dimod/core/sampler.py
```diff
@@ -317,3 +318,6 @@ def remove_unknown_kwargs(self, **kwargs) -> typing.Dict[str, typing.Any]:
             kwargs.pop(kw)
 
         return kwargs
+    
+    def innermost_child(self) -> Sampler:
```
arcondello, line 322: needs a docstring.

dimod/core/composite.py
JoelPasvolsky: Yes, updated, thanks

tests/test_composite.py
```diff
@@ -39,3 +39,12 @@ def children(self):
         sampler = Dummy()
         with self.assertRaises(RuntimeError):
             sampler.child
+
+
+class TestInnermostChild(unittest.TestCase):
+
+    def test_composed_sampler(self):
+        innermost_child = dimod.ExactSolver()
+        sampler = dimod.ClipComposite(dimod.ScaleComposite(innermost_child))
```
arcondello, line 48: I would use `TrackingComposite` and `StructureComposite` since those are staying in dimod through #1127 so the tests stay valid.

arcondello (commented):
Need to add a [release note](https://github.com/dwavesystems/dimod#release-notes).

Probably should be added to the docs as well, see https://github.com/dwavesystems/dimod/blob/main/docs/reference/sampler_composites/api.rst

Issue #969, opened 2021-08-23
`CQM.add_variable` and `QM.add_variable` are inconsistent


arcondello:
`ConstrainedQuadraticModel.add_variable(v, vartype)` vs `QuadraticModel.add_variable(vartype, [v])`. IMO we should switch the CQM version to match the QM with a deprecation period. Something like
```
def add_variable(self, vartype, v):
    if vartype not in [dimod.SPIN, 'SPIN', frozenset([-1, 1]), dimod.BINARY, ...] and v in [[dimod.SPIN, ...]:
        warnings.warn(...)
        vartype, v = v, vartype
    ...
```

**Additional Context**
The other obvious comparable is `BinaryQuadraticModel.add_variable([v])`. In that case the vartype is implicit to the model.

PR #1147, opened 2022-03-11
Swap input argument order in CQM.add_variable()

arcondello:
Closes https://github.com/dwavesystems/dimod/issues/969


dimod/constrained.py
```diff
@@ -681,10 +685,21 @@ def add_variable(self, v: Variable, vartype: VartypeLike,
...
             >>> from dimod import ConstrainedQuadraticModel, Integer
             >>> cqm = ConstrainedQuadraticModel()
-            >>> cqm.add_variable('i', 'INTEGER')
+            >>> cqm.add_variable(INTEGER', 'i')
             'i'
 
         """
+        try:
+            vartype = as_vartype(vartype, extended=True)
+        except TypeError:
```
JoelPasvolsky, line 694: This seems to assume that any `TypeError` is caused by a reversed order. But maybe the order is correct but the type is bad, for example `dimod.Vartype.CONTINUOUS`

dimod/constrained.py
```diff
@@ -681,10 +685,21 @@ def add_variable(self, v: Variable, vartype: VartypeLike,
...
+            # in dimod<0.11 the argument order was v, vartype so let's allow that case
+            warnings.warn(
+                "Providing vartype and v in CQM.add_variable(v, vartype) order "
```
JoelPasvolsky, line 697:
```suggestion
                "Parameter order CQM.add_variable(v, vartype) "
```

dimod/generators/binpacking.py
```diff
@@ -53,15 +54,15 @@ def random_bin_packing(num_items: int,
...
     y = {j: obj.add_variable(f'y_{j}') for j in range(max_num_bins)}
 
     for j in range(max_num_bins):
         obj.set_linear(y[j], 1)
 
     model.set_objective(obj)
 
-    x = {(i, j): model.add_variable(f'x_{i}_{j}', vartype='BINARY') for i in range(num_items) for
+    x = {(i, j): model.add_variable(BINARY, f'x_{i}_{j}') for i in range(num_items) for
          j in range(max_num_bins)}
```
JoelPasvolsky, line 67: I think you have more uses to replace: for example, https://github.com/dwavesystems/dimod/blob/93796d0ceb21d4d98ef03c10a0b42ff1f88e32a8/dimod/constrained.py#L407

dimod/constrained.py
arcondello: In that case the user would get an attribute error.

But even in the `cqm.add_variable('CONTINUUOS', 'i')` case, the user will still get an error message. It will likely be slightly confusing but from the combo of warning and exception they should be able to figure out what happened.

dimod/generators/binpacking.py
arcondello: This is why I didn't want to ignore results on `add_variable()`! :laughing:

dimod/constrained.py
arcondello: Maybe `Parameter order of CQM.add_variable(v, vartype)`? It feels like it's missing a word to me otherwise but I could be wrong

JoelPasvolsky: We're not deprecating the existence of a parameter order in the function, just the particular order. We can make it more explicit, "Parameter order `v, vartype` in CQM.add_variable(v, vartype)"  if you like. 
I just found it not so easy to parse the current first line: the order comes late after the providing

arcondello: Ok, updated

PR #1148, opened 2022-03-11
Don't use a fixed seed in `random_knapsack()` and `random_bin_packing()`

arcondello:


alexzucca90 (commented):
LGTM, quick question: what's the reason for not having fixed seed? Common practice?

arcondello:
Indeed. Especially if I am using a function called `random_knapsack()` I would be surprised if I got the same problem each time.

alexzucca90:
fair enough! approved!

alexzucca90 (approved):

Issue #1149, opened 2022-03-11
Error for overlapping discrete constraints


mhramani:
When I add two discrete constraints over three binary variables using `quicksum` I don't get any error. 
```
import dimod
b1 = dimod.Binary('b1')
b2 = dimod.Binary('b2')
b3 = dimod.Binary('b3')

cqm = dimod.CQM()
cqm.add_discrete(dimod.quicksum([b1, b2]))
cqm.add_discrete(dimod.quicksum([b1, b3]))
```
However, when I add the discrete constraints using list of the binary variables I do get an error message as expected. 
```
cqm.add_discrete(['b1', 'b2'])
cqm.add_discrete(['b1', 'b3'])
``

Issue #966, opened 2021-08-20
`BQM.name` or similar


arcondello:
It would be nice to get the label of one-variable BQMs and QMs. E.g.
```python
>>> i = dimod.Integer('i')
>>> i.name
'i'
```
Has a similar problem to the ones mentioned in #921. Though I think https://github.com/dwavesystems/dimod/issues/921#issuecomment-902922740 provides a good path forward.

Credit to @rahmani15 for the suggestion

PR #1152, opened 2022-03-14
Fix exceptions in `CQM.add_discrete_from_model()`

arcondello:
Previously it would not raise an error when given a binary variable already used in another discrete constraint.

Closes https://github.com/dwavesystems/dimod/issues/1149

### DIFF
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -652,7 +652,13 @@ class ConstrainedQuadraticModel:
             raise ValueError("discrete constraints must have at least two variables")
 
         for v, bias in qm.iter_linear():
-            if not vartype(v) is Vartype.BINARY:
+            if v in self.variables:
+                # it already exists, let's make sure it's not already used
+                if any(v in self.constraints[label].lhs.variables for label in self.discrete):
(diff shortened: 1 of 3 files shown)


dimod/constrained.py
```diff
@@ -652,7 +652,13 @@ def add_discrete_from_model(self,
             raise ValueError("discrete constraints must have at least two variables")
 
         for v, bias in qm.iter_linear():
-            if not vartype(v) is Vartype.BINARY:
+            if v in self.variables:
+                # it already exists, let's make sure it's not already used
+                if any(v in self.constraints[label].lhs.variables for label in self.discrete):
+                    raise ValueError(f"variable {v!r} is already used in a discrete variable")
```
mhramani, line 658: Not related to this PR, but why not something like this:
(f"variable {v.name} is already used in a discrete variable"
I know that v.name doesn't exist and need to do this in another way.

arcondello: Yeah, that's still an open feature request, see https://github.com/dwavesystems/dimod/issues/966

PR #1159, opened 2022-03-29
Add documentation and testing for `dimod.typing`

arcondello:
There is more testing needed, but that will come in followup PRs. For now just putting the minimum amount in that's needed to check that the approach works

### DIFF
--- a/dimod/typing.py
+++ b/dimod/typing.py
@@ -17,8 +17,6 @@ Type hints for common dimod inputs.
 """
 import typing
 
-from typing import Collection, Hashable, Mapping, NamedTuple, Sequence, Tuple, Union
-
 import numpy as np
 
 from dimod.vartypes import VartypeLike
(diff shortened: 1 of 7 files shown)


docs/reference/index.rst
```diff
@@ -14,6 +14,7 @@ Reference Documentation
    generators
    sampler_composites/index
    sampleset
+   typing
```
JoelPasvolsky, line 17: These are not alphabetical, so why not put this after Exceptions? I think users will likely first need utilities and the pages related to variables and their types before they get to this. 
I can live with it here though too.

dimod/typing.py
```diff
@@ -48,67 +46,127 @@
...
+* NumPy's :class:`~numpy.floating` and :class:`~numpy.integer`.
+
+"""
+
+# it would be nice to be able to exclude None from Variable. We can define
+# a special abstract base class that excludes None, but making it behave like
+# a typing class is a pain accross our supported Python versions. In the future
+# we should handle it
+
+Variable = typing.Hashable
```
JoelPasvolsky, line 64: I'm wondering whether the uglier `VariableLabel` would help reduce confusion between QMs representing variables and labels of variables in the models...

dimod/vartypes.py
```diff
@@ -121,7 +121,18 @@ class Vartype(enum.Enum):
 REAL = Vartype.REAL
 
 
+# when we drop 3.7 we can just use typing.Literal
 VartypeLike = typing.Union[Vartype, str, frozenset]
+"""Objects that can be interpreted as a variable type.
```
JoelPasvolsky, line 126: This docstring isn't displayed in the typing.rst output HTML

JoelPasvolsky (approved):
Anything that reduces the typing character count of ``Initialized.parse_initial_states`` from 3000 to 700 is a blessing

dimod/typing.py
```diff
@@ -48,67 +46,127 @@
...
     typing.Mapping[Variable, Bias],
     ArrayLike,  # this is overgenerous, but we cannot easily specify it better
     ]
+"""Objects that can be interpreted as a single sample.
+
+This includes:
+
+* A one-dimensional NumPy array_like_.
+* A 2-:class:`tuple` containing a one-dimensional NumPy array_like_ and a
+  list of variable labels.
```
randomir, line 141: Is this included in the definition above?

tests/test_typing.py
```diff
@@ -0,0 +1,70 @@
...
+    TEST_DIR = os.path.join(os.path.dirname(__file__), 'typing')
+    CACHE_DIR = os.path.join(TEST_DIR, '.mypy_cache')
+
+    def test_mypy(self):
+        stdout, stderr, exit_code = mypy.api.run([
+            '--cache-dir', self.CACHE_DIR,
+            '--follow-imports', 'silent',
+            '--show-absolute-path',
+            self.TEST_DIR])
+
```
randomir, line 41: Perhaps replace this python unittest with a CI job that just calls `mypy`?

randomir (approved):
LGTM

dimod/typing.py
arcondello: Yes. I had suggested that a while ago as you'll recall. I am open to it, though probably as a followup PR

arcondello: Made an issue https://github.com/dwavesystems/dimod/issues/1160

tests/test_typing.py
arcondello: The problem is checking the failure cases. MyPy doesn't really have a very specific exclude syntax.

docs/reference/index.rst
arcondello: Sure, can move it.

dimod/vartypes.py
arcondello: I am going to leave it alone for now - see what RTD does with it

PR #1163, opened 2022-04-06
Improve the performance of QM.update()

arcondello:
Provides a 60% speedup to `QM.update()` and `CQM.set_objective()`.

### DIFF
--- a/dimod/quadratic/cyqm/cyqm_template.pyx.pxi
+++ b/dimod/quadratic/cyqm/cyqm_template.pyx.pxi
@@ -20,13 +20,20 @@ cimport cython
 
 from cython.operator cimport preincrement as inc, dereference as deref
 from libc.math cimport ceil, floor
+from libcpp.vector cimport vector
 
 from dimod.binary.cybqm cimport cyBQM
 from dimod.cyutilities cimport as_numpy_float, ConstInteger, ConstNumeric, cppvartype
 from dimod.libcpp cimport cppvartype_info
+from dimod.quadratic cimport cyQM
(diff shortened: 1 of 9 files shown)


dimod/quadratic/cyqm/cyqm_template.pyx.pxi
```diff
@@ -705,6 +712,71 @@ cdef class cyQM_template(cyQMBase):
...
+                mapping.push_back(self.variables.index(v))
+
+                if self.cppqm.vartype(mapping[vi]) != other.data().vartype(vi):
+                    raise ValueError(f"conflicting vartypes: {v!r}")
+
+                if self.cppqm.lower_bound(mapping[vi]) != other.data().lower_bound(vi):
+                    raise ValueError(f"conflicting lower bounds: {v!r}")
+
+                if self.cppqm.upper_bound(mapping[vi]) != other.data().upper_bound(vi):
+                    raise ValueError(f"conflicting upper bounds: {v!r}")
```
randomir, line 737: I'm parachuting into this PR, but I would naively expect an `update` method in Python to silently overwrite/upgrade target from source. Perhaps warn at most.

randomir (approved):
LGTM!

dimod/quadratic/cyqm/cyqm_template.pyx.pxi
arcondello: Eh... we don't really allow too much manipulation of the bounds once they are set. Also there are a bunch of places that require this consistency. So IMO better to keep this behavior.

PR #1166, opened 2022-04-12
Use cibuildwheel 2.4.0

arcondello:


.circleci/config.yml
```diff
@@ -30,13 +30,19 @@ jobs:
       - restore_cache: &build-linux-restore-cache
           keys:
             - pip-{{ .Environment.CIRCLE_JOB }}-{{ checksum "pyproject.toml" }}
+      - when:
+          condition:
+            equal: [3.10.0, << parameters.python-version>>]
```
randomir, line 35:
```suggestion
            matches:
              pattern: "^3\\.10.*$"
              value: << parameters.python-version >>
```

.circleci/config.yml
```diff
@@ -378,6 +384,7 @@ workflows:
       - build-linux: &build
           matrix:
             parameters:
+              # if 3.10 version changes remember to update build-linux
```
randomir, line 387: No need if we switch to regex match (see suggestion above).

randomir (approved):
LGTM!

PR #1172, opened 2022-04-18
Add a scaling guide to the docs

arcondello:


arcondello:
Ok, I just don't think nbsphinx is mature enough. I will remake this PR without it.

JoelPasvolsky:
Are you seeing problems in CircleCI that you did not have when building locally?

arcondello:
Yes, for instance you need pandoc installed for it to work. It's not such a huge dependency, but it's just one of a whole pile of dependencies needed.

It also doesn't play that nicely with pydata-sphinx-theme.

IMO, let's wait 6 months to see if it gets more adopted.

JoelPasvolsky:
I'm not sure about keeping the JN here now: it means duplicate updating and is not tested.

arcondello:
> I'm not sure about keeping the JN here now: it means duplicate updating and is not tested.

To be fair, the code itself is not tested either. I am fine to remove it, or I can add testing to circleci. IMO the latter is preferred for maintainability

JoelPasvolsky (approved):

PR #1174, opened 2022-04-21
Fix negative indexing in Variables objects

arcondello:


hbarovertwo (approved):
Pulled down and tested before and after, LGTM!

PR #1176, opened 2022-04-21
Drop support for NumPy 1.19

arcondello:
This allows us to build all version of dimod with the same NumPy version which simplifies thing.


randomir (approved):
LGTM!

PR #1180, opened 2022-04-25
Speed up iteration over Variables objects

arcondello:
New timing results for `for v in variables: pass`
```bash
                labels              
              --------- ------------
                string   34.3±0.7μs 
                index    11.2±0.3μs 
               integer   34.1±0.6μs 
```
vs the old timing results
```bash
                labels              
              --------- ------------
                string   63.0±0.3μs 
                index     53.1±1μs  
               integer    62.2±3μs  
```

### DIFF
--- a/benchmarks/variables.py
+++ b/benchmarks/variables.py
@@ -12,15 +12,32 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
-import dimod
+from dimod.variables import Variables
+
+
+class TimeConstuction:
+    num_variables = 1000
+
+    iterables = dict(range=range(num_variables),
+                     strings=list(map(str, range(num_variables))),
+                     integers=list(range(1000)),
+                     empty=[],
+                     none=None,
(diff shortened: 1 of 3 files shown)


juansebastianl (approved):
Looks good to me! Interesting that this is faster, but I guess it takes advantage of the computation in the `_append` method in the constructor.

arcondello:
In this case the speed comes from implementing `.__iter__()` "by hand" rather than using the one inherited from `collections.abc.Sequence` which looks like
```python
    def __iter__(self):
        i = 0
        try:
            while True:
                v = self[i]
                yield v
                i += 1
        except IndexError:
            return
```
([source code](https://github.com/python/cpython/blob/971343eb569a3418aa9a0bad9b638cccf1470ef8/Lib/_collections_abc.py#L1034)).

By doing it "by hand" we can improve the performance in the range case, and even in the case with labels we can bypass a bunch of function calls.

Issue #928, opened 2021-07-27
Add a function to release memory of quadratic model objects.


amahmudwave:
**Application**
Saves memory when we are working in a memory constrained situation.

**Proposed Solution**
We can swap the vectors inside the object with empty vectors. This will release the memory. If we resize the vectors or clear them, the allocated memory may not be released.

Example code : std::vector<your_vector_element_type>().swap(your_vector);

**Alternatives Considered**
Calling delete on the quadratic model objects.

**Additional Context**
The quadratic model objects have limited use and during its usage other memory allocations may happen, so once a quadratic model is used and if there is no more use for the object, but we still need the parent object in which the quadratic model resides in, we can release the memory held by the quadratic model only.

Issue #1132, opened 2022-03-01
Add support for coefficients in scientific notation to LP file parser


randomir:
Currently `CQM.from_lp_file` (using grammar from `make_lp_grammar`) supports coefficients expressed as decimal numbers:

https://github.com/dwavesystems/dimod/blob/1f634676635b6df17c70e589af87556a83fc4084/dimod/serialization/lp.py#L68

but not scientific/exponential notation, for example:
```
Minimize
  obj: 2e3 x0 + 2.1e-04 x0^2
Subject To
  x0 <= 1
General
  x0
End
```
or
```
Minimize
  obj: 2e3 x0 + [4.1e-02 x0*x0]/2
Subject To
  x0 <= 1
General
  x0
End
```

Motivation: [QPLIB](https://qplib.zib.de/) contains problems with coefficients expressed in scientific notation.

arcondello:
Putting this here so I don't forget, it looks like SciPy might be addings support for reading LP files https://github.com/scipy/scipy/issues/14593. It also appears that they use [HiGHS](https://github.com/ERGO-Code/HiGHS) for their support.

randomir:
No one seems to be working on that SciPy feature, so it might be worth considering contributing #1182 (i.e. some variant of it) to SciPy?

arcondello:
> No one seems to be working on that SciPy feature, so it might be worth considering contributing #1182 (i.e. some variant of it) to SciPy?

See response https://github.com/dwavesystems/dimod/pull/1182#issuecomment-1117474634 tldr; yes but need a few things to happen first

PR #1182, opened 2022-05-03
Improve the performance of LP file parsing

arcondello:
Adds a C++ LP file reader from HiGHS (https://github.com/ERGO-Code/HiGHS/tree/501b30a) and wraps it.

So far just a DRAFT to see what CI makes of this. There is a bunch of docs/cleanup still to be done.

Closes https://github.com/dwavesystems/dimod/issues/928


tests/test_serialization_lp.py
```diff
@@ -0,0 +1,154 @@
...
+
+        x, y = dimod.Reals('xy')
+
+        cqm = read_lp(lp)
+
+        self.assertFalse(cqm.constraints)
+        self.assertTrue(cqm.objective.is_equal(x + y))
+
+    # todo
+    # def test_scientific_notation(self):
```
randomir, line 102: :tada: (closes #1132)

randomir:
How about you contribute this as a PR to https://github.com/scipy/scipy/issues/14593 instead? SciPy seems to [already include HiGHS](https://github.com/scipy/scipy/pull/15217).

arcondello:
> How about you contribute this as a PR to [scipy/scipy#14593](https://github.com/scipy/scipy/issues/14593) instead? SciPy seems to [already include HiGHS](https://github.com/scipy/scipy/pull/15217).

I thought about it, but a few reasons to not:
1) This way we don't make any intermediate objects, which is important for large models
2) To access it from the SciPy namespace, HiGHS would need to partly restructure their code (possible can be gotten around)
3) There are a few bug fixes that would also need to be made to HiGHs. Once https://github.com/ERGO-Code/HiGHS/issues/840, https://github.com/ERGO-Code/HiGHS/pull/841, and https://github.com/ERGO-Code/HiGHS/pull/839 are resolved we could proceed
5) Would need HiGHS to also support scientific notation
6) Would need SciPy to then use the latest HiGHS
7) We would need to add SciPy as a dependency to dimod, as well as wait for it to deploy. That would also have downstream consequences on the range of NumPy versions we can support

That said, only (1) is a long-term problem. Once (3) and (5) are resolved, I'll hopefully have time to put together a PR

arcondello:
There is a bit more test restructure that could be done, but I am eager to release 0.11 and want to get this in. In addition to the current tests, I tested it on the entirety of [qplib](https://qplib.zib.de/)

releasenotes/notes/feature-c++-lp-reader-1d5109d86ce778ed.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - Add ``read_lp_file()`` and ``read_lp()`` functions.
+  - Improve the performance of the ``ConstrainedQuadraticModel.from_lp_file()`` method.
+upgrade:
+  - |
+    Some malformed LP-files that previously caused ``ConstrainedQuadraticModel.from_lp_file()``
+    to emit a warning now raise an error instead.
+deprecations:
+  - |
+    Deprecate ``ConstrainedQuadraticModel.from_lp_file()`` method. Users should
+    use the more explicit ``read_lp_file()`` and ``read_lp()`` functions instead.
```
randomir, line 12: Shouldn't users now use `dimod.lp.load` and `dimod.lp.loads`?

randomir: Oh, I see you fixed it in a follow-up commit.

lesshaste:
> > How about you contribute this as a PR to [scipy/scipy#14593](https://github.com/scipy/scipy/issues/14593) instead? SciPy seems to [already include HiGHS](https://github.com/scipy/scipy/pull/15217).
> 
> I thought about it, but a few reasons to not:
> 
> 1. This way we don't make any intermediate objects, which is important for large models
> 2. To access it from the SciPy namespace, HiGHS would need to partly restructure their code (maybe)
> 3. There are a few bug fixes that would also need to be made to HiGHs. Once [Trying to load an LP file without a newline at the end of the file leads to out-of-memory errors ERGO-Code/HiGHS#840](https://github.com/ERGO-Code/HiGHS/issues/840), [Fix EOF detection in LP file reader ERGO-Code/HiGHS#841](https://github.com/ERGO-Code/HiGHS/pull/841), and [Allow LP file parse to handle var*var in expressions ERGO-Code/HiGHS#839](https://github.com/ERGO-Code/HiGHS/pull/839) are resolved we could proceed
> 4. Would need HiGHS to also support scientific notation
> 5. Would need SciPy to then use the latest HiGHS
> 6. We would need to add SciPy as a dependency to dimod, as well as wait for it to deploy. That would also have downstream consequences on the range of NumPy versions we can support
> 
> That said, only (1) is a long-term problem. Once (3), (4), and (5) are resolved, I'll hopefully have time to put together a PR

It looks like 3 and 5 at least have happened now.

PR #1183, opened 2022-05-03
Bug fix nonzero lagrange multiplier

hsadeghidw:
If the objective is empty, set the Lagrange to one.

### DIFF
--- a/releasenotes/notes/set-penatly-to-one-zero-objective-116b50b47ad41051.yaml
+++ b/releasenotes/notes/set-penatly-to-one-zero-objective-116b50b47ad41051.yaml
@@ -1,0 +1,5 @@
+---
+fixes:
+  - |
+    Set the lagrange multiplier to one if the maximum absolute bias is zero
+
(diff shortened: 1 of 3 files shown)


releasenotes/notes/set-penatly-to-one-zero-objective-116b50b47ad41051.yaml
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    Set the lagrange multiplier to one if the maximum absolute bias is zero
```
arcondello, line 4: can you add `... in the ``cqm_to_bqm()`` function.` to the end of this?

arcondello (approved):
Just need to make the release note a bit clearer so it's obvious what it refer to in context. Otherwise looks great

releasenotes/notes/set-penatly-to-one-zero-objective-116b50b47ad41051.yaml
JoelPasvolsky: Also if you could capitalize: "the Lagrange multiplier"

PR #1191, opened 2022-05-25
Add a generator for not-all-equal 3-satisfiability problems

arcondello:


dimod/generators/satisfiability.py
```diff
@@ -0,0 +1,137 @@
...
+        yield (y, z, ys*zs)
+
+
+def nae3sat(variables: typing.Union[int, typing.Sequence[dimod.typing.Variable]],
+            rho: float = 2.1,
+            *,
+            seed: typing.Union[None, int, np.random.Generator] = None,
+            replace: bool = True,
+            ) -> BinaryQuadraticModel:
+    """Generator for Not-All-Equal 3-SAT (NAE3SAT) Binary Quadratic Models.
```
JoelPasvolsky, line 89: binary quadratic models

dimod/generators/satisfiability.py
```diff
@@ -0,0 +1,137 @@
...
+
+    NAE3SAT_ is an NP-complete problem class that consists in satistying a number of conjunctive
+    clauses that involve three variables (or variable negations). The variables on each clause
```
JoelPasvolsky, line 92: clauses of three variables (or variable negations) each,

dimod/generators/satisfiability.py
```diff
@@ -0,0 +1,137 @@
...
+    should be not-all-equal. Ie. all solutions except ``(+1, +1, +1)`` or
```
JoelPasvolsky, line 93: For valid solutions, the variables of each clause should not all be equal; i.e., any assignment of values except ``(+1, +1, +1)`` or ``(-1, -1, -1)``.

dimod/generators/satisfiability.py
```diff
@@ -0,0 +1,137 @@
...
+            to generate the clauses and the negations.
+        replace: If true, then clauses are randomly sampled from the space
+            of all possible clauses. This can result in the same three variables
+            being present in multiple clauses.
+            As the number of variables grows the probability of this happening
+            shrinks rapidly and therefore it is often better to allow sampling
+            with replacement for performance.
+
+    Returns:
+        A binary quadratic models with spin variables.
```
JoelPasvolsky, line 111:
```suggestion
        A binary quadratic model with spin variables.
```

dimod/generators/satisfiability.py
```diff
@@ -0,0 +1,137 @@
...
+    should be not-all-equal. Ie. all solutions except ``(+1, +1, +1)`` or
+    ``(-1, -1, -1)`` are valid for each class.
+
+    .. _NAE3SAT: https://en.wikipedia.org/wiki/Not-all-equal_3-satisfiability
+
+    Args:
+        num_variables: The number of variables in the problem.
+        rho: The clause-to-variable ratio.
+        seed: Passed to :func:`numpy.random.default_rng()`, which is used
+            to generate the clauses and the negations.
```
JoelPasvolsky, line 102: why the "and the negations" addition to "clauses"? the statement doesn't talk about variables

dimod/generators/satisfiability.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Returns:
+        A binary quadratic models with spin variables.
+
+    """
+    if isinstance(variables, collections.abc.Sequence):
+        num_variables = len(variables)
+        labels = variables
+    elif variables < 0:
+        raise ValueError("variables must be a sequence or a positive integer")
```
JoelPasvolsky, line 118:
```suggestion
    elif variables < 3:
        raise ValueError("variables must be a sequence or a positive integer of at least size 3")
```

Currently you can set a size of 2 and you'll get some unhelpful error from ``rng.choice()``

dimod/generators/satisfiability.py
```diff
@@ -0,0 +1,137 @@
...
+    NAE3SAT_ is an NP-complete problem class that consists in satistying a number of conjunctive
+    clauses that involve three variables (or variable negations). The variables on each clause
+    should be not-all-equal. Ie. all solutions except ``(+1, +1, +1)`` or
+    ``(-1, -1, -1)`` are valid for each class.
+
+    .. _NAE3SAT: https://en.wikipedia.org/wiki/Not-all-equal_3-satisfiability
+
+    Args:
+        num_variables: The number of variables in the problem.
+        rho: The clause-to-variable ratio.
```
JoelPasvolsky, line 100: You round this in line 123, and I think it worth noting that here

jackraymond:
We should cite our existing (original) publications on NAE3SAT and MCSAT in the docstrings:
https://link.springer.com/chapter/10.1007/978-3-319-24318-4_9
I think it would be worth adding a generator for TwoinFourSAT given the importance of that special case of MCSAT.
It's quite straightfoward to add solution planting, for k-MCSAT and k odd (e.g. 2in4SAT special case) this planting is 'quiet', meaning that planting a solution in the SAT phase does not change the large-scale (interesting) properties of the energy landscape.

arcondello:
> We should cite our existing (original) publications on NAE3SAT and MCSAT in the docstrings: https://link.springer.com/chapter/10.1007/978-3-319-24318-4_9
Will do

> I think it would be worth adding a generator for TwoinFourSAT given the importance of that special case of MCSAT.
Agree, though IMO that can come in a followup PR

arcondello:
@jackraymond or @pau557 , do you think we should apply an offset and scale? So the energy of the samples corresponds to the number of violated clauses? I.e. use
```python
>>> bqm = dimod.BinaryQuadraticModel({0: 0.0, 1: 0.0, 2: 0.0}, {(1, 0): .25, (2, 0): .25, (2, 1): .25}, .25, 'SPIN')
>>> print(dimod.ExactSolver().sample(bqm))
   0  1  2 energy num_oc.
1 -1 +1 -1    0.0       1
2 +1 +1 -1    0.0       1
3 +1 -1 -1    0.0       1
4 +1 -1 +1    0.0       1
6 -1 +1 +1    0.0       1
7 -1 -1 +1    0.0       1
0 -1 -1 -1    1.0       1
5 +1 +1 +1    1.0       1
['SPIN', 8 rows, 8 samples, 3 variables]
```
rather than
```python
>>> bqm = dimod.BinaryQuadraticModel({0: 0.0, 1: 0.0, 2: 0.0}, {(1, 0): 1, (2, 0): 1, (2, 1): 1}, 0, 'SPIN')
>>> print(dimod.ExactSolver().sample(bqm))
   0  1  2 energy num_oc.
1 -1 +1 -1   -1.0       1
2 +1 +1 -1   -1.0       1
3 +1 -1 -1   -1.0       1
4 +1 -1 +1   -1.0       1
6 -1 +1 +1   -1.0       1
7 -1 -1 +1   -1.0       1
0 -1 -1 -1    3.0       1
5 +1 +1 +1    3.0       1
['SPIN', 8 rows, 8 samples, 3 variables]
```

jackraymond:
> @jackraymond or @pau557 , do you think we should apply an offset and scale? So the energy of the samples corresponds to the number of violated clauses? I.e. use
> 
> ```python
> >>> bqm = dimod.BinaryQuadraticModel({0: 0.0, 1: 0.0, 2: 0.0}, {(1, 0): .25, (2, 0): .25, (2, 1): .25}, .25, 'SPIN')
> >>> print(dimod.ExactSolver().sample(bqm))
>    0  1  2 energy num_oc.
> 1 -1 +1 -1    0.0       1
> 2 +1 +1 -1    0.0       1
> 3 +1 -1 -1    0.0       1
> 4 +1 -1 +1    0.0       1
> 6 -1 +1 +1    0.0       1
> 7 -1 -1 +1    0.0       1
> 0 -1 -1 -1    1.0       1
> 5 +1 +1 +1    1.0       1
> ['SPIN', 8 rows, 8 samples, 3 variables]
> ```
> 
> rather than
> 
> ```python
> >>> bqm = dimod.BinaryQuadraticModel({0: 0.0, 1: 0.0, 2: 0.0}, {(1, 0): 1, (2, 0): 1, (2, 1): 1}, 0, 'SPIN')
> >>> print(dimod.ExactSolver().sample(bqm))
>    0  1  2 energy num_oc.
> 1 -1 +1 -1   -1.0       1
> 2 +1 +1 -1   -1.0       1
> 3 +1 -1 -1   -1.0       1
> 4 +1 -1 +1   -1.0       1
> 6 -1 +1 +1   -1.0       1
> 7 -1 -1 +1   -1.0       1
> 0 -1 -1 -1    3.0       1
> 5 +1 +1 +1    3.0       1
> ['SPIN', 8 rows, 8 samples, 3 variables]
> ```

Normalizing so that a single violation has cost 1, and so that the ground state has zero energy, makes sense for NAE3SAT in my opinion given that the normalization still allows for use of machine numbers for all couplers.

arcondello:
> This doesn't work (exactly) in generalization to k-MCSAT, since in that case different violation patterns have different energies - so perhaps leave MCSAT as is.

That's a good point. Maybe I am missing something but doesn't the typical implementation of 2-in-4 sat have different energies for different excited states? I.e.
```python
>>> bqm = dimod.BQM({}, {(0, 1): 1, (0, 2): 1, (0, 3): 1, (1, 2): 1, (1, 3): 1, (2, 3): 1}, 0, 'SPIN')
>>> print(dimod.ExactSolver().sample(bqm))
    0  1  2  3 energy num_oc.
2  +1 +1 -1 -1   -2.0       1
4  -1 +1 +1 -1   -2.0       1
6  +1 -1 +1 -1   -2.0       1
8  -1 -1 +1 +1   -2.0       1
12 -1 +1 -1 +1   -2.0       1
14 +1 -1 -1 +1   -2.0       1
1  +1 -1 -1 -1    0.0       1
3  -1 +1 -1 -1    0.0       1
5  +1 +1 +1 -1    0.0       1
7  -1 -1 +1 -1    0.0       1
9  +1 -1 +1 +1    0.0       1
11 -1 +1 +1 +1    0.0       1
13 +1 +1 -1 +1    0.0       1
15 -1 -1 -1 +1    0.0       1
0  -1 -1 -1 -1    6.0       1
10 +1 +1 +1 +1    6.0       1
['SPIN', 16 rows, 16 samples, 4 variables]
```
and doesn't that lead to non-uniform sampling since we expect some to be violated?

arcondello:
Put another way, should we only use that formulation for 2-in-4 when we _know_ that the problem is satisfiable.

jackraymond:
> Put another way, should we only use that formulation for 2-in-4 when we _know_ that the problem is satisfiable.

It is not possible to formulate a 2in4SAT clause as a bqm with only two energy levels. For 2in4SAT there must be atleast 2 different penalties for constraint-violating assignments. The way its set up (basically H = (sum_i x_i)^2 ) is the most sensible and compact way to represent as a bqm. 
However, its worth noting that Max 2in4SAT (for the case where SAT is not possible) may return an incorrect result - since it may favour violating 2 clauses (by small amounts) over violating 1 clause (by a large amount). So important to document this somehow. There's not much that can be done about this, short of introducing auxiliary variables that have their own problems.

pau557:
> > This doesn't work (exactly) in generalization to k-MCSAT, since in that case different violation patterns have different energies - so perhaps leave MCSAT as is.
> 
> That's a good point. Maybe I am missing something but doesn't the typical implementation of 2-in-4 sat have different energies for different excited states? I.e.
> 
> ```python
> >>> bqm = dimod.BQM({}, {(0, 1): 1, (0, 2): 1, (0, 3): 1, (1, 2): 1, (1, 3): 1, (2, 3): 1}, 0, 'SPIN')
> >>> print(dimod.ExactSolver().sample(bqm))
>     0  1  2  3 energy num_oc.
> 2  +1 +1 -1 -1   -2.0       1
> 4  -1 +1 +1 -1   -2.0       1
> 6  +1 -1 +1 -1   -2.0       1
> 8  -1 -1 +1 +1   -2.0       1
> 12 -1 +1 -1 +1   -2.0       1
> 14 +1 -1 -1 +1   -2.0       1
> 1  +1 -1 -1 -1    0.0       1
> 3  -1 +1 -1 -1    0.0       1
> 5  +1 +1 +1 -1    0.0       1
> 7  -1 -1 +1 -1    0.0       1
> 9  +1 -1 +1 +1    0.0       1
> 11 -1 +1 +1 +1    0.0       1
> 13 +1 +1 -1 +1    0.0       1
> 15 -1 -1 -1 +1    0.0       1
> 0  -1 -1 -1 -1    6.0       1
> 10 +1 +1 +1 +1    6.0       1
> ['SPIN', 16 rows, 16 samples, 4 variables]
> ```
> 
> and doesn't that lead to biased sample since we expect some to be violated in the ground state? Or do we expect to find a feasible solution usually?

That's correct. One second-exited state is worth 4 first-excited ones. In kMCSAT we're more in the Ising/QUBO benchmarking arena rather than conjuctive-clause benchmarking. That's fine with me, it's a common issue when mapping optimization problems to QUBOS.

dimod/generators/satisfiability.py
```diff
@@ -0,0 +1,137 @@
...
+
+    try:
+        max_num_clauses = math.comb(num_variables, 3)
+    except AttributeError:
+        # Python < 3.8
+        # This is a bit less picky and has overflow issues but
+        # probably fine for any cases users are likely to actually
+        # submit
+        max_num_clauses = math.factorial(num_variables)
+        max_num_clauses //= (6 * math.factorial(num_variables - 3))
```
pau557, line 60:
```suggestion
max_num_clauses = num_variables*(num_variables - 1)*(num_variables - 2) // 6
```

tests/test_generators.py
```diff
@@ -895,3 +896,44 @@ def test_empty_networkx(self):
         self.assertEqual(
             dimod.generators.maximum_weight_independent_set(G.edges, G.nodes('weight')).shape,
             (0, 0))
+
+
+class TestSatisfiability(unittest.TestCase):
+    def test_empty(self):
+        bqm = dimod.generators.nae3sat(0)
```
pau557, line 903: contingent on @JoelPasvolsky comment about variables < 3

pau557 (approved):
LGTM

dimod/generators/satisfiability.py
jackraymond: A variable, or negation thereof, is called a literal, and that is very standard language. I would suggest:
'clauses of three literals (variables, or their negations) each,'
but not important.

dimod/generators/satisfiability.py
```diff
@@ -0,0 +1,137 @@
...
+
+    if num_clauses > max_num_clauses:
+        raise ValueError("rho results in too many clauses")
```
jackraymond, line 63: This error should only be displayed when replace=False I guess?

jackraymond (changes requested):
Looks good.
Big,not required: Personally I would like to see the BQM implementation of NAE3SAT and all of its helper functions defined as a special case of k-MCSAT, but this code is mostly fine for now. Then we just change 1 argument (k) to get 2in4SAT, 3in6SAT etc. These classes have interesting and complementary features to NAE3SAT.
Add citation: Cite this paper in docs: https://link.springer.com/chapter/10.1007/978-3-319-24318-4_9 . First D-Wave research in the area, and complements definitions.
Docs - use 'literals': For documentation purposes replace: 'variables or their negations' with 'literals' (per standard usage https://en.wikipedia.org/wiki/Not-all-equal_3-satisfiability).
Define generator with num_clauses (signed int), make rho an optional argument.
The option replace has some problems (e.g. variable collisions are checked but literal collisons are not), my opinion it should just be removed since collisons are allowed as part of standard definitions, and in any case are extremely rare.

dimod/generators/satisfiability.py
```diff
@@ -0,0 +1,137 @@
...
+        seen.add(fz)
+
+        xs, ys, zs = 2 * rng.integers(0, 1, endpoint=True, size=3) - 1
+
+        yield (x, y, xs*ys)
+        yield (x, z, xs*zs)
+        yield (y, z, ys*zs)
+
+
+def nae3sat(variables: typing.Union[int, typing.Sequence[dimod.typing.Variable]],
```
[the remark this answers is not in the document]
pau557: I suggested using rho instead of num_clauses since that is the indicator of the problem class. Otherwise the user would need to add boilerplate when generating problems. However, @jackraymond  is the expert on this problem class, so I ultimately defer to him.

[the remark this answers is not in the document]
pau557: fwiw, we went with deterministic num_clauses from rho (rounding num_variables*rho) instead of the random distribition approach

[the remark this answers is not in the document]
jackraymond: Personally, I think its easier and more intuitive to define the class in terms of a fixed integer number of clauses, rather than a random number of clauses (determined from distribution parameterized by rho - which is buried deep in code and not clearly documented).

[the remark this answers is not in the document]
arcondello: yes, we'll just add using rho as an example in docs

jackraymond:
Atleast in docs, if not in the explicit function names etc, might be worth calling the problem class 'Random NAE3SAT', reflecting that its generator of random problems (this is also what it would be called in academic arena). We use this sort of nomenclature implicitely elsewhere (e.g. RAN1).

dimod/generators/satisfiability.py
```diff
@@ -0,0 +1,137 @@
...
+    ``(-1, -1, -1)`` are valid for each class.
+
+    .. _NAE3SAT: https://en.wikipedia.org/wiki/Not-all-equal_3-satisfiability
+
+    Args:
+        num_variables: The number of variables in the problem.
+        rho: The clause-to-variable ratio.
+        seed: Passed to :func:`numpy.random.default_rng()`, which is used
+            to generate the clauses and the negations.
+        replace: If true, then clauses are randomly sampled from the space
```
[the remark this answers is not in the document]
JoelPasvolsky: What are your feelings about adding a ``return_sat`` parameter that when true returns the generating sat, maybe as a list of 3-tuples like ``[("a", 1-"c", "e"), (1-"b"), "f", "a"), ...]` also? Be nice to have that for small problems to understand what the BQM represents, for easy validation, and could enable comparisons with sat solvers.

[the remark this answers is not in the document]
jackraymond: In short yes, its useful to have the more standard presentation for intuition sakes, and for conversion to other (non BQM) formats.

boothby:
Using the [Combinatorial Number System](https://en.wikipedia.org/wiki/Combinatorial_number_system), one can leverage python's accelerated `random.sample(range(n), k)` to generate random sets of triples without replacements

```python
def binom_2(n):
    return n*(n-1)//2
def binom_3(n):
    return n*(n-1)*(n-2)//6

from math import sqrt
def unrank_triple(N):
    if N == 0:
        return (0, 1, 2)

    #find the floor of the unique real root of N = (n choose 3)
    T = (3*N + 1/3*sqrt(81*N*N - 1/3))**(1/3)
    C1 = int(T + 1/3/T + 1)
    N -= binom_3(C1)
    
    if N == 0:
        return (0, 1, C1)

    #find the floor of the unique real root of N == (n choose 2)
    C2 = int((sqrt(8*N+1) + 1)//2)

    C3 = N - binom_2(C2)
    return (C3, C2, C1)

def rank_triple(t):
    #use this for testing unrank_triple
    a, b, c = sorted(t)
    return a + binom_2(b) + binom_3(c)

def random_triples_without_replacement(n, k):
    return [
        unrank_triple(N)
        for N in random.sample(range(binom_3(n)), k)
    ]
```

pau557:
👍

PR #1192, opened 2022-05-27
Simplify install instructions

arcondello:
...and remove the hardcoded python version reference.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -59,7 +59,7 @@ Installation
 
 .. installation-start-marker
 
-Installation from [PyPI](https://pypi.org/project/dimod/):
+Installation from `PyPI <https://pypi.org/project/dimod>`_:
 
 .. code-block:: bash
 
--- a/README.rst
+++ b/README.rst
@@ -59,26 +59,11 @@ Installation
 
 .. installation-start-marker
 
-Compatible with Python 3.6+:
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -59,27 +59,12 @@ Installation
 
 .. installation-start-marker
 
-Compatible with Python 3.6+:
+Installation from [PyPI](https://pypi.org/project/dimod/):
```
JoelPasvolsky, line 62:
```suggestion
Installation from `PyPI <https://pypi.org/project/dimod>`_:
```

This is RST format

JoelPasvolsky (commented):
Why don't we want installation from source anymore?

arcondello:
> Why don't we want installation from source anymore?

Under the assumption that anyone who wants to make a contribution probably understands `pip` and `setuptools`. If we see a lot of confusion, then we can restore it.

JoelPasvolsky:
> > Why don't we want installation from source anymore?
> 
> Under the assumption that anyone who wants to make a contribution probably understands `pip` and `setuptools`. If we see a lot of confusion, then we can restore it.

okay. I used to look here for a reminder on the commands but it's not hard to find that elsewhere

JoelPasvolsky (approved):

PR #1194, opened 2022-05-28
Added planting option for random_nae3sat and random_2in4sat, and minor changes to documentation

jackraymond:
Added planting to random_nae3sat, random_2in4sat, and random_kmcsat problem generators.
Planting in random constraint satisfaction problems is useful for benchmarking, particularly in the SAT phase - problems can be generated with a guaranteed ground state without impacting hardness properties of the problem class. This is useful in benchmarking where knowing the ground state allows unambiguous tests of optimality. Planting is also important with respect to the SAT filter application.
Minor changes to docstrings independent of the planting option are also implemented.

### DIFF
--- a/dimod/generators/fcl.py
+++ b/dimod/generators/fcl.py
@@ -12,13 +12,13 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
-import math
 import warnings
 
 from typing import Callable, Collection, List, Mapping, Optional
 
 import numpy.random
 
+from numpy import prod
 from dimod.binary_quadratic_model import BinaryQuadraticModel
 from dimod.decorators import graph_argument
 from dimod.typing import GraphLike, Variable
@@ -142,7 +142,7 @@ def frustrated_loop(graph: GraphLike,
         else:
             # randomly select from all frustrated loops (odd number of AFM edges)
             cycle_J = {(cycle[i], cycle[i+1]) : -1 for i in range(len(cycle)-1)}
-            cycle_J[(cycle[-1],cycle[0])] = (1 - 2*(len(cycle_J) & 1))*math.prod(cycle_J.values())
+            cycle_J[(cycle[-1],cycle[0])] = (1 - 2*(len(cycle_J) & 1))*prod(list(cycle_J.values()))
             
         # update the bqm
         bqm.add_interactions_from(cycle_J)
@@ -161,10 +161,11 @@ def frustrated_loop(graph: GraphLike,
                       "spin labels to match desired planting or "
                       "(2) SpinReversalComposite for presentation of a randomized "
                       "planted solution to some solver.",
-                      DeprecationWarning)
+                      DeprecationWarning, stacklevel=3)
         # A spin-reversal transform for a BQM with zero linear biases
         for e in bqm.quadratic:
-            bqm.quadratic[e] = bqm.quadratic[e]*planted_solution[e[0]]*planted_solution[e[1]]
+            bqm.set_quadratic(e[0], e[1],
+                              bqm.quadratic[e]*planted_solution[e[0]]*planted_solution[e[1]])
 
     return bqm
 
--- a/dimod/generators/fcl.py
+++ b/dimod/generators/fcl.py
@@ -12,6 +12,9 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+import math
+import warnings
+
 from typing import Callable, Collection, List, Mapping, Optional
 
 import numpy.random
@@ -33,18 +36,21 @@ def frustrated_loop(graph: GraphLike,
                     R: float = float('inf'),
                     cycle_predicates: Collection[Predicate] = tuple(),
                     max_failed_cycles: int = 100,
+                    plant_solution: bool = True,
                     planted_solution: Optional[Mapping[Variable, int]] = None,
                     seed: Optional[int] = None,
                     ) -> BinaryQuadraticModel:
     """Generate a frustrated-loop problem.
 
-    A generic frustrated-loop (FL) problem is a sum of Hamiltonians, each generated
-    from a single "good" loop, as follows:
+    A generic frustrated-loop (FL) problem is a sum of Hamiltonians, each 
+    generated from a single loop, as follows:
 
     1. Generate a loop by random walking on the specified graph, ``graph``.
     2. If the cycle meets provided predicates, continue; if not, return to  step 1.
-    3. Choose one edge of the loop to be anti-ferromagnetic (an interaction value
-       of `+1` for the edge) and all others ferromagnetic (`-1`).
+    3. Choose one edge of the loop uniformly at random to be anti-ferromagnetic 
+       (AFM, coupling value of `1`) and all others ferromagnetic 
+       (`-1`). Or if :code:`plant_solution is False`, sample uniformly couplers
+       subject to the constraint that there are an odd number of AFM couplers.
     4. Add the loop's interactions to the FL problem.
        If at any time the absolute value of an interaction in the FL problem,
        accumulated on an edge over good loops, exceeds a given maximum, ``R``,
@@ -73,10 +79,17 @@ def frustrated_loop(graph: GraphLike,
         max_failed_cycles:
             Maximum number of failures to find a cycle before terminating.
 
+        plant_solution: 
+            Select frustrated loops with only 1 AFM coupler all 1 (and all -1) 
+            solutions are among the ground states. 
+
         planted_solution:
-            Other solutions with the same energy may exist, but the energy value
-            of the (possibly degenerate) ground state will hold. If None,
-            planted_solution is: ``{v: -1 for v in graph}``
+            A dictionary assigning variables to spin states.  When provided, and
+            :code:`planted_solution=True` spin states are relabeled so that the
+            variable assignment becomes a planted ground state in place of the 
+            all 1 solution.  
+            This option is deprecated; use of spin-reversal transforms is 
+            recommended for this purpose.
 
         seed: Random seed.
 
@@ -94,8 +107,7 @@ def frustrated_loop(graph: GraphLike,
         raise ValueError("R should be a positive integer")
     if max_failed_cycles <= 0:
         raise ValueError("max_failed_cycles should be a positive integer")
-    if planted_solution is None:
-        planted_solution = {v: -1 for v in nodes}
+    
     r = numpy.random.RandomState(seed)
 
     adj = {v: set() for v in nodes}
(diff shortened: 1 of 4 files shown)


dimod/generators/satisfiability.py
```diff
@@ -31,17 +31,24 @@
 
 def _kmcsat_interactions(num_variables: int, k: int, num_clauses: int,
                          *,
+                         is_planted: bool = False,
```
arcondello, line 34: probably  `plant_solution: bool = False` or even just `plant: bool = False`  is fine, here and elsewhere.

dimod/generators/satisfiability.py
```diff
@@ -59,25 +67,39 @@ def random_kmcsat(variables: typing.Union[int, typing.Sequence[dimod.typing.Vari
     that consists in satisfying a number of conjunctive
     clauses of ``k`` literals (variables, or their negations).
     Each clause should encode a max-cut problem over the clause literals.
+    
+    Each clause contributes -:code:`k//2' to the energy when the clause is
```
arcondello, line 71: you're right, I forgot to adjust the offset/scale. You can feel free to add the code to do that so that satisfying solutions have 0 energy and violations (in the nae3sat case) add energy of 1. Or I can do that as a followup

arcondello, line 71: You can just use double ` `` ` for code snippets

dimod/generators/satisfiability.py
```diff
@@ -59,25 +67,39 @@ def random_kmcsat(variables: typing.Union[int, typing.Sequence[dimod.typing.Vari
     that consists in satisfying a number of conjunctive
     clauses of ``k`` literals (variables, or their negations).
     Each clause should encode a max-cut problem over the clause literals.
+    
+    Each clause contributes -:code:`k//2' to the energy when the clause is 
+    satisfied, and at least 0 when unsatisfied. The energy :math:`H(s)` for a 
+    spin assignment :math:`s` is thus lower bounded by :math:`E_{SAT}=-`:code:`k//2 * num_clauses`,
```
arcondello, line 73: and single `` ` `` for math (unless you're doing special characters like `\sum`)

dimod/generators/satisfiability.py
```diff
@@ -86,15 +108,19 @@ def random_kmcsat(variables: typing.Union[int, typing.Sequence[dimod.typing.Vari
     else:
         num_variables = variables
         labels = None
-
-    if num_variables < k:
+    if k < 1:
+        raise ValueError("Number of variables per clause must be atleast 1")
+    elif k > num_variables:
+        raise ValueError(f"Number of variables per clause must be at most {num_variables}")
+    elif num_variables < k:
```
arcondello, line 115: unless I am blind this is now redundant

tests/test_generators.py
```diff
@@ -946,7 +947,28 @@ def integers(self, *args, **kwargs):
                 # in the ground state they should not be all equal
                 ss = dimod.ExactSolver().sample(bqm)
                 self.assertEqual(sum(ss.first.sample.values()), 0)
-
+                
+    def test_planting_sat(self):
+        #NP-hard to prove successful planting (rule out lower energies), this test
+        #could fail coincidentally, but unlikely.        
+        num_var = 6 # test is 2^num_var, keep small.
```
arcondello, line 954: two spaces before comments (elsewhere as well)

tests/test_generators.py
```diff
@@ -946,7 +947,28 @@ def integers(self, *args, **kwargs):
...
+        num_clauses = 24 #Deep in UNSAT phase (num_clause/num_var>>2.1), very unlikely to be SAT by chance.
+        
+        bqm = dimod.generators.random_nae3sat(num_var, num_clauses, is_planted=True)
+        E_SAT = - num_clauses;
+        all_energies = bqm.energies((np.array(list(itertools.product([-1,1], repeat=num_var))),bqm.variables))
```
arcondello, line 959: space between -1 and 1

tests/test_generators.py
```diff
@@ -946,7 +947,28 @@ def integers(self, *args, **kwargs):
...
-
+                
+    def test_planting_sat(self):
+        #NP-hard to prove successful planting (rule out lower energies), this test
+        #could fail coincidentally, but unlikely.        
+        num_var = 6 # test is 2^num_var, keep small.
+        num_clauses = 24 #Deep in UNSAT phase (num_clause/num_var>>2.1), very unlikely to be SAT by chance.
+        
+        bqm = dimod.generators.random_nae3sat(num_var, num_clauses, is_planted=True)
+        E_SAT = - num_clauses;
```
arcondello, line 958: unneeded semicolon

tests/test_generators.py
```diff
@@ -946,7 +947,28 @@ def integers(self, *args, **kwargs):
...
+        all_energies = bqm.energies((np.array(list(itertools.product([-1,1], repeat=num_var))),bqm.variables))
+        self.assertEqual(np.min(all_energies),E_SAT)
+        self.assertEqual(all_energies[0],E_SAT) #all -1 state
```
arcondello, line 961: space after comma (here and elsewhere)

arcondello:
Recommend that you also get a `pep8` linter for your IDE for some of the spacing etc

dimod/generators/satisfiability.py
```diff
@@ -164,24 +203,39 @@ def random_2in4sat(variables: typing.Union[int, typing.Sequence[dimod.typing.Var
...
         collisions are allowed in standard problem definitions, are absent with
         high probability in interesting cases, and are almost always harmless
-        when they do occur.
+        when they do occur. Planting of a not all 1 ground state
+        can be achieved with a spin-reversal transform without loss of 
+        generality, planting can significantly modify the hardness of 
+        optimization problems. However, for large problems planting of a 
+        solution can be achieved (in the SAT phase) without modification of the
+        hardness qualities of the instance class. [#]_ 
+    .. [#] Lenka Zdeborova and Florent Krzakala,
```
arcondello, line 236: The placement of this citation (and others) looks a bit strange in the built docs 
image

would recommend moving the other citation down to the bottom with it.

Also nice to have a space between the previous note and the citation for visual clarity

arcondello:
You also need to add a [release note](https://github.com/dwavesystems/dimod#release-notes), something like 
```
Add ``planted_solution`` keyword argument to ``dimod.generators.random_nae3sat()`` and ``dimod.generators.random_2in4sat()`` functions.
```

jackraymond:
You mean 'planted_solution' for compatibility with other generators? Or plant_solution

dimod/generators/satisfiability.py
jackraymond: You mean 'planted_solution' for compatibility with other generators in the module. I'll use that.

dimod/generators/satisfiability.py
jackraymond: Current definition matches the literature we are citing, and has some other advantages (Ising models without a constant is fairly standard the main one). So I'll leave as is.

dimod/generators/satisfiability.py
arcondello: There is a distinction between `planted_solution: Mapping[Variable, int]` and `plant_solution: bool`. In the former case the user is providing it, in the latter it's an instruction to the function.

jackraymond: To make the generators more consistent would make sense to have a plant: bool and planted_solution:  Optional[Mapping[Variable, int]], for both fcl and kcsat(nae3sat/2in4SAT). plant defaulted to True in the former and False in the latter.
The way planting is dealt with in frustrated_loop() is a bit inefficient and over complicated. Would be clearer and more efficient to plant 1 and apply an SRT at the end if planted_solution is not None. Might make this change unless there is an objection.

arcondello: That's fine. I would even go so far as:
1) Make this one `plant_solution: bool`
2) Add fcl `plant_solution: bool`
3) Throw a deprecation warning if `planted_solution` is provided, and suggest using SRTs (though keep the functionality for now for backwards compatibility)

jackraymond: That's probably the outcome I prefer. I think we should encourage people to use SRTs for planted instances - pushing the randomness into the solver rather than the instance class (where it is a superficial feature). Some solvers are SRT dependent and others not, we should encourage users to notice this.

dimod/generators/fcl.py
```diff
@@ -122,17 +134,16 @@ def frustrated_loop(graph: GraphLike,
...
-
+        if plant_solution:
+            # randomly select from all frustrated loops with one AFM edge
+            cycle_J = {(cycle[i - 1], cycle[i]) : -1 for i in range(len(cycle))}
+            idx = r.randint(len(cycle))
+            cycle_J[(cycle[idx - 1], cycle[idx])] = 1
+        else:
+            # randomly select from all frustrated loops (odd number of AFM edges)
+            cycle_J = {(cycle[i], cycle[i+1]) : -1 for i in range(len(cycle)-1)}
+            cycle_J[(cycle[-1],cycle[0])] = (1 - 2*(len(cycle_J) & 1))*math.prod(cycle_J.values())
```
arcondello, line 145: unfortunately [math.prod](https://docs.python.org/3/library/math.html#math.prod) only came in in Python 3.8. You could use [np.prod](https://numpy.org/doc/stable/reference/generated/numpy.prod.html) instead.

dimod/generators/fcl.py
```diff
@@ -144,6 +155,17 @@ def frustrated_loop(graph: GraphLike,
...
+    if planted_solution is not None:
+        warnings.warn("planted_solution is deprecated; use spin-reversal transforms: "
+                      "(1) bqm.flip_variable() for superficial reorientation of "
+                      "spin labels to match desired planting or "
+                      "(2) SpinReversalComposite for presentation of a randomized "
+                      "planted solution to some solver.",
+                      DeprecationWarning)
+        # A spin-reversal transform for a BQM with zero linear biases
+        for e in bqm.quadratic:
+            bqm.quadratic[e] = bqm.quadratic[e]*planted_solution[e[0]]*planted_solution[e[1]]
```
arcondello, line 167: doesn't really matter, but there is a slight performance benefit to using `bqm.set_quadratic()` rather than `bqm.quadratic[]`

dimod/generators/fcl.py
```diff
@@ -144,6 +155,17 @@ def frustrated_loop(graph: GraphLike,
         raise RuntimeError(f"number of found cycles, {good_cycles}, is below " 
                            f"requested {num_cycles}")
 
+    if planted_solution is not None:
+        warnings.warn("planted_solution is deprecated; use spin-reversal transforms: "
+                      "(1) bqm.flip_variable() for superficial reorientation of "
+                      "spin labels to match desired planting or "
+                      "(2) SpinReversalComposite for presentation of a randomized "
+                      "planted solution to some solver.",
+                      DeprecationWarning)
```
arcondello, line 164: need `stacklevel=2` (or maybe `stacklevel=3` because of the decorator) as well so the warning actually tells the user where to change things.

releasenotes/notes/planted-sat-f0200528096b73e3.yaml
```diff
@@ -0,0 +1,9 @@
+---
+features:
+  - |
+    Add ``plant_solution`` keyword argument to ``dimod.generators.random_nae3sat()``, ``dimod.generators.random_2in4sat()`` and ``dimod.generators.frustrated_loop`` functions.
+    Add 
+deprecations:
+  - |
+    Deprecated ``planted_solution`` keyword argument in ``dimod.generators.frustrated_loop``.
```
arcondello, line 8: ` ``dimod.generators.frustrated_loop()`` `

dimod/generators/fcl.py
```diff
@@ -144,6 +155,17 @@ def frustrated_loop(graph: GraphLike,
         raise RuntimeError(f"number of found cycles, {good_cycles}, is below " 
                            f"requested {num_cycles}")
 
+    if planted_solution is not None:
+        warnings.warn("planted_solution is deprecated; use spin-reversal transforms: "
```
arcondello, line 159: make sure you're catching all the warnings emitted in the unittests with `with self.assertWarns(DeprecationWarning):`, running the unittests should not emit warnings

arcondello (approved):
LGTM. Rerunning CI now because I think the errors were unrelated to this PR

PR #1201, opened 2022-06-06
Speed up `QM.from_file()`

arcondello:
I hate the duplicated code, but we need to do it this way to maintain binary compatibility, see https://github.com/dwavesystems/dimod/pull/1193.

Gives ~2x speed up on the benchmark problems. Will get a bigger speedup on larger problems.

### DIFF
--- a/dimod/quadratic/cyqm/cyqm_template.pyx.pxi
+++ b/dimod/quadratic/cyqm/cyqm_template.pyx.pxi
@@ -21,6 +21,7 @@ cimport cython
 
 from cython.operator cimport preincrement as inc, dereference as deref
 from libc.math cimport ceil, floor
+from libc.string cimport memcpy
 from libcpp.vector cimport vector
 
 import dimod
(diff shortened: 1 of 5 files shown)


dimod/quadratic/cyqm/cyqm_template.pyx.pxi
```diff
@@ -105,17 +106,24 @@ cdef class cyQM_template(cyQMBase):
...
+        if num_neighbors*itemsize > buff.size:
+            raise RuntimeError
 
         cdef Py_ssize_t i
+        cdef index_type ui
+        cdef bias_type bias
         for i in range(num_neighbors):
-            self.cppqm.add_quadratic(vi, index_view[i], bias_view[i])
+            memcpy(&ui, &buff[i*itemsize], index_itemsize)
+            memcpy(&bias, &buff[i*itemsize+index_itemsize], bias_itemsize)
```
randomir, line 124: Why is this copy needed? Could you get away just with a reference and recast?

randomir (approved):
LGTM!

dimod/quadratic/cyqm/cyqm_template.pyx.pxi
arcondello: I wasn't able to get Cython to let me do that. There may be some way but it seemed to get confused.

PR #1203, opened 2022-06-13
Refactor LP dumping to dimod.lp namespace

arcondello:
Because `CQM.from_lp_file()` was deprecated as part of https://github.com/dwavesystems/dimod/pull/1182 it makes sense to create LP files with `lp.dump()` and `lp.dumps()` rather than `CQM.to_lp_file()`.

This way we're not introducing a method only to immediately deprecate it.


seatim (approved):
LGTM!

I can confirm a nice performance improvement in serialization time with this change: `dimod.lp.dumps` runs about 40% faster than `CQM.to_lp_file` did on a benchmark I have (consisting of 100 cqm files with a median file size of ~1MB and max file size of ~22MB).

PR #1204, opened 2022-06-13
Support numpy>=1.17.3

arcondello:
This makes it easier for packages that depend on dimod to use `oldest-supported-numpy`.

The issue is when a package specifies say `dimod==0.11.0,oldest-supported-numpy` in their pyproject.toml, the pip install fails because dimod requires `numpy>=1.20.0` at runtime.
E.g. https://app.circleci.com/pipelines/github/dwavesystems/dwave-preprocessing/83/workflows/1bd72543-e11d-463d-a56e-5cdcad38f3c9/jobs/2368

So we loosen the range as much as possible. Unfortunately, the oldest supported numpy for Python 3.7 (1.14.5) is missing some pretty key stuff, like `default_rng`. So it's not really practical to support it. That means that [dwave-preprocessing](https://github.com/dwavesystems/dwave-preprocessing) will need to do something like
```
    'numpy==1.17.3;python_version<"3.8"',
    'oldest-supported-numpy;python_version>="3.8"',
```
in its `pyproject.toml`. At least until we can drop Python 3.7 on [June 27 2023](https://endoflife.date/python).

See also https://github.com/dwavesystems/dwave-ocean-sdk/issues/208

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -79,7 +79,9 @@ setup(
         'dimod/include/',
         ],
     install_requires=[
-        'numpy>=1.20.0,<2.0.0',
+        # this is the oldest supported NumPy on Python 3.8, we cannot
+        # support the oldest for 3.7, 1.14.5
+        'numpy>=1.17.3,<2.0.0',
         ],
     # we use the generic 'all' so that in the future we can add or remove
     # packages without breaking things
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -254,7 +254,7 @@ jobs:
     parameters:
       python-version:
         type: string
-      numpy-version:
+      dependencies:
         type: string
 
     docker:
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -79,7 +79,9 @@ def finalize_options(self):
         'dimod/include/',
         ],
     install_requires=[
-        'numpy>=1.20.0,<2.0.0',
+        # this is the oldest supported NumPy on Python 3.8, we cannot
+        # support the oldest for 3.7, 1.14.5
+        'numpy>=1.17.3,<2.0.0',
```
randomir, line 84: I also think we should lift the `2.0` upper bound, but we can discuss that and/or do it in a separate PR.

randomir (approved):
LGTM

setup.py
arcondello: Agree it's a bit pointless, we should probably either lift it altogether or tighten it down to 1.n+2.0. But yeah, followup PR.

Issue #1188, opened 2022-05-13
Add support for printing CQMs


arcondello:
It would be useful to have a string representation for CQM models.

**Proposed Solution**
The most obvious thing would be something like
```python
def __str__(self):
    return self.to_lp_file().read()
```
(see https://github.com/dwavesystems/dimod/pull/1052) or
```python
def __str__(self):
    return dimod.lp.dumps(self)
```
assuming that we add `dump` and `dumps` to go with https://github.com/dwavesystems/dimod/pull/1182

LP files are fairly standard and pretty clear to read. However, there are two issues
1) LP files do not handle SPIN variables, but we do. This is fairly easily handled by adding a `Spin` section.
2) LP files have a weird rule where in the objective, quadratic terms need to be bounded by `[ ... ]/2`. We could either keep that quirk, thereby keeping valid LP files, or ignore it for readability.

arcondello:
**Different Proposed Solution**

Something LP-like, but a bit less in the uncanny valley might be
```python
>>> cqm = dimod.CQM()
>>> a, b = dimod.Reals('ab')
>>> i, j = dimod.Integers('ij'
>>> cqm.set_objective(i + 2*j + i*j)
>>> cqm.add_constraint(a + b <= 5, label='hello')
>>> print(cqm)
Minimize
  Integer('i') + 2 * Integer('j') + Integer('i') * Integer('j')
Subject To
  'hello': Real('a') + Real('b') <= 5
```
At least in this case we dodge problem (3) from above by just doing `f'Real({v.variables[0]!r})'`

randomir:
Speaking of `repr`, I'd print only a few significant stats (num vars, constraints, types, etc), instead going for full-eval-ready repr we have on `BQM`.

arcondello:
I don't think we can do the "Variables" section in full generality because `Binary(0) + Binary('0')` is a perfectly valid model and it wouldn't be obvious how to cast them to intermediate labels.
So I think that brings us to
```
Objective
    Binary('i') + 2*Real('j') +  Binary('i')*Real('j')
Constraints:
  'hello': Integer('a') + Integer('b') <= 5
Bounds
  -5 <= Integer('a') <= 5
```
it's a bit verbose but is at least unambiguous.

PR #1207, opened 2022-06-16
Add ConstrainedQuadraticModel.__str__() method

seatim:
For #1188.

### DIFF
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -1949,7 +1949,7 @@ class ConstrainedQuadraticModel:
 
         return obj
 
-    _STR_MAX_DISPLAY_ITEMS = 4
+    _STR_MAX_DISPLAY_ITEMS = 10
 
     def __str__(self):
         vartype_name = {Vartype.SPIN: 'Spin',
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -1970,7 +1970,7 @@ class ConstrainedQuadraticModel:
         sio.write('\n')
         sio.write('Bounds\n')
         for v in self.variables:
-            if self.vartype(v) is not Vartype.BINARY:
+            if self.vartype(v) not in (Vartype.BINARY, Vartype.SPIN):
                 sio.write(f'  {self.lower_bound(v)} <= {var_encoder(v)} <= {self.upper_bound(v)}\n')
 
         return sio.getvalue()
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -1979,7 +1979,7 @@ class ConstrainedQuadraticModel:
         sio.write('\n')
         sio.write('Bounds\n')
         for v in self.variables:
-            if self.vartype(v) not in (Vartype.BINARY, Vartype.SPIN):
+            if self.vartype(v) in (Vartype.INTEGER, Vartype.REAL):
                 sio.write(f'  {self.lower_bound(v)} <= {var_encoder(v)} <= {self.upper_bound(v)}\n')
 
         return sio.getvalue()
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -1960,12 +1960,16 @@ class ConstrainedQuadraticModel:
 
         sio = StringIO()
         sio.write('Objective\n')
-        sio.write(f'  {self.objective.to_polystring(encoder=var_encoder)}\n')
+        sio.write('  ')
+        sio.write(self.objective.to_polystring(encoder=var_encoder))
+        sio.write('\n')
 
         sio.write('\n')
         sio.write('Constraints\n')
         for label, c in self.constraints.items():
-            sio.write(f'  {label}: {c.to_polystring(encoder=var_encoder)}\n')
+            sio.write(f'  {label}: ')
+            sio.write(c.to_polystring(encoder=var_encoder))
+            sio.write('\n')
 
         sio.write('\n')
         sio.write('Bounds\n')
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -1949,6 +1949,8 @@ class ConstrainedQuadraticModel:
 
         return obj
 
+    _STR_MAX_DISPLAY_ITEMS = 4
+
     def __str__(self):
         vartype_name = {Vartype.SPIN: 'Spin',
                         Vartype.BINARY: 'Binary',
@@ -1968,7 +1970,7 @@ class ConstrainedQuadraticModel:
                 assert x is not None
 
                 if last is not None:
-                    if k < 4:
+                    if k < self._STR_MAX_DISPLAY_ITEMS:
                         render_element(last)
                     elif not limited:
                         sio.write('  ...\n')
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -28,6 +28,7 @@ import uuid
 import warnings
 import zipfile
 
+from io import StringIO
 from numbers import Number
 from typing import Hashable, Optional, Union, BinaryIO, ByteString, Iterable, Collection, Dict
 from typing import Callable, MutableMapping, Iterator, Tuple, Mapping, Any, NamedTuple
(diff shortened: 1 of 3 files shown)


dimod/constrained.py
```diff
@@ -1948,6 +1949,32 @@ def from_lp_file(cls,
...
+
+        sio.write('\n')
+        sio.write('Constraints\n')
+        for label, c in self.constraints.items():
+            sio.write(f'  {label}: {c.to_polystring(encoder=var_encoder)}\n')
+
+        sio.write('\n')
+        sio.write('Bounds\n')
+        for v in self.variables:
+            if self.vartype(v) is not Vartype.BINARY:
```
arcondello, line 1973: or `Vartype.SPIN`

dimod/constrained.py
```diff
@@ -1948,6 +1949,32 @@ def from_lp_file(cls,
...
+                        Vartype.BINARY: 'Binary',
+                        Vartype.INTEGER: 'Integer',
+                        Vartype.REAL: 'Real'}
+
+        def var_encoder(v):
+            return f'{vartype_name[self.vartype(v)]}({v!r})'
+
+        sio = StringIO()
+        sio.write('Objective\n')
+        sio.write(f'  {self.objective.to_polystring(encoder=var_encoder)}\n')
```
arcondello, line 1963: not that performance really matters for printing, but this way you're doing a big string copy. I imagine that 
```python
sio.write('  ')
sio.write(self.objective.to_polystring(encoder=var_encoder))
sio.write('\n')
```
will end up being faster for large models.

dimod/constrained.py
seatim: Fixed

dimod/constrained.py
seatim: ✔️

arcondello (approved):
I am happy with this as-is. But a few more things we might consider doing as part of this PR or a future one:
1) Maybe a summary section at the top? Something like
```
Shape:
  num variables: 100
  num constraints: 106
```
though not sure if "shape" is a good word there.

2) Large model handling. We could do something like
```
Constraints
    c0: Binary('m')*Integer('j') - Binary('m')*Binary('n') <= 0
    c1: Binary('m')*Integer('j') - Binary('m')*Binary('n') <= 0
    c2: Binary('m')*Integer('j') - Binary('m')*Binary('n') <= 0
    ...
    c1000: Binary('m')*Integer('j') - Binary('m')*Binary('n') <= 0
```

3) No "constraints" section when there are no constraints.

releasenotes/notes/cqm-__str__-method-b9f37742662c0810.yaml
```diff
@@ -0,0 +1,3 @@
+---
+features:
+  - Add ``ConstrainedQuadraticModel.__str__()`` method.
```
arcondello, line 3: It's nice to add issue references when we have them.

```suggestion
  - Add ``ConstrainedQuadraticModel.__str__()`` method.  See  #1188 <https://github.com/dwavesystems/dimod/issues/1188>`_.
```

JoelPasvolsky (approved):
LGTM.
I guess my only thoughts are
1. The printing would be nicer if it didn't include vars with zero weightings (versus the constraint representation that has all linear weightings):

```
>>> cqm.constraints["c1"]
Le(QuadraticModel({'a': 0.0, 'b': 1.0}, {}, 0.0, {'a': 'INTEGER', 'b': 'INTEGER'}, dtype='float64'), 2)
>>> print(cqm)
...
Constraints
  c1: 0*Integer('a') + Integer('b') <= 2
```

2. Be nice if one could set a global parameter that specified rounding of points after the decimal. 

```
 Objective
  8*Real('rice') + 22*Real('tofu') ... + 4.800000000000001*Real('lentils') + ...
```

I understand if neither properly belong to this PR and are ignored. Just making a wish here.

arcondello:
For (1), we use the same rules as `to_polystring()`, see discussion(s) in https://github.com/dwavesystems/dimod/pull/1054. IMO it is important to show them somewhere, since they are present in the object. We could adjust the rules to only show the multiplications by 0 if they are not in any quadratic interactions _and not in any constraints_`.

releasenotes/notes/cqm-__str__-method-b9f37742662c0810.yaml
seatim: ✔️ Did that

arcondello:
For (1) I think I like the single line. 

For (2) I think the 80char limit would be ideal. But agree it would take a lot of fiddling to always break between the variables. I am fine with it being out of scope for now. Really the change should be upstream in `to_polystring()` anyway.

For (3) I take your point about explicitness. I don't mind the empty constraint section

dimod/constrained.py
```diff
@@ -1959,6 +1959,35 @@ def var_encoder(v):
             return f'{vartype_name[self.vartype(v)]}({v!r})'
 
         sio = StringIO()
+
+        def render_limited_number(iterable, render_element):
```
seatim, line 1963: @arcondello What do you think of this function?

I'm not 100% happy with it.  I feel there is a better way, some `itertools` or other simpler solution.  But such hasn't occurred to me yet.

seatim:
Summary of the state of this PR from my perspective:

I'm on the whole satisified with this as it stands, as it adds a nice capability and I'm not aware of any blocking issues.  And I'm happy to improve it if there is consensus on what is needed.

To summarize the possible future work we've discussed here:

1. Add ability to only show the multiplications by 0 if they are not in any quadratic interactions _and not in any constraints_. 
2. Add ability to limit line length to given number of characters, by omitting sequence of terms and substituting ellipses in their place.
3. Add ability to control numerical precision of string representation of biases.
4. Add ability for user to control all of the above via some combination of global functions (e.g. surfaced in a new `display_options` module), environment variables, or config file variables.

dimod/constrained.py
arcondello: Nothing jumps out at me either

dimod/constrained.py
```diff
@@ -1959,6 +1959,35 @@ def var_encoder(v):
...
+            limited = False
+            last = None
+
+            for k, x in enumerate(iterable):
+                assert x is not None
+
+                if last is not None:
+                    if k < 4:
```
arcondello, line 1971: IMO this number could be higher. More like 10 or even 20.

seatim: 👍 
Maybe the number of head and tail items should be symmetric also?  That would be consistent with both `numpy` and `pandas` (and shouldn't be too hard to implement):

```
>>> import pandas as pd
>>> pd.DataFrame({str(x): [y*x for y in range(100)] for x in range(100)})
    0   1    2    3    4    5  ...    94    95    96    97    98    99
0   0   0    0    0    0    0  ...     0     0     0     0     0     0
1   0   1    2    3    4    5  ...    94    95    96    97    98    99
2   0   2    4    6    8   10  ...   188   190   192   194   196   198
3   0   3    6    9   12   15  ...   282   285   288   291   294   297
4   0   4    8   12   16   20  ...   376   380   384   388   392   396
.. ..  ..  ...  ...  ...  ...  ...   ...   ...   ...   ...   ...   ...
95  0  95  190  285  380  475  ...  8930  9025  9120  9215  9310  9405
96  0  96  192  288  384  480  ...  9024  9120  9216  9312  9408  9504
97  0  97  194  291  388  485  ...  9118  9215  9312  9409  9506  9603
98  0  98  196  294  392  490  ...  9212  9310  9408  9506  9604  9702
99  0  99  198  297  396  495  ...  9306  9405  9504  9603  9702  9801

[100 rows x 100 columns]
>>> import numpy as np
>>> np.array([[y*x for y in range(100)] for x in range(100)])
array([[   0,    0,    0, ...,    0,    0,    0],
       [   0,    1,    2, ...,   97,   98,   99],
       [   0,    2,    4, ...,  194,  196,  198],
       ...,
       [   0,   97,  194, ..., 9409, 9506, 9603],
       [   0,   98,  196, ..., 9506, 9604, 9702],
       [   0,   99,  198, ..., 9603, 9702, 9801]])
```

BTW, FWIW, I did spend a few minutes looking for the functions that in both `numpy` and `pandas` that produce those outputs, so that we might reuse them.  Came up empty in both cases... I am guessing it's done in C code, so we can't reuse.

arcondello: I believe it's [here](https://github.com/numpy/numpy/blob/main/numpy/core/arrayprint.py). We used some of it for [sample set printing](https://github.com/dwavesystems/dimod/blob/main/dimod/serialization/format.py).

But yes, I like the centered format.

seatim: OK, I didn't reuse any of that, but did get the centered format and introduced a private class attribute to control the number of display items, with default value = 10.

arcondello (approved):
LGTM!

PR #1208, opened 2022-06-16
Output bounds for REAL variables in LP files

seatim:


arcondello (approved):

dimod/lp.py
```diff
@@ -221,7 +221,7 @@ def dump(cqm: dimod.ConstrainedQuadraticModel, file_like: typing.TextIO):
     f.write('Bounds\n')
 
     for v, vartype in vartypes.items():
-        if vartype is Vartype.INTEGER:
+        if vartype is not Vartype.BINARY:
```
arcondello, line 224: I wonder if in the spirit of future-proofing, we should do something like
```python
if vartype is in (Vartype.INTEGER, Vartype.REAL):
   # do the thing
elif vartype is not Vartype.BINARY:
    raise RuntimeError("unexpected vartype")
```
sure it means we'll get errors when we add a new vartype, but I would almost prefer to get the error rather than potentially unexpected behavior.

seatim: I agree, will do.

PR #1222, opened 2022-07-11
Speed up `SampleSet.from_samples_cqm()`

arcondello:


dimod/sampleset.py
```diff
@@ -850,21 +863,41 @@ def from_samples_cqm(cls, samples_like, cqm, rtol=1e-6, atol=1e-8, **kwargs):
...
 
-        is_satisfied = [[datum.violation <= atol + rtol*abs(datum.rhs_energy)
-                         for datum in cqm.iter_constraint_data((s, labels))] for s in samples]
-        is_feasible = [all(satisfied) for satisfied in is_satisfied]
+        constraint_labels = []
+        is_satisfied = np.empty((samples.shape[0], len(cqm.constraints)), dtype=bool)
+        for i, (label, comparison) in enumerate(cqm.constraints.items()):
+            constraint_labels.append(label)
+
+            lhs = comparison.lhs.energies(samples_like)
```
wbernoudy, line 886: Isn't this roughly what `iter_constraint_data()` was doing? I guess it's that the `lhs.energies()` function only needs to figure out the labels once for all the samples, while you had to pass in one sample at a time to `iter_constraint_data()`?

wbernoudy (approved):

dimod/sampleset.py
arcondello: Yeah basically it saves a bunch of overhead to just do it in bulk rather than iteratively

Issue #1221, opened 2022-07-11
Mypy stub files for Cython files


arcondello:
It would be good to add [stub files](https://mypy.readthedocs.io/en/stable/stubs.html#stub-files) for some of the most commonly used cython packages because cython current does not support type annotations.

Especially `dimod.cyvariables` and `dimod.cyutilities` would benefit because the `Variables` object is used in many places.

Should be as simple as adding a `dimod/cyvariables.pyi` file etc.

arcondello:
As a downside, it does not seem like those type annotations are picked up by sphinx (see https://github.com/sphinx-doc/sphinx/issues/7630).

PR #1223, opened 2022-07-12
Add typing support to the Variables class

arcondello:
See https://github.com/dwavesystems/dimod/issues/1221


dimod/variables.py
```diff
@@ -70,27 +72,50 @@ def iter_deserialize_variables(variables):
...
 
+    Examples:
+
+        The variables object can be used to encode an ordered set of variables.
+
+        >>> variables = dimod.variables.Variables(['a', 'b', 0, 1])
+        >>> print(variables)
+        Variables(['a', 'b', 0, 1])
+
+        The variables object can be interacted with like a
```
JoelPasvolsky, line 91: Perhaps something like, 
"The variables object supports a similar interface to the Python :class:`~collections.abc.Sequence` class."

dimod/variables.py
```diff
@@ -70,27 +72,50 @@ def iter_deserialize_variables(variables):
...
+        :class:`~collections.abc.Sequence`.
+
+        >>> variables[0]
+        'a'
+        >>> variables[1::2]
+        Variables(['b', 1])
+        >>> len(variables)
+        4
+
+        And like a :class:`~collections.abc.Set`.
```
JoelPasvolsky, line 101: Can we remove this half sentence and just add this last example to the previous three or fill in the sentence with a description of the fourth example?

dimod/cyvariables.pyx
```diff
@@ -171,11 +182,20 @@ cdef class cyVariables:
...
         """Remove the last variable.
 
-        This method is semi-public. it is intended to be used by
-        classes that have :class:`.Variables` as an attribute, not by the
-        the user.
+        Returns:
+            :class:`~dimod.typing.Variable`: The removed variable.
+
+        Raises:
+            IndexError: If the container is empty.
```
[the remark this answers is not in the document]
arcondello: Sphinx does this one automatically. But no harm in being explicit I suppose

PR #1224, opened 2022-07-13
Add CQM.spin_to_binary() method

arcondello:


jackraymond (approved):
Looks good, tests are sufficient.

Why not add binary_to_spin as well? 
Might be a large family of related transformation, some of them 1 directional for practical purposes. Might want to think about broader structure at some point.
binary  -> integer + constraint bounds (for testing primarily)
binary or integer -> real + range bounds.  (for linear relaxations)

arcondello:
> Why not add binary_to_spin as well?

Definitely open to it. Though I think given that no MIQP solver (other than us) that I know of supports SPIN variables, this seems like the most useful. Happy to add more in the future.

> Might be a large family of related transformation, some of them 1 directional for practical purposes. Might want to think about broader structure at some point. binary -> integer + constraint bounds (for testing primarily) binary or integer -> real + range bounds. (for linear relaxations)

Agree. There is [a bit](https://github.com/dwavesystems/dimod/blob/eecde171848f9690e3199ff2565d8336f3dcac7c/dimod/include/dimod/quadratic_model.h#L1237) of architecture for this in the C++ code. That basically kicks the can down the road. Should get around to that at some point though.

Issue #1198, opened 2022-06-02
`CQM.is_linear()` or similar


arcondello:
As simple as
```python
def is_linear(self) -> bool:
    return self.objective.is_linear() and all(comp.lhs.is_linear() for comp in self.constraints.values())
```

PR #1225, opened 2022-07-14
Add CQM.add_linear() method

arcondello:
Closes https://github.com/dwavesystems/dimod/issues/1198

### DIFF
--- a/releasenotes/notes/CQM.is_linear-25156a452fc4e012.yaml
+++ b/releasenotes/notes/CQM.is_linear-25156a452fc4e012.yaml
@@ -1,0 +1,3 @@
+---
+features:
+  - Add ``ConstrainedQuadraticModel.is_linear()`` method.
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -1321,6 +1321,11 @@ class ConstrainedQuadraticModel:
                 and all(constraint_eq(constraint, other.constraints[label])
                         for label, constraint in self.constraints.items()))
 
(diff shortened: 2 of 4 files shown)


releasenotes/notes/CQM.is_linear-25156a452fc4e012.yaml
```diff
@@ -0,0 +1,3 @@
+---
+features:
+  - Add ``ConstrainedQuadraticModel.is_linear()`` method.
```
thisac, line 3: Is this generated automatically by Reno?

thisac (approved):
Looks good to me!

releasenotes/notes/CQM.is_linear-25156a452fc4e012.yaml
arcondello: No, the file is hand written by me. Reno then turns the files in `releasenotes/` into a changelog. See [here](https://docs.ocean.dwavesys.com/en/stable/docs_dimod/release_notes.html) and [here](https://github.com/dwavesystems/dimod/releases). The former is generated as part of the sphinx build (see [here](https://github.com/dwavesystems/dimod/blob/main/docs/release_notes.rst)), the latter I just generate locally and then copy in.

arcondello: I should clarify that the file is generated by reno, but the contents are hand written.

PR #1226, opened 2022-07-18
Document rhs comparison requirements

arcondello:
The right-hand-side of a symbolic comparison must be numeric. This PR spells that out in a few places.

### DIFF
--- a/dimod/constrained.py
+++ b/dimod/constrained.py
@@ -213,21 +213,14 @@ class ConstrainedQuadraticModel:
             raise TypeError("model should be a QuadraticModel or a BinaryQuadraticModel")
 
     def add_constraint(self, data, *args, **kwargs) -> Hashable:
-        """A convenience wrapper for other methods that add constraints.
+        """Add a constraint to the model.
 
-        Examples:
-            >>> from dimod import ConstrainedQuadraticModel, Integers
-            >>> i, j = Integers(['i', 'j'])
-            >>> cqm = ConstrainedQuadraticModel()
-            >>> cqm.add_constraint(i + j <= 3, label='Constrained i-j range')
-            'Constrained i-j range'
-
-        See also:
-            :meth:`~.ConstrainedQuadraticModel.add_constraint_from_model`
-
-            :meth:`~.ConstrainedQuadraticModel.add_constraint_from_comparison`
-
-            :meth:`~.ConstrainedQuadraticModel.add_constraint_from_iterable`
+        This method dispatches to one of several specific methods based on
+        the type of the first argument.
+        For a detailed description of the accepted argument types, see
+        :meth:`~.ConstrainedQuadraticModel.add_constraint_from_model`,
+        :meth:`~.ConstrainedQuadraticModel.add_constraint_from_comparison`,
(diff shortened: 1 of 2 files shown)


TanjidIslam (approved):
Looks good to me

dimod/constrained.py
```diff
@@ -335,23 +333,62 @@ def add_constraint_from_comparison(self,
...
+
+                x + y \le x y
+
+
+            First create the relevant variables and the model.
+
+            >>> x, y = dimod.Binaries(['x', 'y'])
+            >>> cqm = dimod.ConstrainedQuadraticModel()
+
+            Trying to directly compare the left- and right-hand-side
```
JoelPasvolsky, line 371:
```suggestion
            Trying to directly compare the left-hand and right-hand sides
```

dimod/constrained.py
```diff
@@ -335,23 +333,62 @@ def add_constraint_from_comparison(self,
...
+            of the equation will raise an error, because the right-hand-side
+            is not an integer or float.
+
+            >>> try:
+            ...     cqm.add_constraint(x + y <= x * y)
+            ... except TypeError:
+            ...     print("Not allowed!")
+            Not allowed!
```
JoelPasvolsky, line 379: Seems a bit of an overkill -- why not just state that you cannot have the model ``x*y`` in the RHS and then say that to overcome this, do the subtraction?

dimod/constrained.py
```diff
@@ -493,7 +530,7 @@ def add_discrete_from_comparison(self,
         Args:
             comp: Comparison object. The comparison must be a linear
                 equality constraint with all of the linear biases on the
-                left-hand side equal to one and the right hand side equal
+                left-hand-side equal to one and the right-hand-side equal
```
JoelPasvolsky, line 533:
```suggestion
                left-hand side equal to one and the right-hand side equal
```

JoelPasvolsky (commented):
1. In several places: "right-hand-side" should be "right-hand side" (single hyphen for the modifier only); lhs too 
2. Nice to have consistency with end-of-sentence periods in the examples

dimod/constrained.py
arcondello: I see, I had copied your double hyphen from https://github.com/dwavesystems/dimod/blame/main/docs/intro/intro_symbolic_math.rst#L163, but I can change them all to single.

JoelPasvolsky: > I see, I had copied your double hyphen from https://github.com/dwavesystems/dimod/blame/main/docs/intro/intro_symbolic_math.rst#L163, but I can change them all to single.

Yes, the "left-hand side(``lhs``)" is correct but the ", right-hand-side" has an extra hyphen there

dimod/constrained.py
arcondello: Honestly so the doctest picks it up. A little bit of future proofing honestly.

JoelPasvolsky: Okay, FYI there are options to hide doctest tests in the built HTML, so you can use that for future proofing

PR #1227, opened 2022-07-18
Add methods to clear QMs/BQMs

arcondello:
Add the ability to clear the contents of BQMs and CQMs.

Would also be useful for https://github.com/dwavesystems/dimod/issues/928


dimod/include/dimod/quadratic_model.h
```diff
@@ -1196,6 +1203,11 @@ class QuadraticModel : public QuadraticModelBase<Bias, Index> {
         return v;
     }
 
+    void clear() {
+        this->varinfo_.clear();
+        base_type::clear();
+    }
```
thisac, line 1209: So, `base_type::clear()` calls back to the `QuadraticModelBase` clear method above, right? I assume nothing else needs to be cleared for `BinaryQuadraticModel`, e.g., `vartype`, meaning that only the base class method is called there?

dimod/quadratic/quadratic_model.py
```diff
@@ -672,6 +672,10 @@ def change_vartype(self, vartype: VartypeLike, v: Variable) -> "QuadraticModel":
         self.data.change_vartype(vartype, v)
         return self
 
+    def clear(self) -> None:
+        """Remove the offset and all variables and interactions from the model."""
+        self.data.clear()
```
thisac, line 677: ...and this line calls clear in `cyqm_template.pyx.pxi` (via the corresponding data class), right?

dimod/quadratic/cyqm/cyqm_template.pyx.pxi
```diff
@@ -359,6 +359,10 @@ cdef class cyQM_template(cyQMBase):
             raise TypeError(f"cannot change vartype {self.vartype(v).name!r} "
                             f"to {vartype.name!r}") from None
 
+    def clear(self):
+        self.cppqm.clear()
+        self.variables._clear()
```
thisac, line 364: `self.cppqm.clear()` calls back to `QuadraticModel::clear()` in `quadratic_model.hpp`, while `self.variables._clear()` calls `cyvariables.pyx`, right? Just trying to understand how everything fits together.

tests/test_bqm.py
```diff
@@ -329,6 +329,17 @@ def test_change_vartype_spin_to_binary_copy(self, name, BQM):
         self.assertConsistentEnergies(spin=bqm, binary=new)
 
 
+class TestClear(unittest.TestCase):
```
thisac, line 332: Both tests look good, as far as I can tell. I guess `BQMs` goes through all valid types of BQMs, and there's no need to add tests for the base quadratic model class (should be covered by BQM and QM).

thisac (approved):
As far as I can tell this should work as intended. 🙂

dimod/include/dimod/quadratic_model.h
arcondello: Exactly.

dimod/quadratic/cyqm/cyqm_template.pyx.pxi
arcondello: Yes. For the models, Python calls Cython calls C++ :smile: 
For Variables you're correct as well.

tests/test_bqm.py
arcondello: Yup!

PR #1230, opened 2022-07-22
Fix as_samples() when given a sampleset and a labels_type

arcondello:
Adds missing case from 156d9787978a620d010dd0ddb7bdf168482d1cef. See https://github.com/dwavesystems/dimod/pull/1222

### DIFF
--- a/dimod/sampleset.py
+++ b/dimod/sampleset.py
@@ -1915,9 +1915,10 @@ def _as_samples_sampleset(samples_like: SampleSet,
                           dtype: Optional[DTypeLike] = None,
                           copy: bool = False,
                           order: ArrayOrder = 'C',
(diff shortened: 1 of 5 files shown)


wbernoudy:
Should there be a unit test for `dimod.QuadraticModel.energies(sampleset)`? (This is what triggered this bug for me initially)

arcondello:
You know, why not. We should probably have something like https://github.com/dwavesystems/dimod/blob/44f23f410138bdc4483512572de01c7d4334ba19/dimod/testing/sampler.py#L76 but for samples-like. And can then use that everywhere.

wbernoudy (approved):

arcondello:
Didn't end up adding a new testing function, but did add some more test cases.

PR #1233, opened 2022-08-12
Adding weighted constraints info in CQM

alexzucca90:
It is possible to add information about weighted constraints in the `dimod.ConstrainedQuadraticModel` class. Other than encoding the functional information, there are no functional changes in how objectives and feasibility are calculated.


dimod/constrained.py
```diff
@@ -18,6 +18,7 @@
 
 from __future__ import annotations
 
+import collections
```
arcondello, line 21: I don't think you're using this any more

dimod/constrained.py
```diff
@@ -63,6 +65,14 @@ class ConstraintData(NamedTuple):
     violation: float
 
 
+class SoftConstraint(NamedTuple):
+    weight: float
+    penalty: str
+
+
+# SoftConstraint = collections.namedtuple('SoftConstraint', ['weight', 'norm'])
```
arcondello, line 73: Don't need this any more

dimod/constrained.py
```diff
@@ -157,11 +167,12 @@ class ConstrainedQuadraticModel:
         'capacity_bin_3'
 
     """
+
     def __init__(self):
         # discrete variable tracking, we probably can do this with less memory
         # but for now let's keep it simple
-        self.discrete: Set[Hashable] = set()  # collection of discrete constraints
-
+        self.discrete: Set[Hashable] = set()  # collection of discrete constraints\
```
arcondello, line 174: not sure why you added the trailing `\`

dimod/constrained.py
```diff
@@ -255,6 +269,13 @@ def add_constraint_from_model(self,
                 to improve performance, but subsequently mutating ``qm`` can
                 cause issues.
 
+            weight: weight for soft-constraint. If None, the constraints
+                is interpreted as hard.
+
+            penalty: Penalty type for soft-constraint. Must be one of
+                'linear', 'quadratic'. Ignored is weight is None. 'quadratic'
```
arcondello, line 276: "Ignored **if** the weight..."

dimod/constrained.py
```diff
@@ -307,12 +328,32 @@ def add_constraint_from_model(self,
         else:
             raise RuntimeError("unexpected sense")
 
+        if weight is not None:
+
+            if weight < 0:
+                raise ValueError("weight for soft-constraint should be >= 0")
+
+            if penalty not in ['linear', 'quadratic']:
```
arcondello, line 336:
```python
if penalty == 'quadratic':
    ...
elif penalty != 'linear':
    raise ValueError(...
```

releasenotes/notes/cqm-add-weighted-constraints-0642ec7f3f094b6f.yaml
```diff
@@ -0,0 +1,7 @@
+features:
+  - |
+    Added ``weight`` and `penalty` parameters on ``ConstrainedQuadraticModel.add_constraint`` methods
+    to encoded weighted constraints.
+  - |
+      Add `weight` and `penalty` files the file created by
```
arcondello, line 6: Indentation. Also, in this case I would just say
```
Add constrained quadratic model serialization version 1.3 to support soft constraints.
```

dimod/constrained.py
```diff
@@ -35,6 +36,7 @@
 
 import numpy as np
 
+import dimod
```
arcondello, line 39: I don't think you need this, you use it for `dimod.QuadraticModel`, but that's already imported lower down

arcondello:
Reminder for myself, we should also add `num_weighted_constraints` to the header.

arcondello:
Reminder for myself, we need to update `SampleSet.from_samples_cqm()` to handle soft constraints.

arcondello:
Reminder for myself (found by @alexzucca90 ), we should update the [CQM equations](https://github.com/dwavesystems/dimod/blob/5d2043d0b662f38bd51e062367c2e4b703b457a1/dimod/constrained.py#L73) to include soft constraints. Possibly as a second set of equations under the assumption that not all users will use that feature.

arcondello:
Reminder for myself (found by @dbusto), support SPIN variables with quadratic penalties.

alexzucca90 (commented):
Checked the latest changes, they look good to me.

PR #1243, opened 2022-09-07
compactified CQM equations

alexzucca90:
NFC

The two separate equality and inequality constraints in the `ConstrainedQuadraticModel` docstring are replaced with single more compact notation that also reflects the structure of the class.

### DIFF
--- a/releasenotes/notes/compactify-cqm-equations-5456497655b5e24d.yaml
+++ b/releasenotes/notes/compactify-cqm-equations-5456497655b5e24d.yaml
@@ -1,3 +1,0 @@
----
-features:
-  - Replaced separate equality and inequality constraints in CQM docstring with single compact notation.
--- a/releasenotes/notes/compactify-cqm-equations-5456497655b5e24d.yaml
+++ b/releasenotes/notes/compactify-cqm-equations-5456497655b5e24d.yaml
@@ -1,0 +1,3 @@
+---
+features:
+  - Replaced separate equality and inequality constraints in CQM docstring with single compact notation.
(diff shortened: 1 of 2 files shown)


releasenotes/notes/compactify-cqm-equations-5456497655b5e24d.yaml
```diff
@@ -0,0 +1,3 @@
+---
+features:
```
arcondello, line 2: I don't think this needs a release note.

JoelPasvolsky (commented):
my one comment on the the implementation is whether `\circ` is the optimal symbol (is it standard? I'm not familiar with its use): it's small, so it looked at first glance a bit strange. 

I'm not sure I would find this easier to parse, it seems a trade between space (two equations) and time (one equation and then unpacking an unfamiliar symbol), but it works okay for me. 

We would want to also make [this](https://docs.ocean.dwavesys.com/en/stable/concepts/cqm.html) consistent.

alexzucca90:
@JoelPasvolsky For the `\circ` symbol, that is used in some publications, e.g. equation 2.1 in [this publication](https://www.researchgate.net/publication/337127039_Presolve_Reductions_in_Mixed_Integer_Programming). For the Ocean docs, I can make an accompanying PR, maybe after merging this?

JoelPasvolsky:
> @JoelPasvolsky For the `\circ` symbol, that is used in some publications, e.g. equation 2.1 in [this publication](https://www.researchgate.net/publication/337127039_Presolve_Reductions_in_Mixed_Integer_Programming). For the Ocean docs, I can make an accompanying PR, maybe after merging this?

Thanks, yes, I meant that the SDK update would be a different PR, it was just so we would keep it in mind.

JoelPasvolsky:
@alexzucca90, I'm doing a little A/B testing and will get back to you soon.

arcondello:
It might be slightly clearer to use `\bigcirc`, but have not tested it myself to be sure.

JoelPasvolsky (approved):

arcondello (approved):

Issue #1240, opened 2022-08-27
Please don't hard-code catch paths


yurivict:
**Description**
Instead of:
```
#include "../Catch2/single_include/catch2/catch.hpp"
```
it should be
```
#include <catch2/catch.hpp>
```
Then makefile should take care of the include path.
If the git submodule is present - it should be used. If not - ``` `pkg-config --cflags catch2` ``` should be used.

And it doesn't make sense to run git from makefiles because in many cases this directory comes from a tarball and it isn't a git repository.

arcondello:
> Then makefile should take care of the include path.

This makes sense to me.

> And it doesn't make sense to run git from makefiles because in many cases this directory comes from a tarball and it isn't a git repository.

In this case - because we include catch2 as a git submodule with a specific commit - I think it makes sense to keep it here. Worth remembering that although this package has some C++ and may some day get more direct C++ use, for now 99% of its users interface with it as a Python package. Keeping the instructions for running the C++ tests simple and in a MakeFile is useful. At some point in the future we could consider a more sophisticated CMake infrastructure.

Issue #1239, opened 2022-08-27
Please use $(CXX) instead of g++ in makefiles


yurivict:
**Description**
[Here](https://github.com/dwavesystems/dimod/blob/main/testscpp/Makefile#L20) and everywhere.

```$(CXX)``` is a standard variable for a C++ compiler, and ```g++``` is one of compilers.
On BSDs C++ compiler executable has the name ```c++```..

arcondello:
Good idea

Issue #1238, opened 2022-08-27
clang fails to compile C++ tests


yurivict:
**Description**
```
In file included from tests/test_quadratic_model.cpp:18:
In file included from ..//dimod/include/dimod/quadratic_model.h:24:
..//dimod/include/dimod/iterators.h:82:5: error: 'const' qualifier on reference type 'dimod::ConstQuadraticIterator::reference' (aka 'dimod::ConstQuadraticIterator::value_type &') has no effect [-Werror,-Wignored-reference-qualifiers]
    const reference operator*() const { return *term_ptr_; }
    ^~~~~~
In file included from tests/test_quadratic_model.cpp:18:
..//dimod/include/dimod/quadratic_model.h:1259:25: error: 'constexpr' non-static member function will not be implicitly 'const' in C++14; add 'const' to avoid a change in behavior [-Werror,-Wconstexpr-not-const]
    constexpr bias_type max_integer() { return vartype_limits<bias_type, Vartype::INTEGER>::max(); }
                        ^
                                      const
2 errors generated.
```

**Environment**
 - clang-14
 - OS: FreeBSD 13.1
 - Python version: 3.9

arcondello:
Interesting. 

While these are just warnings that we're treating as errors (via `-Werror`), it probably makes sense for us to test compiling with both clang and gcc in our CI. Certainly no harm from doing so.

PR #1245, opened 2022-09-14
Update C++ testing framework

arcondello:
See https://github.com/dwavesystems/dimod/issues/1238, https://github.com/dwavesystems/dimod/issues/1239, https://github.com/dwavesystems/dimod/issues/1240

Ideally we would test with the clang/gcc versions used by manylinux. But this is a lot simpler and gives us decent coverage.


randomir (approved):
LGTM!

PR #1246, opened 2022-09-14
Add Variables._remove() method

arcondello:
This is used in https://github.com/dwavesystems/dimod/pull/1242 but pulling it out to make that code slightly easier to review.


randomir (approved):
LGTM!

PR #1249, opened 2022-09-20
C++ Abstract Base Class

arcondello:
This PR pulls the QM/BQM changes from https://github.com/dwavesystems/dimod/pull/1242. The CQM changes are not included.


dimod/include/dimod/binary_quadratic_model.h
```diff
@@ -0,0 +1,195 @@
...
+     * data. `length` must be the number of (row, column, bias) entries.
+     */
+    template <class ItRow, class ItCol, class ItBias>
+    void add_quadratic(ItRow row_iterator, ItCol col_iterator, ItBias bias_iterator,
+                       index_type length);
+
+    /// Add one (disconnected) variable to the BQM and return its index.
+    index_type add_variable();
+
+    /// Change the vartype of the binary quadratic model
```
amahmudwave, line 86: Some comments end with full stops some don't like the above one.

dimod/include/dimod/binary_quadratic_model.h
```diff
@@ -0,0 +1,195 @@
...
+    void change_vartype(Vartype vartype);
+
+    bias_type lower_bound() const;
```
amahmudwave, line 89: Same comment as the one below.

dimod/include/dimod/vartypes.h
```diff
@@ -0,0 +1,130 @@
...
+    }
+    static constexpr Bias min() noexcept { return -max(); }
+};
+template <class Bias>
+class vartype_limits<Bias, Vartype::REAL> {
+ public:
+    static constexpr Bias default_max() noexcept { return max(); }
+    static constexpr Bias default_min() noexcept { return 0; }
+    static constexpr Bias max() noexcept { return 1e30; }
+    static constexpr Bias min() noexcept { return -1e30; }
```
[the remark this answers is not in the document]
arcondello: I see now harm being explicit

testscpp/tests/test_quadratic_model.cpp
```diff
@@ -19,567 +19,58 @@
...
+        WHEN("we use remove_variable()") {
+            qm.remove_variable(2);
+
+            THEN("the variable is removed and the model is reindexed") {
+                REQUIRE(qm.num_variables() == 4);
+                REQUIRE(qm.num_interactions() == 0);
+                CHECK(qm.offset() == 5);
+                CHECK(qm.linear(0) == 0);
+                CHECK(qm.linear(1) == -1);
+                CHECK(qm.linear(2) == -3);  // this was reindexed
```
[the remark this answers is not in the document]
arcondello: There isn't a way to avoid it, because we're deleting a variable in the middle of the model. The performance of deleting is pretty bad, but luckily it's not used very much.

dimod/include/dimod/abc.h
```diff
@@ -0,0 +1,1083 @@
...
+    bool is_equal(const QuadraticModelBase<B, I>& other) const;
+
+    /// Return True if the model has no quadratic biases.
+    bool is_linear() const;
+
+    /// The linear bias of variable `v`.
+    bias_type linear(index_type v) const;
+
+    /// The lower bound of variable `v`.
+    virtual bias_type lower_bound(index_type) const = 0;
```
amahmudwave, line 250: indetx_type v   for clarity/ consisency

dimod/include/dimod/abc.h
```diff
@@ -0,0 +1,1083 @@
...
+
+    /**
+     * Remove variable `v` from the model.
+     *
+     * Note that this causes a reindexing, where all variables above `v` have
+     * their index reduced by one.
+     */
+    void remove_variable(index_type v);
+
+    /// Multiply all biases 'scalar'
```
amahmudwave, line 320: by scalar

dimod/include/dimod/abc.h
```diff
@@ -0,0 +1,1083 @@
...
+ private:
+    std::vector<bias_type> linear_biases_;
+
+    std::unique_ptr<std::vector<std::vector<LinearTerm<bias_type, index_type>>>> adj_ptr_;
+
+    bias_type offset_;
+
+    // Assumes adj exists!
+    // Creates the bias if it doesn't already exist
+    bias_type& asym_quadratic(index_type u, index_type v) {
```
amahmudwave, line 419: asym -> What does this mean ? Better name could be used.

dimod/include/dimod/abc.h
```diff
@@ -0,0 +1,1083 @@
...
+        if (!adj_ptr_) {
+            adj_ptr_ = std::unique_ptr<std::vector<std::vector<LinearTerm<bias_type, index_type>>>>(
+                    new std::vector<std::vector<LinearTerm<bias_type, index_type>>>(
+                            num_variables()));
+        }
+    }
+
+    /// Return an empty neighborhood - useful when we don't have an adj
+    static const std::vector<LinearTerm<bias_type, index_type>>& empty_neighborhood() {
+        static std::vector<LinearTerm<bias_type, index_type>> empty;
```
amahmudwave, line 445: static ? Is it intenional ?

dimod/include/dimod/abc.h
```diff
@@ -0,0 +1,1083 @@
...
+template <class T>
+void QuadraticModelBase<bias_type, index_type>::add_quadratic_from_dense(const T dense[],
+                                                                         index_type num_variables) {
+    static_assert(std::is_arithmetic<T>::value, "T must be numeric");
+    assert(0 <= num_variables);
+    assert(static_cast<size_type>(num_variables) <= this->num_variables());
+
+    enforce_adj();
+
+    if (is_linear()) {
```
amahmudwave, line 585: This code could be made faster with some fancy tricks, but I guess that is not too important.

dimod/include/dimod/abc.h
```diff
@@ -0,0 +1,1083 @@
...
+template <class bias_type, class index_type>
+template <class Iter>
+bias_type QuadraticModelBase<bias_type, index_type>::energy(Iter sample_start) const {
+    static_assert(std::is_same<std::random_access_iterator_tag,
+                               typename std::iterator_traits<Iter>::iterator_category>::value, "iterators must be random access");
+
+    bias_type en = offset();
+
+    if (has_adj()) {
+        for (index_type u = 0; static_cast<size_type>(u) < num_variables(); ++u) {
```
amahmudwave, line 758: This form of energy calculation is slightly less accurate than the initial calculation in the c++ version where simd code is used and in essence multiple accumulators are used. If this needs to be the most accurate version then we need to use simd code or at least have multiple accumulators or apply a compensation based algorithm like the Kahan sum.

But then again, its probably okay as it is :D

dimod/include/dimod/abc.h
```diff
@@ -0,0 +1,1083 @@
...
+    }
+}
+
+template <class bias_type, class index_type>
+bias_type QuadraticModelBase<bias_type, index_type>::offset() const {
+    return offset_;
+}
+
+template <class bias_type, class index_type>
+bias_type QuadraticModelBase<bias_type, index_type>::quadratic(index_type u, index_type v) const {
```
amahmudwave, line 905: This one does not return a reference but the asym_quad does, so probably better name that one as asym_quadratic_ref ?

dimod/include/dimod/abc.h
arcondello: Yes, we only want one to be allocated, and then we can refer to it globally.

dimod/include/dimod/abc.h
```diff
@@ -0,0 +1,1083 @@
...
+    using value_type = QuadraticTerm<bias_type, index_type>;
+    using pointer = const value_type*;
+    using reference = const value_type&;
+    using difference_type = std::ptrdiff_t;
+    using iterator_category = std::forward_iterator_tag;
+
+    ConstQuadraticIterator() : adj_ptr_(nullptr), term_(0) {}
+
+    ConstQuadraticIterator(
+            const std::vector<std::vector<LinearTerm<bias_type, index_type>>>* adj_ptr,
```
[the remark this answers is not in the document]
arcondello: I'll go with `OneVarTerm` and `TwoVarTerm`. That way we can use `OneVarTerm` with both neighborhoods and linear terms.

dimod/include/dimod/abc.h
arcondello: Agree, but let's leave it alone for right now.

dimod/include/dimod/abc.h
arcondello: Agree, let's leave it alone for right now.

dimod/include/dimod/abc.h
arcondello: Will rename to `asymmetric_quadratic_ref`

arcondello:
I agree with all of the docs feedback, but will do a round updating *all* of the docstrings in a followup PR. For now I am just going to leave them alone.

PR #1258, opened 2022-10-03
Add C++ CQM and Presolve

arcondello:
This is a placeholder PR for ongoing CQM/Presolve work.

Note that this PR is built on top of https://github.com/dwavesystems/dimod/pull/1257, so should probably be merged after.

PR #1262, opened 2022-10-17
Add C++ CQM class

arcondello:
Pulls the C++ CQM code from https://github.com/dwavesystems/dimod/pull/1258.

Does *not* implement docs. I am going to do that in a followup PR.

The `Expression` class also has some `not implemented` errors that will need to be implemented. Also will do in a followup PR.

### DIFF
--- a/dimod/include/dimod/constrained_quadratic_model.h
+++ b/dimod/include/dimod/constrained_quadratic_model.h
@@ -1,0 +1,493 @@
+// Copyright 2022 D-Wave Systems Inc.
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#pragma once
+
+#include <unordered_set>
+#include <utility>
+#include <vector>
+
+#include "dimod/abc.h"
+#include "dimod/constraint.h"
+#include "dimod/expression.h"
+#include "dimod/vartypes.h"
+
+namespace dimod {
+
+template <class Bias, class Index = int>
+class ConstrainedQuadraticModel {
+ public:
+    /// The first template parameter (Bias).
+    using bias_type = Bias;
+
+    /// The second template parameter (Index).
+    using index_type = Index;
+
+    /// Unsigned integral type that can represent non-negative values.
+    using size_type = std::size_t;
+
+    ConstrainedQuadraticModel();
+
+    ConstrainedQuadraticModel(const ConstrainedQuadraticModel& other);
+    ConstrainedQuadraticModel(ConstrainedQuadraticModel&& other) noexcept;
+    ~ConstrainedQuadraticModel() = default;
+    ConstrainedQuadraticModel& operator=(ConstrainedQuadraticModel other);
+    ConstrainedQuadraticModel& operator=(ConstrainedQuadraticModel&& other) noexcept;
+
+    index_type add_constraint();
+
+    template <class B, class I, class T>
+    index_type add_constraint(const abc::QuadraticModelBase<B, I>& lhs, Sense sense, bias_type rhs,
+                              const std::vector<T>& mapping);
+
+    index_type add_constraints(index_type n);
+
+    index_type add_linear_constraint(std::initializer_list<index_type> variables,
+                                     std::initializer_list<bias_type> biases, Sense sense,
+                                     bias_type rhs);
+
+    index_type add_variable(Vartype vartype, bias_type lb, bias_type ub);
+
+    index_type add_variable(Vartype vartype);
+
+    index_type add_variables(Vartype vartype, index_type n);
+
+    index_type add_variables(Vartype vartype, index_type n, bias_type lb, bias_type ub);
+
+    void change_vartype(Vartype vartype, index_type v);
+
+    Constraint<bias_type, index_type>& constraint_ref(index_type c);
+    const Constraint<bias_type, index_type>& constraint_ref(index_type c) const;
+
+    template <class T>
+    void fix_variable(index_type v, T assignment);
+
+    /// Return the lower bound on variable ``v``.
+    bias_type lower_bound(index_type v) const;
+
+    /// Return the number of constraints in the model.
+    size_type num_constraints() const;
+
+    /// Return the number of variables in the model.
+    size_type num_variables() const;
+
+    /// Remove a constraint from the model.
+    void remove_constraint(index_type c);
+
+    void remove_variable(index_type v);
+
+    void set_lower_bound(index_type v, bias_type lb);
+
+    template <class B, class I>
+    void set_objective(const abc::QuadraticModelBase<B, I>& objective);
+
+    template <class B, class I, class T>
+    void set_objective(const abc::QuadraticModelBase<B, I>& objective,
+                       const std::vector<T>& mapping);
+
+    void set_upper_bound(index_type v, bias_type ub);
+    void set_vartype(index_type v, Vartype vartype);
+
+    void substitute_variable(index_type v, bias_type multiplier, bias_type offset);
+
+    /// Return the upper bound on variable ``v``.
+    bias_type upper_bound(index_type v) const;
+
+    /// Return the variable type of variable ``v``.
+    Vartype vartype(index_type v) const;
+
+    Expression<bias_type, index_type> objective;
+
+    friend void swap(ConstrainedQuadraticModel& first, ConstrainedQuadraticModel& second) {
+        first.myswap(second);
+    }
+
+ private:
+    // I can't figure out how to get Expression to grant friendship to friend swap...
+    // Will look into it some other time.
+    void myswap(ConstrainedQuadraticModel& other) {
+        using std::swap;
+
+        // for objective and constraints, we need to make sure that their
+        // parent_ pointers are pointing at the correct object.
+
+        swap(this->objective, other.objective);
+        this->objective.parent_ = this;
+        other.objective.parent_ = &other;
+
+        swap(this->constraints_, other.constraints_);
+        for (auto& c : this->constraints_) {
+            c.parent_ = this;
+        }
+        for (auto& c : other.constraints_) {
+            c.parent_ = &other;
+        }
+
+        swap(this->varinfo_, other.varinfo_);
+    }
+
+    std::vector<Constraint<bias_type, index_type>> constraints_;
+
+    struct varinfo_type {
+        Vartype vartype;
+        bias_type lb;
+        bias_type ub;
+
+        varinfo_type(Vartype vartype, bias_type lb, bias_type ub)
+                : vartype(vartype), lb(lb), ub(ub) {
+            assert(lb <= ub);
+
+            assert(lb >= vartype_info<bias_type>::min(vartype));
+            assert(ub <= vartype_info<bias_type>::max(vartype));
+
+            assert(vartype != Vartype::BINARY || lb == 0);
+            assert(vartype != Vartype::BINARY || ub == 1);
+
+            assert(vartype != Vartype::SPIN || lb == -1);
+            assert(vartype != Vartype::SPIN || ub == +1);
+        }
+
+        explicit varinfo_type(Vartype vartype)
+                : vartype(vartype),
+                  lb(vartype_info<bias_type>::default_min(vartype)),
+                  ub(vartype_info<bias_type>::default_max(vartype)) {}
+    };
+
+    std::vector<varinfo_type> varinfo_;
+};
+
+template <class bias_type, class index_type>
+ConstrainedQuadraticModel<bias_type, index_type>::ConstrainedQuadraticModel()
+        : objective(this), constraints_(), varinfo_() {}
+
+template <class bias_type, class index_type>
+ConstrainedQuadraticModel<bias_type, index_type>::ConstrainedQuadraticModel(
+        const ConstrainedQuadraticModel& other)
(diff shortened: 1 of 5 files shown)


dimod/include/dimod/constrained_quadratic_model.h
```diff
@@ -0,0 +1,493 @@
...
+    void set_lower_bound(index_type v, bias_type lb);
+
+    template <class B, class I>
+    void set_objective(const abc::QuadraticModelBase<B, I>& objective);
+
+    template <class B, class I, class T>
+    void set_objective(const abc::QuadraticModelBase<B, I>& objective,
+                       const std::vector<T>& mapping);
+
+    void set_upper_bound(index_type v, bias_type ub);
```
amahmudwave, line 99: line break

dimod/include/dimod/constrained_quadratic_model.h
```diff
@@ -0,0 +1,493 @@
...
+    void set_vartype(index_type v, Vartype vartype);
+
```
amahmudwave, line 101: This function deserves a comment more than the commented ones.

dimod/include/dimod/constrained_quadratic_model.h
```diff
@@ -0,0 +1,493 @@
...
+template <class bias_type, class index_type>
+ConstrainedQuadraticModel<bias_type, index_type>&
+ConstrainedQuadraticModel<bias_type, index_type>::operator=(
+        ConstrainedQuadraticModel&& other) noexcept {
+    swap(*this, other);
+    return *this;
+}
+
+template <class bias_type, class index_type>
+index_type ConstrainedQuadraticModel<bias_type, index_type>::add_constraint() {
```
amahmudwave, line 207: We could do away with this function, just lets the option of adding 100s of empty constraints. I do not see any value of having this function.

dimod/include/dimod/constrained_quadratic_model.h
```diff
@@ -0,0 +1,493 @@
...
+
+template <class bias_type, class index_type>
+index_type ConstrainedQuadraticModel<bias_type, index_type>::add_linear_constraint(
+        std::initializer_list<index_type> variables, std::initializer_list<bias_type> biases,
+        Sense sense, bias_type rhs) {
+    Constraint<bias_type, index_type> constraint(this);
+
+    auto vit = variables.begin();
+    auto bit = biases.begin();
+    for (; vit != variables.end() && bit != biases.end(); ++vit, ++bit) {
```
amahmudwave, line 264: What happens when one iterator ends up shorter than the other ?

dimod/include/dimod/constrained_quadratic_model.h
```diff
@@ -0,0 +1,493 @@
...
+    constraint.set_sense(sense);
+
+    constraints_.push_back(std::move(constraint));
+    return constraints_.size() - 1;
+}
+
+template <class bias_type, class index_type>
+index_type ConstrainedQuadraticModel<bias_type, index_type>::add_variable(Vartype vartype) {
+    index_type v = num_variables();
+    varinfo_.emplace_back(vartype);
```
amahmudwave, line 278: We dont need to modify the objective ?

dimod/include/dimod/constrained_quadratic_model.h
```diff
@@ -0,0 +1,493 @@
...
+}
+
+template <class bias_type, class index_type>
+std::size_t ConstrainedQuadraticModel<bias_type, index_type>::num_variables() const {
+    return varinfo_.size();
+}
+
+template <class bias_type, class index_type>
+void ConstrainedQuadraticModel<bias_type, index_type>::remove_constraint(index_type c) {
+    constraints_.erase(constraints_.begin() + c, constraints_.begin() + c + 1);
```
amahmudwave, line 387: Are all variables present in the objective ? Are they inserted in the objective when they are added ? If not will removing constraints need to trigger removal of variables which belong only to the constraint removed ?

dimod/include/dimod/constrained_quadratic_model.h
```diff
@@ -0,0 +1,493 @@
...
+}
+
+template <class bias_type, class index_type>
+void ConstrainedQuadraticModel<bias_type, index_type>::remove_variable(index_type v) {
+    assert(v >= 0 && static_cast<size_type>(v) < num_variables());
+    for (auto& constraint : constraints_) constraint.reindex_variables(v);
+    objective.reindex_variables(v);
+    varinfo_.erase(varinfo_.begin() + v);
```
amahmudwave, line 395: What happens when constraints become empty due to removal of variables, do they remain in the system. This will trigger the addition of empty variables to the serverside cqm.

dimod/include/dimod/constrained_quadratic_model.h
```diff
@@ -0,0 +1,493 @@
...
+        constraint.add_linear(mapping[i], lhs.linear(i));
+    }
+
+    // quadratic biases
+    for (auto it = lhs.cbegin_quadratic(); it != lhs.cend_quadratic(); ++it) {
+        constraint.add_quadratic(mapping[it->u], mapping[it->v], it->bias);
+    }
+
+    // offset
+    constraint.add_offset(lhs.offset());
```
amahmudwave, line 237: Does clearing set the offset to zero ?   
Add offset or set offset ? Depending on the question above.

dimod/include/dimod/constrained_quadratic_model.h
```diff
@@ -0,0 +1,493 @@
...
+        this->objective.add_linear(mapping[i], objective.linear(i));
+    }
+
+    // quadratic biases
+    for (auto it = objective.cbegin_quadratic(); it != objective.cend_quadratic(); ++it) {
+        this->objective.add_quadratic(mapping[it->u], mapping[it->v], it->bias);
+    }
+
+    // offset
+    this->objective.add_offset(objective.offset());
```
amahmudwave, line 460: same question as above.

dimod/include/dimod/constrained_quadratic_model.h
```diff
@@ -0,0 +1,493 @@
...
+void ConstrainedQuadraticModel<bias_type, index_type>::remove_variable(index_type v) {
+    assert(v >= 0 && static_cast<size_type>(v) < num_variables());
+    for (auto& constraint : constraints_) constraint.reindex_variables(v);
+    objective.reindex_variables(v);
+    varinfo_.erase(varinfo_.begin() + v);
+}
+
+template <class bias_type, class index_type>
+void ConstrainedQuadraticModel<bias_type, index_type>::set_lower_bound(index_type v, bias_type lb) {
+    varinfo_[v].lb = lb;
```
amahmudwave, line 400: need asserts ?

dimod/include/dimod/constrained_quadratic_model.h
```diff
@@ -0,0 +1,493 @@
...
+        this->objective.add_quadratic(mapping[it->u], mapping[it->v], it->bias);
+    }
+
+    // offset
+    this->objective.add_offset(objective.offset());
+}
+
+template <class bias_type, class index_type>
+void ConstrainedQuadraticModel<bias_type, index_type>::set_upper_bound(index_type v, bias_type ub) {
+    varinfo_[v].ub = ub;
```
amahmudwave, line 465: need asserts ?

dimod/include/dimod/constrained_quadratic_model.h
```diff
@@ -0,0 +1,493 @@
...
+}
+
+template <class bias_type, class index_type>
+void ConstrainedQuadraticModel<bias_type, index_type>::set_vartype(index_type v, Vartype vartype) {
+    varinfo_[v].vartype = vartype;
```
amahmudwave, line 470: need asserts ? , like if the old ub and lb are still valid or not.

dimod/include/dimod/constrained_quadratic_model.h
```diff
@@ -0,0 +1,493 @@
+// Copyright 2022 D-Wave Systems Inc.
```
amahmudwave, line 1: I see lots of asserts, but for this cqm, where performance is less important should we have std::throw insstead of asserts ?

arcondello: The division of labour here is at the C++ level we use asserts for performance, but we test everything and throw at the Python/Cython level.

dimod/include/dimod/constrained_quadratic_model.h
arcondello: In this case, no, because the user might want to temporarily make an invalid model. Applies to the others.

dimod/include/dimod/constrained_quadratic_model.h
arcondello: Yes, it leaves empty variables in the system. This is desired. Presolve removes them later.

dimod/include/dimod/constrained_quadratic_model.h
arcondello: We do not remove unused variables, a user might want to use them later.

dimod/include/dimod/constrained_quadratic_model.h
arcondello: Nope.

dimod/include/dimod/constrained_quadratic_model.h
```diff
@@ -0,0 +1,493 @@
...
+    std::vector<Constraint<bias_type, index_type>> constraints_;
+
+    struct varinfo_type {
+        Vartype vartype;
+        bias_type lb;
+        bias_type ub;
+
+        varinfo_type(Vartype vartype, bias_type lb, bias_type ub)
+                : vartype(vartype), lb(lb), ub(ub) {
+            assert(lb <= ub);
```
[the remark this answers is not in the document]
arcondello: That's all allowed here. Presolve checks it later.

dimod/include/dimod/constrained_quadratic_model.h
arcondello: I use it :smile:

dimod/include/dimod/constrained_quadratic_model.h
arcondello: This is a new one, so the offset starts at 0.

dimod/include/dimod/constrained_quadratic_model.h
arcondello: It uses the shortest.

dimod/include/dimod/expression.h
```diff
@@ -0,0 +1,587 @@
...
+        using difference_type = std::ptrdiff_t;
+        using iterator_category = std::forward_iterator_tag;
+
+        ConstQuadraticIterator() : expression_(nullptr), it_(), end_(), term_(0) {}
+        ConstQuadraticIterator(const Expression* expression,
+                               const abc::ConstQuadraticIterator<bias_type, index_type>& it,
+                               const abc::ConstQuadraticIterator<bias_type, index_type>& end)
+                : expression_(expression), it_(it), end_(end), term_(0) {
+            if (it_ != end_) {
+                term_ = value_type(expression_->variables()[it_->u],
```
amahmudwave, line 63: Could write a short function to create a term from interator and use it here and below.

dimod/include/dimod/expression.h
```diff
@@ -0,0 +1,587 @@
...
+template <class bias_type, class index_type>
+template <class Iter>
+bias_type Expression<bias_type, index_type>::energy(Iter sample_start) const {
+    static_assert(std::is_same<std::random_access_iterator_tag,
+                               typename std::iterator_traits<Iter>::iterator_category>::value,
+                  "iterator must be random access");
+
+    // we could try to do this "virtually" but almost certainly we'll get better
+    // performance by just making a new sample in the order of the underlying base type
+    std::vector<typename std::iterator_traits<Iter>::value_type> subsample;
```
amahmudwave, line 449: Reserve first.

dimod/include/dimod/constraint.h
```diff
@@ -0,0 +1,178 @@
...
+
+template <class bias_type, class index_type>
+Constraint<bias_type, index_type>::Constraint() : Constraint(nullptr) {}
+
+template <class bias_type, class index_type>
+Constraint<bias_type, index_type>::Constraint(parent_type* parent)
+        : base_type(parent),
+          sense_(Sense::EQ),
+          rhs_(0),
+          weight_(std::numeric_limits<bias_type>::infinity()),
```
amahmudwave, line 86: This infinite weight could produce weird values if somehow it remains till the end and this constraint is marked soft. But a smaller value would also allow bugs to be undetected.

dimod/include/dimod/constraint.h
```diff
@@ -0,0 +1,178 @@
...
+    if (!base_type::is_linear() || base_type::num_variables() < 2) return false;
+
+    // all of out variables must be binary
+    for (const auto& v : base_type::variables()) {
+        if (base_type::vartype(v) != Vartype::BINARY) return false;
+    }
+
+    // we get a bit cute here and check the linear biases using the
+    // underlying model, thereby bypassing the dict lookup. This is not
+    // super future-proof but faster.
```
amahmudwave, line 101: I did not understand the logic behind the last part.

dimod/include/dimod/constraint.h
```diff
@@ -0,0 +1,178 @@
...
+        if (base_type::base_type::linear(i) != rhs_) return false;
+    }
+
+    return true;
+}
+
+template <class bias_type, class index_type>
+bool Constraint<bias_type, index_type>::is_soft() const {
+    // some sort of tolerance value?
+    return weight_ != std::numeric_limits<bias_type>::infinity();
```
amahmudwave, line 112: In cqm we have a default weight but a separate variable is_soft, depending on that we either use the wieght or not.

dimod/include/dimod/constraint.h
arcondello: Yup, would much prefer this method and let any potential solver test whether it knows how to handle it.

dimod/include/dimod/constraint.h
arcondello: And that means multiple sources of truth

arcondello:
I am going to go ahead and merge, though I will return to many of the docs comments etc later.

Issue #849, opened 2021-06-21
Consider making BQM.num_variables, BQM.num_interactions and BQM.shape methods rather than properties


arcondello:
Currently, `BQM.num_variables`, `BQM.num_interactions` and `BQM.shape` are all properties rather than methods. I think that it would be better to make them proper methods.

Pro:
* Determining `num_interactions` is an `O(num_variables)` operation, so having it as a property is misleading
* Potentially extendable in the future, e.g. `bqm.num_interactions(v)` could return the degree of `v`
* `DQM.num_variables()` and `DQM.num_cases()` are currently implemented as methods, which is inconsistent

Con:
* Backwards compatibility break, though I think we can mitigate by making them inherit from numeric and callable (see below)
* Having `.shape` as a property is consistent with NumPy

**Additional context**
If we don't want to break backwards compatibility through a deprecation period
```
from collections.abc import Callable

class CallableInt(int, Callable):
    def __call__(self):
        return self
```
works out of the box. Though getting it to raise a deprecation warning if it's not treated as a callable would require something like
```
import warnings

from collections.abc import Callable
from numbers import Integral

class CallableInt(Integral, Callable):
    def __init__(self, value):
       self.value = int(value)

    def __call__(self):
        return self.value

    def __add__(self, other):
        warnings.warn("In the future will be callable", DeprecationWarning)
       return self.value + other

    ...
```

arcondello:
Also `.vartype` while we're at it...

randomir:
Functionally, it makes perfect sense to me to keep all of those as properties.

`num_variables` (or `num_interactions` or `vartype`) is a property of BQM, regardless of implementation details, and the fact that determining the actual value involves some calculation, or it has > O(1) complexity.

OTOH, I also get the performance argument, and "property calc should be cheap".

A middle ground would be caching. Either of property/method result (in which case invalidation is a pain I assume), or keeping actual `num_variables`/etc attributes up to date on each operation?

arcondello:
To hone in on a specific example that this has already come up on, let's compare `BQM.vartype` (property) and `QM.vartype(v)` (method).

If I want to write a function that takes either, I have to do something like
```
for v in model.variables:
    if isinstance(model, BQM):
       vt = model.vartype
    else:
       vt = model.vartype(v)
```
whereas, if we implemented `vartype` as a method in BQM
```
def vartype(self, v=None):
    return self._vartype
```
for BQM, we could have nice generic code.

[...]

PR #1265, opened 2022-10-18
Make BQM.vartype callable

arcondello:
There are many places in the code like https://github.com/dwavesystems/dimod/blob/273fb86cdb3af495b490db11a1ecc941e44da0d2/dimod/constrained.py#L209 that could be generalized by the ability to call `bqm.vartype(v)`.

It would also allow us to make `QuadraticViewsMixin` much more full-featured.

See also #849

### DIFF
--- a/dimod/vartypes.py
+++ b/dimod/vartypes.py
@@ -52,11 +52,15 @@ Examples:
     True
 
 """
+from __future__ import annotations
+
 import enum
 import typing
 import warnings
 
 from collections.abc import Container, Iterable
+
+import dimod.typing
 
 __all__ = ['as_vartype',
            'Vartype', 'ExtendedVartype',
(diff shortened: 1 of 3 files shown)


dimod/vartypes.py
```diff
@@ -117,6 +121,16 @@ class Vartype(enum.Enum):
     INTEGER = DISCRETE = Integers()  # DISCRETE is an alias for INTEGER
     REAL = Real()
 
+    # Dev note:
+    # This allows us to treat QMs and BQMs in the same way, i.e. we can
+    # do bqm.vartype(v).
```
randomir, line 126: Not to mention we now can do `bqm.vartype()(v)()()()()(v)(v)(v)()()` :laughing:

dimod/vartypes.py
```diff
@@ -117,6 +121,16 @@ class Vartype(enum.Enum):
...
+    # It would be better to make BQM.vartype a method, but that would be an
+    # enormous backward compatibility break.
+    # I tried instead making a proxy object, but that doesn't allow syntax like
+    # bqm.vartype is SPIN
+    def __call__(self, v: typing.Optional[dimod.typing.Variable] = None) -> Vartype:
+        return self
```
randomir, line 132: Should `QM.vartype` impl differ (use `v`)?

randomir (approved):
LGTM

dimod/vartypes.py
arcondello: `QM.vartype` is already just a method that required `v`. Though it does mean a user can do `QM.vartype(v)()` :shrug: .

I tried a lot of more restrictive approaches but unfortunately it's super buttoned down.

randomir: Oh, got it. Missed it's already a method.

PR #1272, opened 2022-10-21
Fix performance when loading linear BQMs

arcondello:
There is still a regression when the BQM has quadratic interactions, but that would be a more invasive fix. Will do that in a followup PR.


wbernoudy (approved):
:+1:

PR #1275, opened 2022-10-25
Specify include path in .pxd files rather than setup.py

arcondello:
This lets Cython detect changes to the include files and recompile appropriately.


arcondello:
[df0b382](https://github.com/dwavesystems/dimod/pull/1275/commits/df0b382d21c1cfbc9f690a3fd5eb63c2898642e9) is not actually related, but it resolved the CI errors so might as well do it here.

Issue #1030, opened 2021-10-20
Add a method to specify the value domain of a variable in CQM


qinmaomao:
## Description
Is it possible to add a method to specify the value domain of an integer variable while defining it in CQM? For example, I can define an integer variable by `x = dimod.Integer(lower_bound=0, upper_bound=10)`. However, in some cases, I would like to define it with a specific domain (maybe a list) such as `x = dimod.Integer(domain=[0, 1, 3, 4, 8])`. 

## Possible solution
I noticed this is similar to the `fix_variable()` method in BQM. Since we can *fix* the value of a binary variable, maybe it is also necessary to fix some possible values for an integer variable.

Or, is it possible to add a `!=` constraint to CQM?

Would you please tell me if I missed some information?

arcondello:
Hi @qinmaomao, we don't currently support inequality constraints, although that it something that we're looking into for future releases.

For now, if your domain is small, you could use a discrete variable. E.g. if you have a variable `"i"` that you wish to take `[0, 1, 3, 4, 8]`
```
cqm.add_discrete([("i", 0), ("i", 1), ("i", 3), ("i", 4), ("i", 8)])
i = 0*dimod.Binary(('i', 0)) + 1*dimod.Binary(('i', 1)) + 3*dimod.Binary(('i', 3)) + ...
```
this will create 5 binary variables, one for each value of `"i"`. You can then use `i` in expressions like usual.

alexzucca90:
Another solution to add the `x != a` constraint is to add an extra binary variable `y` to the model and add the following constraint: `2 x*y - x > 2b*y -b`. It may be useful if you only have a small amount of these type of constraints.

PR #1284, opened 2022-11-03
Add experimental support for != constraints 

arcondello:
See also https://github.com/dwavesystems/dimod/issues/1030


alexzucca90:
we need to add  something like this:
```
            elif sense is Sense.Ne:
                # hard-coding violation of not equal constraints to 1
                violation = 0 if activity != 0 else 1
```
In the function `iter_constraint_data` in `constrained.py`

alexzucca90:
I would also raise a `ValueError` when trying to add a constraint with real values, unless we think they make sense.

arcondello:
Closing as stale

PR #1285, opened 2022-11-03
Add test for C++ Constraint::scale()

arcondello:
Written as part of #1284, but generally useful so might as well add it.


alexzucca90 (approved):
ok, LGTM, it took a while because I had to catch up with the latest C++ implementation

PR #1286, opened 2022-11-07
Add nicer iteration over constraints in C++

arcondello:
~Breaks binary compatibility as currently implemented.~ edit: went with a different approach


arcondello:
Looking back over this PR, I think it makes sense to use a method rather than a property. Not least because the method approach does not create a binary incompatibility. The only trick is naming the method.

A. Borrowing from Python 2
```c++
for (auto& constraint : cqm.iterconstraints()) {}
```
B.
```c++
for (auto& constraint: cqm.constraintsview()) {}
```
C.
```c++
for (auto& constraint: cqm.constraints()) {}
```

I think I like (C) the best, and it following the already-establish pattern of
```c++
for (auto& v : cqm.variables()) {}
```
though in the variables case the return value is always const, whereas in the constraints case they are modifiable.

dimod/include/dimod/iterators.h
```diff
@@ -0,0 +1,124 @@
...
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#pragma once
+
+namespace dimod {
+namespace iterators {
+
+template <class Iter, class T = typename std::iterator_traits<Iter>::value_type::element_type>
```
amahmudwave, line 20: Will it work correctly if T is not the default one you are setting. Are there corner cases for that ?

dimod/include/dimod/iterators.h
```diff
@@ -0,0 +1,124 @@
...
+ public:
+    using value_type = T;
+    using pointer = value_type*;
+    using reference = value_type&;
+    using difference_type = typename std::iterator_traits<Iter>::difference_type;
+    using iterator_category = std::random_access_iterator_tag;
+
+    explicit ConstraintIterator(Iter& iterator) : iterator_(iterator) {}
+
+    reference operator*() { return *(*iterator_); }
```
amahmudwave, line 40: Is it trying to create a temporary object and then returning its reference, will it exist after the function quits ?

dimod/include/dimod/iterators.h
```diff
@@ -0,0 +1,124 @@
...
+        tmp += n;
+        return tmp;
+    }
+    ConstraintIterator operator-(difference_type n) {
+        ConstraintIterator tmp(*this);
+        tmp -= n;
+        return tmp;
+    }
+
+    friend difference_type operator-(const ConstraintIterator& a, const ConstraintIterator& b) {
```
amahmudwave, line 83: Is there any form of check needed. like for vectors I pass in the iterators of two different vectors to get the difference. Do we need to check if the parent object of both iterators are the same ?

testscpp/tests/test_constrained_quadratic_model.cpp
```diff
@@ -488,6 +488,70 @@ TEST_CASE("Bug 0") {
...
+        auto x = cqm.add_variable(Vartype::BINARY);
+        for (int i = 0; i < 10; ++i) {
+            cqm.add_linear_constraint({x}, {static_cast<double>(i)}, Sense::EQ, i);
+        }
+
+        THEN("range-based for loops work over the constraints") {
+            int i = 0;
+            for (auto& c : cqm.constraints()) {
+                CHECK(c.linear(x) == i);
+                c.set_linear(x, i - 1);  // can modify
```
amahmudwave, line 503: Need to check modification by another CHECK after set_linear.

testscpp/tests/test_iterators.cpp
```diff
@@ -0,0 +1,83 @@
...
+                CHECK(it == make_constraint_iterator(vec.begin()));
+                CHECK(it != make_constraint_iterator(vec.end()));
+
+                CHECK(it < make_constraint_iterator(vec.end()));
+                CHECK(!(it < make_constraint_iterator(vec.begin())));
+
+                CHECK(make_constraint_iterator(vec.end()) > it);
+                CHECK(!(make_constraint_iterator(vec.begin()) > it));
+
+                CHECK(it <= make_constraint_iterator(vec.end()));
```
amahmudwave, line 42: For the last two operators we are missing one case each.

testscpp/tests/test_iterators.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+                CHECK(it-- == make_constraint_iterator(vec.begin() + 1));
+                CHECK(it == make_constraint_iterator(vec.begin()));
+            }
+        }
+
+        THEN("we can add and subtract") {
+            auto it1 = make_constraint_iterator(vec.begin() + 3);
+            auto it2 = make_constraint_iterator(vec.begin());
+            CHECK(it1 - it2 == 3);
```
amahmudwave, line 65: you are checking subtraction not addition.

dimod/include/dimod/constrained_quadratic_model.h
```diff
@@ -21,13 +21,64 @@
...
+    explicit ConstraintsView(CQM* parent) : parent_(parent) {}
+
+    constraint_type& operator[](index_type c) { return *(parent_->constraints_[c]); }
+    const constraint_type& operator[](index_type c) const { return *(parent_->constraints_[c]); }
+
+    constraint_type& at(index_type c) { return *(parent_->constraints_.at(c)); }
+    const constraint_type& at(index_type c) const { return *(parent_->constraints_.at(c)); }
+
+    iterator begin() { return iterators::make_constraint_iterator(parent_->constraints_.begin()); }
+    const_iterator begin() const {
```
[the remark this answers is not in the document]
arcondello: Following the [std::vector::begin()](https://cplusplus.com/reference/vector/vector/begin/) in c++11 and newer

dimod/include/dimod/constrained_quadratic_model.h
```diff
@@ -86,6 +140,15 @@ class ConstrainedQuadraticModel {
 
     void clear();
 
+    /// Return a view over the constraints that can be iterated over.
```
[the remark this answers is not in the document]
arcondello: Bad grammar, can adjust the docstring. The _view_ can be iterated over, not the constraints.

dimod/include/dimod/constrained_quadratic_model.h
```diff
@@ -86,6 +140,15 @@ class ConstrainedQuadraticModel {
...
+    /// @code
+    /// for (auto& constraint : cqm.constraints()) {}
+    /// @endcode
+    ConstraintsView<ConstrainedQuadraticModel<bias_type, index_type>> constraints();
+
+    /// @copydoc ConstrainedQuadraticModel::constraints()
```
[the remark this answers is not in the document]
arcondello: Doxygen. It just is copying the docs from the non-const

dimod/include/dimod/iterators.h
arcondello: No temporary objects here I don't think? It's the reference to the object pointed to by the iterator.

dimod/include/dimod/iterators.h
arcondello: We defer that to the underlying iterator type.

arcondello:
CI errors are due to codecov, see https://github.com/dwavesystems/dimod/pull/1323

PR #1290, opened 2022-11-14
Use cibuildwheel to run tests in CI

arcondello:
Let cibuildwheel handle the Python install etc for the tests. This also saves a lot of boilerplate and a bit of time when running CI


arcondello:
I played around with using cibw caching with circleci caches, but circleci is so slow at retrieving the cache that it doesn't make any difference in time.

randomir (approved):
LGTM!

PR #1294, opened 2022-11-21
Improve constraint weight manipulation

arcondello:


dimod/constrained/cyexpression.pyx
```diff
@@ -362,16 +362,73 @@ cdef class cyConstraintView(_cyExpression):
...
         elif penalty == cppPenalty.QUADRATIC:
             return "quadratic"
         elif penalty == cppPenalty.CONSTANT:
-            return "constrant"
+            # user should never see this, but might as well be future proof
+            return "constant"
         else:
             raise RuntimeError("unexpected penalty")
 
+    def set_weight(self, weight, penalty='linear'):
```
randomir, line 387: Perhaps use enums for penalty type?

dimod/constrained/cyexpression.pyx
```diff
@@ -362,16 +362,73 @@ cdef class cyConstraintView(_cyExpression):
...
+
+        cdef cppConstraint[bias_type, index_type]* constraint = self.constraint()
+
+        cdef cppPenalty _penalty
+        if penalty == 'linear':
+           _penalty = cppPenalty.LINEAR
+        elif penalty == 'quadratic':
+            for i in range(constraint.num_variables()):
+                if (self.parent.cppcqm.vartype(constraint.variables()[i]) != cppVartype.BINARY
+                        and self.parent.cppcqm.vartype(constraint.variables()[i]) != cppVartype.SPIN):
```
randomir, line 418: Perhaps this would be faster _and_ more readable:
```suggestion
                vartype = self.parent.cppcqm.vartype(constraint.variables()[i])
                if vartype not in (cppVartype.BINARY, cppVartype.SPIN):
```

randomir (approved):

dimod/constrained/cyexpression.pyx
arcondello: I would love to, unfortunately to maintain compatibility with 0.11.6 we're stuck with strings. Could go through a deprecation period but...

dimod/constrained/cyexpression.pyx
arcondello: I had done it this way under the assumption that Cython wouldn't support the syntax you propose, but it generates
```C++
      switch (__pyx_v_vartype) {
        case dimod::BINARY:
        case dimod::SPIN:
        __pyx_t_2 = 0;
        break;
        default:
        __pyx_t_2 = 1;
        break;
      }
      __pyx_t_9 = (__pyx_t_2 != 0);
      if (unlikely(__pyx_t_9)) {
/* … */
      }
```
which is pretty damn neat.

Issue #1235, opened 2022-08-16
Add compress keyword argument to CQM.to_file() and BQM.to_file()


arcondello:
To match the DQM https://github.com/dwavesystems/dimod/blob/5d2043d0b662f38bd51e062367c2e4b703b457a1/dimod/discrete/discrete_quadratic_model.py#L767

arcondello:
CQM is handled by https://github.com/dwavesystems/dimod/pull/1296. Doing BQM would be a bit more complicated because we don't use zipfile.

PR #1296, opened 2022-11-22
Add compress kwarg to CQM.to_file()

arcondello:
Address the CQM part of https://github.com/dwavesystems/dimod/issues/1235.

Because we use `zipfile` for our CQMs, we get backwards compatibility for free.

### DIFF
--- a/dimod/constrained/constrained.py
+++ b/dimod/constrained/constrained.py
@@ -1668,7 +1668,10 @@ class ConstrainedQuadraticModel(cyConstrainedQuadraticModel):
 
         return mapping
 
-    def to_file(self, *, spool_size: int = int(1e9)) -> tempfile.SpooledTemporaryFile:
+    def to_file(self, *,
+                spool_size: int = int(1e9),
+                compress: bool = False,
+                ) -> tempfile.SpooledTemporaryFile:
         """Serialize to a file-like object.
 
         Args:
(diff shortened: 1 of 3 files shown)


dimod/constrained/constrained.py
```diff
@@ -1783,7 +1789,8 @@ def to_file(self, *, spool_size: int = int(1e9)) -> tempfile.SpooledTemporaryFil
         write_header(file, CQM_MAGIC_PREFIX, data, version=(1, 3))
 
         # write the values
-        with zipfile.ZipFile(file, mode='a') as zf:
+        kwargs = dict(compression=zipfile.ZIP_DEFLATED) if compress else dict()
+        with zipfile.ZipFile(file, mode='a', **kwargs) as zf:
```
randomir, line 1793: My only doubt in this PR was about the compression method and compression level to use.

We have `bzip2` and `lzma` at our disposal, in addition to the default `gzip`.

[Looks like](https://tukaani.org/lzma/benchmarks.html) `bzip2` provides 15% better compression over `gzip`, but it's ~8x slower at (de)compression.

`lzma` halves the archive size over `gzip`, but it's a lot slower than `bzip2` during compression, although roughly as fast as `gzip` during decompression.

`gzip`/deflate dominates in memory overhead.

In addition, `bzip2`/`lzma` modules might not be available on all systems, especially if user self-compiled Python, because system packages need to installed beforehand.

Taking all that into consideration, looks like `ZIP_DEFLATED` with the default compression level (6 - balance between speed and size) is the optimal choice here. It's widely supported, it's fast, with minimal memory overhead, and provides decent compression (at least to handle obviously redundant sequences).

randomir (approved):
LGTM!

PR #1297, opened 2022-11-24
Restructure reference and intro docs

JoelPasvolsky:
Mostly just a proposed restructuring, content will get refreshed in a separate PR.


docs/intro/intro_models.rst
```diff
@@ -140,15 +196,30 @@ non-zero biases.
...
+QM Example: Interaction Between Integer Variables
+-------------------------------------------------
 
 This example constructs a QM with an interaction between two integer variables.
 
 >>> qm = dimod.QuadraticModel()
 >>> qm.add_variables_from('INTEGER', ['i', 'j'])
 >>> qm.add_quadratic('i', 'j', 1.5)
 
+Higher-Order Example
```
arcondello, line 208: I wouldn't even bother with this.

docs/reference/generators.rst
```diff
@@ -4,6 +4,18 @@
 Generators and Application Modeling
 ===================================
 
+Constructors
```
arcondello, line 7: This one, I think, does make sense to include with the `BQM` class. I know it's a bit inconsistent, but I think that's much closer to its intent.

arcondello (approved):
Looks great!

PR #1301, opened 2022-12-09
Add property_or_method decorator

### LINKED ISSUE #849: Consider making BQM.num_variables, BQM.num_interactions and BQM.shape methods rather than properties

arcondello:
Closes https://github.com/dwavesystems/dimod/issues/849

### DIFF
--- a/dimod/binary/binary_quadratic_model.py
+++ b/dimod/binary/binary_quadratic_model.py
@@ -41,7 +41,7 @@ except ImportError:
 from dimod.binary.cybqm import cyBQM_float32, cyBQM_float64
 from dimod.binary.pybqm import pyBQM
 from dimod.binary.vartypeview import VartypeView
-from dimod.decorators import forwarding_method, unique_variable_labels
+from dimod.decorators import forwarding_method, unique_variable_labels, property_or_method
 from dimod.quadratic import QuadraticModel, QM
 from dimod.quadratic.quadratic_model import _VariableArray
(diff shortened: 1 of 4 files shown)


arcondello:
Closing as stale

Issue #1303, opened 2022-12-13
Update CQM serialization to allow variable data to be encoded outside of the objective


arcondello:
Right now, the CQM serialization tracks the variable types and bounds in the objective. This is a legacy from `dimod<0.12.0` where the CQM itself tracked them that way. These days, the objective can also be sparse, but [we've kept the serialization](https://github.com/dwavesystems/dimod/blob/f4ad3853592f6a80786590be25cabf20263f6f1d/dimod/constrained/constrained.py#L1708-L1726). 

We should create a new CQM serialization version 1.4 that stores the variable info globally so that the objective can be sparse or even not present.

Issue #1302, opened 2022-12-12
Create sparser quadratic model serialization for linear models


arcondello:
[Currently for quadratic models](https://github.com/dwavesystems/dimod/blob/0.12.3/dimod/quadratic/quadratic_model.py#L1374-L1376), each neighborhood is encoded whether or not it's empty. [Each neighborhood uses 8 bytes to encode its magic string and the length of the section, and then is padded to be 64 bytes](https://github.com/dwavesystems/dimod/blob/0.12.3/dimod/serialization/fileview.py#L75-L101). This is, of course, very inefficient for linear models.

Some alternatives:
1) Add a flag to the `data` header that indicates that the model is linear, and skip the neighborhoods sections entirely in that case.
2) Update the loader to check whether the sections are present or not, and proceed if they are not.
3) Update the `NeighborhoodSection` specification to encode the variable index for each neighborhood, so that we only save/load the neighborhoods that are actually used.

I think (3) ends up being best, but unfortunately requires a backwards incompatible bump in the serialization format version. Whereas (1) and (2) can be done in a backwards compatible way.

arcondello:
Another option:

4) Create a CQM-specific serialization format that doesn't use the QM serialization format and allow that to handle this issue.

I don't really like this option because it breaks the hierarchy and complicates the CQM's serialization. But worth mentioning for completeness.

And of course there is

5) Do something totally different. E.g. adopt another MIQP serialization format.

randomir:
I don't see how any of these options is backwards compatible, except that (1) and (2) can be made so that we on server-side can relatively easy disambiguate between "old format" and "new format" used by older/newer client-side dimod, right?

Can't the best solution (3) be implemented as a new format (with new version), and then we add support for it server-side, in addition to old formats? We'll have to support a range of input formats (versions) anyway.

arcondello:
You're right. I mean to say that for e.g. (1) we can easily interpret the lack of flag as falsy thereby making it a minor version bump. Whereas for (3) we need to significantly change the reading to switch based on version number, which IMO means going to serialization 2.0. Of course we could stretch the definition such that (3) can be version 1.1, but that would be a mess.

randomir:
Incrementing to 2.0 sounds fine, but why can't we support 1.x and 2.x? Should be technically feasible. I mean, we support(ed) different serialization formats/versions in dimod.

arcondello:
[...]

PR #1314, opened 2023-03-06
CQM serialization 2.0

arcondello:
Closes https://github.com/dwavesystems/dimod/issues/1303
Closes https://github.com/dwavesystems/dimod/issues/1302 by creating a new "quadratic" section that encodes the quadratic interactions in COO-format. This will be about 50% larger than necessary in a fully connected quadratic model, but much smaller when the model is linear or sparse. Given that the majority of constraints are sparse, it makes sense to optimize for that case.


arcondello:
Closing for now. Will re-open after another pass.

dimod/constrained/constrained.py
```diff
@@ -1021,32 +1092,40 @@ def from_file(cls,
 
         header_info = read_header(file_like, CQM_MAGIC_PREFIX)
 
-        if not (1, 0) <= header_info.version <= (1, 3):
+        num_variables = header_info.data["num_variables"]
+
+        if not (1, 0) <= header_info.version <= (2, 0):
```
randomir, line 1097: Is `1.0` actually supported?

dimod/constrained/constrained.py
```diff
@@ -1623,41 +1713,55 @@ def to_file(self, *,
...
+                        [penalty]
+                        [weight]
+                    ...
+                objective
+                varinfo
+                [variable_labels.json]
+
+            The ``objective`` file encodes the objective.
+            See :meth:`~dimod.constrained.expression.ObjectiveView._to_file()`
+            for details about the file format.
```
randomir, line 1745: It would be nice to have the complete format spec decoupled from dimod internals.

dimod/constrained/constrained.py
```diff
@@ -1771,7 +1851,7 @@ def violations(self, sample_like: SamplesLike, *,
             requirement of a second, and violates a third.
 
             >>> i, j, k = dimod.Binaries(['i', 'j', 'k'])
-            >>> cqm = dimod.ConstrainedQuadraticModel()
+                >>> cqm = dimod.ConstrainedQuadraticModel()
```
randomir, line 1854: typo

randomir (approved):
Looks good to me overall, but I would suggest extending/introducing a bit more of LBYL safety checks (e.g. edge cases when input data is invalid/unexpected/malicious). Also, extending tests beyond these nice high-level ones to catch lower-level decoding/encoding errors of unexpected inputs.

dimod/constrained/constrained.py
arcondello: Yep, the difference is actually quite small. Agree this need a documentation pass to help make this stuff clearer.

dimod/constrained/constrained.py
arcondello: This is just deferring the documentation to another function so that it's easier to maintain. Though overall the documentation for this could/should be improved.

dimod/constrained/constrained.py
arcondello, line 1854:
```suggestion
            >>> cqm = dimod.ConstrainedQuadraticModel()
```

PR #1316, opened 2023-03-10
Make optional sampler args keyword-only

arcondello:
This helps avoid confusing results when switching between samplers.

This is a backwards compatibility break, but a minor one so I think it's safe.


randomir (approved):
LGTM!

Issue #1317, opened 2023-03-15
fix_variables optimised for fixing large numbers of variables


mhlr:
**Application**
We should implement a variant of fix_variables that is optimised for fixing a large number of variables. The current implementation iterates over the variables to be fixed first, then finds the constraints the current variable participates in and fixes it in those constraints. This is good for fixing a small number of variables that contribute to a small number of constraints, but inefficient on large number of variables.

It does not complete overnight whe used to extract a solver sized subproblems from a problem 10x the solver limit

**Proposed Solution**
When fixing many/most of the variables of a large model, we want to iterate over all constraints directly and fix all the variables in that constraint in one pass. I have implemented a pure python version of this for pure discrete linear problems that runs in minutes on problems that fix_variables is unble to complete overnight

**Alternatives Considered**

**Additional Context**

PR #1319, opened 2023-03-27
Improve the performance of removing variables from expressions

arcondello:
This provides a pretty significant performance improvement when fixing variables in individual constraints.
```python
import time

import dimod

num_variables = 10_000

cqm = dimod.ConstrainedQuadraticModel()
cqm.add_variables("BINARY", num_variables)
c0 = cqm.add_constraint(((v, 1) for v in range(num_variables)), '==', 1)
expr = cqm.constraints[c0].lhs

t = time.perf_counter()
    expr.fix_variable(v, 0)
print(time.perf_counter() - t)
```
Gives a time of ~0.15s vs 1.37s on 0.12.4 on my system.

The majority of the improvement comes from optimizing `remove_variable()`. In fact, the C++ `fix_variable()` is not even called from the Python code. I did test adding a dedicated Cython call rather than [using the view](https://github.com/dwavesystems/dimod/blob/5ef7897af313fcad59b4fd38617d07904ffb6470/dimod/views/quadratic.py#L429-L448), but that doesn't change the performance very much at all since the vast majority of the time is spent on rebuilding the indices.

*Additional Context*
This is a first step to addressing https://github.com/dwavesystems/dimod/issues/1317. But it doesn't actually touch the relevant code (yet).

### DIFF
--- a/dimod/include/dimod/abc.h
+++ b/dimod/include/dimod/abc.h
@@ -707,7 +707,7 @@ void QuadraticModelBase<bias_type, index_type>::fix_variable(index_type v, T ass
     add_offset(assignment * linear(v));
 
     // finally remove v
-    remove_variable(v);
+    QuadraticModelBase<bias_type, index_type>::remove_variable(v);
 }
 
 template <class bias_type, class index_type>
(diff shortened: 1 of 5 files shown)


randomir (approved):
LGTM!

PR #1324, opened 2023-04-14
Generate an .xml file for Python coverage

arcondello:
Codecov does not support `coverage`'s default file format ([link](https://docs.codecov.com/docs/supported-report-formats)). So we explicitly cast to xml.


randomir (approved):
LGTM!

PR #1333, opened 2023-06-12
Feature/lprefresh

arcondello:
The first two commits are just adding HiGHS' LP file reader using git subtree. Previously we had wrapped it directly in the code and made invasive changes. This approach will be far more maintainable as it will be easier to add upstream features/fixes as they come.

This PR also includes the fix included in https://github.com/ERGO-Code/HiGHS/pull/1318. If/when that is merged we can pull in the changes.


randomir (approved):
Looks good to me!

PR #1334, opened 2023-06-13
Ignore .dev and rc version in dimod serialization tests

arcondello:


tests/test_serialization_cqm.py
```diff
@@ -31,13 +31,14 @@
...
 import os.path
+import re
 import shutil
 import typing
 import unittest
 
 import dimod
 
-DIMOD_VERSION = tuple(map(int, dimod.__version__.split(".")))[:3]
+DIMOD_VERSION = tuple(map(int, re.search(r"\d+.\d+.\d+", dimod.__version__).group(0).split(".")))
```
randomir, line 41: Should escape the dot. As it stands, this regex matches `"1a2b3"`.

PR #1337, opened 2023-06-19
Support large constraints in serialization zip file

arcondello:
Unfortunately there is not a fast distinguishing case that I can add to the tests without significantly slowing them down. I have tested this locally though and because it is always on we can test that it works correctly with the existing tests.


arcondello:
Fixes a bug where the 4 bytes encoding the length of the quadratic section were not sufficient for very large constraints/objectives. We could not simply change the number of bytes for all sections without incurring other backwards compatibility breaks, so we extend it only for the new `QuadraticSection` introduced in 0.12.8rc0.

**This breaks backwards compatibility with 0.12.8rc0 and 0.12.8rc1**. But, because those are not official releases, I think that's OK.

dimod/constrained/constrained.py
```diff
@@ -1717,6 +1714,37 @@ def to_file(self, *,
...
+                data = dict(shape=expr.shape,
+                            dtype=expr.dtype.name,
+                            itype=expr.index_dtype.name,
+                            type=type(expr).__name__,
+                            )
+
+            it is terminated by a newline character and padded with spaces to
+            make the entire length of the entire header divisible by 64.
+
+            The expression data comes after the header.
```
arcondello, line 1746: This still needs more explanation, but I have simply copy and pasted from the other file to minimize the changes introduced in this PR. Improving docs can come in followup PRs.

randomir (approved):
LGTM!

PR #1338, opened 2023-06-19
Handle tuple labels (and others) when deserializing CQMs

arcondello:


tests/test_constrained.py
```diff
@@ -1225,6 +1225,44 @@ def test_functional_soft(self):
...
+        cqm = dimod.ConstrainedQuadraticModel()
+
+        dimod_vars = [
+            dimod.Integer(i, lower_bound=-5000, upper_bound=5000) for i in range(63)
+        ]
+        num_discrete = 3
+        num_cases = 7
+        dimod_vars.extend(
+            [
+                dimod.Binary((c, i))
```
randomir, line 1238: Perhaps also add a case with nested tuples, and/or other collection types, like: `v = [{(1,(2,3)),4},5]`.

randomir (approved):
LGTM!

PR #1344, opened 2023-06-27
Add C++ CQM::remove_constraints_if() method

arcondello:
This method provides a significant speedup when removing constraints in bulk at the C++ level.


wbernoudy (approved):
:basketball:

PR #1348, opened 2023-06-29
Feature/expose onehot

kevinchern:


arcondello (commented):
This is going to need some tests.

kevinchern:
@arcondello added tests. I did not include cases where the constant is split, e.g., `x + y - 1 == 0`. The current implementation does not recognize this constraint as a one-hot constraint.

arcondello:
Hmm, I see my own note about this https://github.com/dwavesystems/dimod/blob/3873226f4d32d45d0554f569769fec4f7bf80c49/dimod/include/dimod/constraint.h#L125-L127 but honestly I cannot remember my reasoning.. ah well. Let's assume there was a good reason for now.

releasenotes/notes/expose-is_onehot-044d860d5eb7b115.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Expose method ``is_onehot``.
```
arcondello, line 4:
```suggestion
    Add ``Constraint.is_onehot()`` method.
```
We know that we're just exposing it, but not sure a typical user does.

tests/test_constrained.py
```diff
@@ -2003,3 +2031,6 @@ def test_set_weight_binary(self):
         self.assertFalse(cqm.constraints[c].lhs.is_soft())
         self.assertEqual(cqm.constraints[c].lhs.weight(), float('inf'))
         self.assertIs(cqm.constraints[c].lhs.penalty(), None)
+
+if __name__ == "__main__":
+    unittest.main()
```
arcondello, line 2036: This is not needed to run the tests. In fact it's preferred to do `python -m unittest tests/test_constrained.py`

arcondello (approved):

PR #1354, opened 2023-08-11
Convert floats representing integers to int in sample set serialization

arcondello:
This does come with a performance hit. However it is relatively minor. Testing two edge cases on my system

```python
import time

import dimod
import numpy as np

arr = np.full((100, 500000), 1, dtype=float)  # can be converted

t = time.perf_counter()
dimod.serialization.utils.serialize_ndarray(arr)
print(time.perf_counter() - t)
```
takes `~5.5` seconds on this branch and `~1` second on dimod 0.12.10.

```python
import time

import dimod
import numpy as np

arr = np.full((100, 500000), 1.5, dtype=float)  # cannot be converted

t = time.perf_counter()
dimod.serialization.utils.serialize_ndarray(arr)
print(time.perf_counter() - t)
```
takes `~3.5` seconds on this branch and `~1` second on dimod 0.12.10.

For context, actually Json-serializing the resulting object at this scale takes `~27.8` seconds to dump to a file and `~6.7` seconds to dump to a string.

### DIFF
--- a/dimod/sampleset.py
+++ b/dimod/sampleset.py
@@ -1767,7 +1767,7 @@ class SampleSet(abc.Iterable, abc.Sized):
             :meth:`~.SampleSet.from_serializable`
 
         """
-        schema_version = "3.1.0"
+        schema_version = "3.2.0"
 
         # developer note: numpy's record array stores the samples, energies,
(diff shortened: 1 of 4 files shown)


randomir (approved):
LGTM

jdklassen (approved):

PR #1356, opened 2023-08-14
Fix SampleSet.relabel_variables() method

arcondello:


dimod/sampleset.py
```diff
@@ -1450,17 +1449,33 @@ def relabel_variables(self, mapping, inplace=True):
...
+
+        elif done:  # and not inplace
             return self.copy().relabel_variables(mapping, inplace=True)
 
-        if not self.done():
+        elif inplace:  # and not done
+            old_hook = self._result_hook
+
+            def new_hook(future):
+                sampleset = old_hook(future)
```
JoelPasvolsky, line 1465: Why define `old_hook` rather than just use ` sampleset = self._result_hook(future)` ?

JoelPasvolsky (approved):
A bit quick but nothing stands out to me

dimod/sampleset.py
arcondello: That creates a copy (and was the bugged pattern it was using before).

PR #1357, opened 2023-08-18
Add MIMO generators

JoelPasvolsky:


dimod/generators/mimo.py
```diff
@@ -0,0 +1,1085 @@
...
+    if attenuation_matrix is not None:
+        if np.iscomplex(attenuation_matrix).any():
+            raise ValueError('attenuation_matrix must not have complex values')
+        F = F * attenuation_matrix    #Dense format for now, this is slow.
+        channel_power *= np.mean(np.sum(attenuation_matrix * attenuation_matrix,
+                                        axis=0)) / num_receivers
+
+    return F, channel_power
+    
+constellation = {   # bits per transmitter (bpt) and amplitudes (amps)
```
arcondello, line 397: I am not sure why this is defined here specifically. It should probably be at the top unless it's only used in one function.

dimod/generators/mimo.py
```diff
@@ -0,0 +1,1085 @@
...
+    Returns:
+        Four-tuple of received signals (``y``), transmitted symbols (``v``),
+        channel noise, and random_state, where ``y`` is a column vector of length
+        equal to the rows of ``F``.
+    """
+
+    num_receivers = F.shape[0]
+    num_transmitters = F.shape[1]
+
+    random_state = _make_random_state(random_state)
```
arcondello, line 513:
```suggestion
    rng = np.random.default_rng(seed)
```
to be consistent with NumPy and other (modern) parts of Ocean.

dimod/generators/mimo.py
```diff
@@ -0,0 +1,1085 @@
...
+        channel_noise = sigma*(random_state.normal(0, 1, size=(num_receivers, 1)) \
+            + 1j*random_state.normal(0, 1, size=(num_receivers, 1)))
+        if modulation == 'BPSK' and np.isreal(F).all():
+            channel_noise = channel_noise.real
+
+        y = channel_noise + np.matmul(F, transmitted_symbols)
+
+    return y, transmitted_symbols, channel_noise, random_state
+
+def spin_encoded_mimo(modulation: str,
```
arcondello, line 564: As this method returns a BQM that has a `.vartype` attribute, IMO we should just call this `encoded_mimo()`.

arcondello, line 564:
```suggestion
def spin_encoded_mimo(modulation: typing.Literal["BPSK", "QPSK", "16QAM", "64QAM", "256QAM"], 
```

dimod/generators/mimo.py
```diff
@@ -0,0 +1,1085 @@
...
+    subject to additive white Gaussian noise. Given the received signal, 
+    :math:`y`, the log likelihood of a given symbol set, :math:`v`, is 
+    :math:`MLE = argmin || y - F v ||_2`. When :math:`v` is encoded as 
+    a linear sum of spins, the optimization problem is a binary quadratic model.
+    
+    Depending on its parameters, this function can model code division multiple
+    access (CDMA) _[#T02, #R20], 5G communication networks _[#Prince], or 
+    other problems.
+
+    Args:
```
arcondello, line 593: Args are not documented in the same order as the function signature.

requirements.txt
```diff
@@ -4,3 +4,6 @@ cython==0.29.33
 reno==3.3.0  # for changelog
 
 setuptools>=46.4.0  # to support setup.cfg getting __version__
+
+networkx
```
arcondello, line 8: Unfortunately I don't think we want to add NetworkX as a direct dependency of dimod just for these methods. It doesn't look like any of NetworkX's algorithms are used so probably we can replace the use of the Graph class internally.

We do provide some support for functions accepting graphs if NetworkX is installed, see https://github.com/dwavesystems/dimod/blob/8433f221a1e79101e1db0d80968ab5a2f59b865d/dimod/decorators.py#L228 but we use more generic structured internally so we don't require the user have it installed.

tests/test_generators.py
```diff
@@ -1129,3 +1127,558 @@ def test_constraints_squares(self):
...
+            #uniform encoding (max spins = max amplitude symbols):
+            spins = np.ones(num_spins)
+            symbols = max_symb*np.ones(num_symbols) + 1j*max_symb*np.ones(num_symbols)
+            symbols_enc = dimod.generators.mimo._spins_to_symbols(spins=spins, modulation=mod)
+            self.assertTrue(np.all(symbols_enc == symbols ))
+            spins_enc = dimod.generators.mimo._symbols_to_spins(symbols=symbols, modulation=mod)
+            self.assertTrue(np.all(spins_enc == spins))
+
+            #random encoding:
+            spins = np.random.choice([-1, 1], size=num_spins)
```
arcondello, line 1399: We should probably fix a seed.

dimod/generators/mimo.py
```diff
@@ -0,0 +1,1085 @@
...
+            For large numbers of receivers and transmitters, statistical 
+            properties of the likelihood are weakly dependent on the 
+            distribution. Choosing ``'binary'`` allows for integer-valued 
+            Hamiltonians while ``'normal'`` is a more typical model. The channel 
+            can be real or complex; in many cases this represents a superficial 
+            distinction up to rescaling. For real-valued symbols (BPSK) the 
+            default is ``('normal', 'real')``; otherwise, the default is 
+            ``('normal', 'complex')``.
+        
+        use_offset: Adds a constant to the Ising model energy so that the
```
arcondello, line 931: Why not always include the offset? It's hard for me to imagine what harm it would do. And if the user really doesn't want it they can just
```python
bqm.offset = 0
```

dimod/generators/mimo.py
```diff
@@ -0,0 +1,1085 @@
...
+                amount from :math:`[1, 3]`.
+            * '256QAM'
+                A QPSK symbol set is generated and symbols are further amplitude 
+                modulated by an independently and uniformly distributed random 
+                amount from :math:`[1, 3, 5]`.
+ 
+        y:  Complex- or real-valued received signal, as a NumPy array. If 
+            ``None``, generated from other arguments.
+
+        F:  Transmission channel. Currently not supported and must be ``None``.
```
arcondello, line 871: If it's not supported, why include it at all?

dimod/generators/mimo.py
```diff
@@ -0,0 +1,1085 @@
...
+import dimod
+
+__all__ = ['spin_encoded_mimo', 'spin_encoded_comp', ]
+
+mod_config = { # bits per transmitter, amplitudes, transmitters_per_spin, number of amps, spins per symbol
+    "BPSK": {"bpt": 1, "amps": 1, "tps": 1, "na": 1, "sps": 1},
+    "QPSK": {"bpt": 2, "amps": 1, "tps": 2, "na": 1, "sps": 1},
+    "16QAM": {"bpt": 4, "amps": 2, "tps": 4, "na": 2, "sps": 2},
+    "64QAM": {"bpt": 6, "amps": 4, "tps": 6, "na": 3, "sps": 3},
+    "256QAM": {"bpt": 8, "amps": 8, "tps": 8, "na": 5, "sps": 4}   #JP: check numbers for 256QAM
```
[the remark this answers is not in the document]
JoelPasvolsky: yes, that has not yet been verified

dimod/generators/mimo.py
```diff
@@ -0,0 +1,1085 @@
...
+    # Here, for BPSK F-induced complex parts of h and J are discarded:
+    # Given y = F x + nu, for independent and identically distributed channel F
+    # and complex noise nu, the system of equations defined by the real part is
+    # sufficient to define the canonical decoding problem.
+    # In essence, rotate y to the eigenbasis of F, throw away the orthogonal noise
+    # (the complex problem as the real part with suitable adjustment factor 2 to
+    # signal to noise ratio: F^{-1}*y = I*x + F^{-1}*nu)
+    # JR: revisit and prove
+
+    if modulation and modulation not in mod_config.keys():
```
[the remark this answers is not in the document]
JoelPasvolsky: Replacing `not in mod_config.keys()` with `not in mod_config` is good but removing `if modulation and` will raise an error when modulation is not specified: the default is currently None, but I can probably change that to default to BPSK, and then this is good

dimod/generators/mimo.py
JoelPasvolsky: I think I forgot to delete this after adding the `mod_config` superset of modulation properties.

dimod/generators/mimo.py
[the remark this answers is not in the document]
arcondello: Right, have an explicit modulation default makes sense to me.

dimod/generators/mimo.py
```diff
@@ -0,0 +1,1058 @@
...
+            i in range(transmitters_per_node)] for n in node_to_int}
+
+    return A, node_to_transmitters, node_to_receivers
+
+def create_channel(num_receivers: int = 1, 
+                   num_transmitters: int = 1,
+                   F_distribution: Optional[Tuple[str, str]] = None,
+                   random_state: Optional[Union[int, np.random.mtrand.RandomState]] = None,
+                   attenuation_matrix: Optional[np.ndarray] = None) -> Tuple[
+                       np.ndarray, float, np.random.mtrand.RandomState]:
```
arcondello, line 327: The `typing` claims this returns a 3-tuple, but the docs and code return a 2-tuple.

dimod/generators/mimo.py
```diff
@@ -0,0 +1,1058 @@
...
+            ...     F_distribution = ('binary', 'real'))
+
+    """
+
+    if not hasattr(lattice, 'edges') or not hasattr(lattice, 'nodes'): # not nx.Graph:
+        raise ValueError('Lattice must be a :class:`networkx.Graph`')
+
+    if modulation is None:
+        modulation = 'BPSK'
+
```
arcondello, line 937: Probably should just set the default `modulation` to `"BPSK"` so it's obvious that it's optional.

dimod/generators/mimo.py
```diff
@@ -0,0 +1,1058 @@
...
+    if transmitted_symbols is not None:
+        if modulation == 'BPSK' and np.iscomplex(transmitted_symbols).any():
+            raise ValueError(f"BPSK transmitted signals must be real")
+        if modulation != 'BPSK' and np.isreal(transmitted_symbols).any():
+            raise ValueError(f"Quadrature transmitted signals must be complex")
+    else:
+        if type(random_state) is not np.random.mtrand.RandomState:
+            random_state = np.random.RandomState(random_state)
+
+        quadrature = False if modulation == 'BPSK' else True
```
randomir, line 520:
```suggestion
        quadrature = modulation != 'BPSK'
```

dimod/generators/mimo.py
```diff
@@ -0,0 +1,1058 @@
...
+    h, J, offset = _yF_to_hJ(y, F, modulation)
+
+    if use_offset:
+        #NB - in this form, offset arises from
+        return dimod.BQM(h[:, 0], J, 'SPIN', offset=offset)
+    else:
+        np.fill_diagonal(J, 0)
+        return dimod.BQM(h[:, 0], J, 'SPIN')
+
+def spin_encoded_comp(lattice: GraphLike,
```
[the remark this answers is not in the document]
JoelPasvolsky: Now I remember why I used ``GraphLike`` rather than ``networkx.Graph``: we don't want to make Networkx a dependency of dimod:

```
 def spin_encoded_comp(lattice: networkx.Graph,
NameError: name 'networkx' is not defined
```

It's a bit more expansive but I think it's the closest we can do here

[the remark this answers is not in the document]
randomir: You can also just say:
```
def spin_encoded_comp(lattice: 'networkx.Graph',
```

But the problem with using `GraphLike` here is that we don't actually support it -- i.e. you can specify a lattice by providing an e.g. 5 -- which would mean `K5` as `GraphLike`.

dimod/generators/mimo.py
```diff
@@ -0,0 +1,1085 @@
...
+import networkx as nx
+import numpy as np
+from typing import Optional, Tuple, Union
+
+import dimod
+
+__all__ = ['spin_encoded_mimo', 'spin_encoded_comp', ]
+
+mod_config = { # bits per transmitter, amplitudes, transmitters_per_spin, number of amps, spins per symbol
+    "BPSK": {"bpt": 1, "amps": 1, "tps": 1, "na": 1, "sps": 1},
```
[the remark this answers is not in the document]
JoelPasvolsky: Done

dimod/generators/mimo.py
JoelPasvolsky: Removed

dimod/generators/mimo.py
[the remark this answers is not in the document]
arcondello: Why not https://github.com/dwavesystems/dimod/blob/8433f221a1e79101e1db0d80968ab5a2f59b865d/dimod/typing.py#L76 ?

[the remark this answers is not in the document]
randomir: @arcondello, because that function doesn't actually accept `GraphLike`, [but only `nx.Graph`](https://github.com/dwavesystems/dimod/blob/40ae9c2406319133e09007c61eb05167354550aa/dimod/generators/mimo.py#L933).

dimod/generators/mimo.py
JoelPasvolsky: I renamed to `comp` and `mimo`, which required renaming the module, which is more in keeping with other generator names (encoding as a BQM is implied).

dimod/generators/wireless.py
```diff
@@ -0,0 +1,985 @@
+# Copyright 2023 D-Wave Systems Inc.
```
arcondello, line 1: I think I liked this file being named `mimo` more. But `wireless` is also fine :shrug:

dimod/generators/wireless.py
```diff
@@ -0,0 +1,985 @@
...
+from collections import namedtuple
+from functools import wraps
+from itertools import product
+from typing import Literal, Optional, Tuple, Union
+
+import numpy as np
+
+import dimod
+
+__all__ = ['mimo', 'comp', ]
```
arcondello, line 24: `mimo` is OK, but `comp` is quite overloaded. [The wikipedia page](https://en.wikipedia.org/wiki/Cooperative_MIMO) uses `CO-MIMO` so IMO we should use `co_mimo` or `cooperative_mimo` here. I prefer the latter personally.

Alternatively we could spell both out, but I agree that `multiple_input_and_multiple_output` is quite long.

dimod/generators/wireless.py
```diff
@@ -0,0 +1,985 @@
...
+
+mod_params = namedtuple("mod_params", ["bits_per_transmitter", 
+                                       "amplitudes", 
+                                       "transmitters_per_spin", 
+                                       "number_of_amps", 
+                                       "spins_per_symbol"])
+mod_config = {
+    "BPSK": mod_params(1, 1, 1, 1, 1),
```
arcondello, line 32: If we want to be even more explicit we could do
```suggestion
    "BPSK": mod_params(bits_per_transmitter=1, amplitudes=1, transmitters_per_spin=1, number_of_amps=1, spins_per_symbol=1),
```
but that might be getting long

dimod/generators/wireless.py
```diff
@@ -0,0 +1,985 @@
...
+        matrix.
+    """
+
+    if len(F.shape) != 2 or F.shape[0] != y.shape[0]:
+        raise ValueError("F should have shape (n, m) for some m, n "
+                         "and n should equal y.shape[1];"
+                         f" given: {F.shape}, n={y.shape[1]}")
+
+    offset = np.matmul(y.imag.T, y.imag) + np.matmul(y.real.T, y.real)
+    h = -2 * np.matmul(F.T.conj(), y)    # Be careful with interpretation!
```
arcondello, line 63: Same comment as before, it's not at all obvious how I should be careful here. This comment should be expanded or removed.

dimod/generators/wireless.py
```diff
@@ -0,0 +1,985 @@
...
+        raise ValueError(f"Unsupported modulation: {modulation}")
+
+    if modulation != 'BPSK' and (np.iscomplex(h).any() or np.iscomplex(J).any()):
+        hR = np.concatenate((h.real, h.imag), axis=0)
+        JR = np.concatenate((np.concatenate((J.real, J.imag), axis=0),
+                             np.concatenate((J.imag.T, J.real), axis=0)),
+                             axis=1)
+        return hR, JR
+    else:
+        return h.real, J.real
```
arcondello, line 107: Very much a nit but....
```suggestion
    return h.real, J.real
```
I am reasonably convinced by the [LLVM Style Guide](https://llvm.org/docs/CodingStandards.html#don-t-use-else-after-a-return)'s arguments on this sort of flow.

dimod/generators/wireless.py
```diff
@@ -0,0 +1,985 @@
...
+
+    if np.iscomplex(amps).any():
+        raise ValueError('Amplitudes cannot have complex values')
+
+    if any(np.modf(amps)[0]):
+        raise ValueError('Amplitudes must have integer values')
+
+    random_state = np.random.default_rng(random_state)
+
+    if quadrature == False:
```
arcondello, line 438:
```suggestion
    if not quadrature:
```

dimod/generators/wireless.py
```diff
@@ -0,0 +1,985 @@
...
+                      F: Union[np.array, None] = None,
+                      *,
+                      transmitted_symbols: Union[np.array, None] = None,
+                      channel_noise: Union[np.array, None] = None,
+                      num_transmitters: int = None,
+                      num_receivers: int = None,
+                      SNRb: float = float('Inf'),
+                      seed: Union[None, int, np.random.RandomState] = None,
+                      F_distribution: Union[None, tuple] = None,
+                      attenuation_matrix = None) -> dimod.BinaryQuadraticModel:
```
arcondello, line 555: weird indentation

dimod/generators/wireless.py
```diff
@@ -0,0 +1,985 @@
...
+def comp(lattice: 'networkx.Graph',
+                      modulation: Literal["BPSK", "QPSK", "16QAM", "64QAM", "256QAM"] = "BPSK", 
+                      y: Optional[np.array] = None,
+                      F: Union[np.array, None] = None,
+                      *,
+                      transmitted_symbols: Union[np.array, None] = None,
+                      channel_noise: Union[np.array, None] = None,
+                      SNRb: float = float('Inf'),
+                      seed: Union[None, int, np.random.RandomState] = None,
+                      F_distribution: Union[None, str] = None) -> dimod.BinaryQuadraticModel:
```
arcondello, line 760: weird indentation

dimod/generators/wireless.py
```diff
@@ -0,0 +1,985 @@
...
+    if method == 'matched_filter':  # F = root(Nt/P) Fcompconj
+        return F.conj().T / np.sqrt(PoverNt)
+
+    # method == 'MMSE':
+    return np.matmul(
+        F.conj().T,
+        np.linalg.pinv(np.matmul(F, F.conj().T) + np.identity(Nr)/SNRoverNt)
+            ) / np.sqrt(PoverNt)
+
+def filter_marginal_estimator(x: np.array, modulation: str):
```
arcondello, line 961: Does this one need a docstring?

dimod/generators/wireless.py
```diff
@@ -0,0 +1,985 @@
...
+        x_R = np.where(x_R < -max_abs, -max_abs, x_R)
+        x_R = np.where(x_R > max_abs, max_abs, x_R)
+
+        if modulation != 'BPSK':
+            x_I = 2*np.round((x.imag - 1)/2) + 1
+            x_I = np.where(x_I <- max_abs, -max_abs, x_I)
+            x_I = np.where(x_I > max_abs, max_abs, x_I)
+            return x_R + 1j*x_I
+        else:
+            return x_R
```
arcondello, line 985:
```suggestion
        return x_R
```

setup.py
```diff
@@ -81,6 +81,7 @@ def finalize_options(self):
     install_requires=[
         # this is the oldest supported NumPy on Python 3.8
         'numpy>=1.17.3,<2.0.0',
+        'networkx',
```
arcondello, line 84: Do we still require nextworkx?

dimod/generators/wireless.py
jackraymond: Remove it

dimod/generators/wireless.py
JoelPasvolsky: It would make these next five lines much longer and having the named parameters just above the value-setting lines should be clear enough

dimod/generators/wireless.py
JoelPasvolsky: Removed

dimod/generators/wireless.py
JoelPasvolsky: Fixed, thanks!

dimod/generators/wireless.py
JoelPasvolsky: fixed, thanks!

setup.py
JoelPasvolsky: Removed

dimod/generators/wireless.py
jackraymond: """Takes the continuous filter output and maps each symbol estimate to the nearest constellation value (valid symbol)"""

dimod/generators/wireless.py
JoelPasvolsky: Okay, so just to clarify, ours is [Coordinated multipoint](https://en.wikipedia.org/wiki/Cooperative_MIMO) on that same page. I [don't see](https://en.wikipedia.org/wiki/Comp) that `comp` is overloaded, and would prefer to keep it, but if there is objection, are we okay with `mimo` and `coordinated_multipoint` to avoid the long  `multiple_input_and_multiple_output`? (@jackraymond, @arcondello )

dimod/generators/mimo.py
jackraymond: No strong preference.
The natural statement of the mimo problem is a quadratic function on constellation states. This is how the problem would be presented to the hybrid solver (for example). In light of Alex's observation I'd suggest bqm_encoded_mimo, but if thats not standard I dont mind. Note that there are multiple ways to encode (gray coding etc), but the simplest form is used here.

dimod/generators/wireless.py
jackraymond: No strong opinion, docstring makes sense to me, no name is perfect.

arcondello: I mean overloaded in the sense of when I see `comp` out of context, I think "comparison" or "compilation"

dimod/generators/wireless.py
JoelPasvolsky: Updated, thanks!

dimod/generators/wireless.py
JoelPasvolsky: Okay, closed in [this commit](https://github.com/dwavesystems/dimod/pull/1357/commits/b00e943aad189ff1590d225877611c06f071d0cd)

dimod/generators/mimo.py
JoelPasvolsky: Closed as part of [this discussion](https://github.com/dwavesystems/dimod/pull/1357#discussion_r1310526351)

releasenotes/notes/feature-mimo-59c3d6d02335061c.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    Add ``mimo()`` function for generating a multi-input 
+    multiple-output (MIMO) channel-decoding problem. 
+  - |  
+    Add ``comp()`` function for generating a coordinated multi-point
```
arcondello, line 7: Need to update the release note to match the new function name.

JoelPasvolsky: I felt I was missing something. Maybe this next commit will trigger CI, which seems to have missed [the previous commit](https://github.com/dwavesystems/dimod/pull/1357/commits/b00e943aad189ff1590d225877611c06f071d0cd)

JoelPasvolsky, line 7:
```suggestion
    Add ``coordinated_multipoint()`` function for generating a coordinated multi-point 
```

JoelPasvolsky: Updated with [this commit](https://github.com/dwavesystems/dimod/pull/1357/commits/7edb35e3197ea3d1a5ba5d25ba6aeb0a3f10b7bf)

dimod/generators/wireless.py
```diff
@@ -0,0 +1,988 @@
...
+    return A, node_to_transmitters, node_to_receivers
+
+def create_channel(num_receivers: int = 1, 
+                   num_transmitters: int = 1,
+                   F_distribution: Optional[Tuple[str, str]] = None,
+                   random_state: Optional[Union[int, np.random.mtrand.RandomState]] = None,
+                   attenuation_matrix: Optional[np.ndarray] = None) -> Tuple[np.ndarray, float]:
+    """Create a channel model.
+
+    Channel power is the expected root-mean-square signal per receiver (i.e.,
```
jackraymond, line 325:
```suggestion
    Channel power is the expected mean-square signal amplification per receiver (i.e.,
```

dimod/generators/wireless.py
```diff
@@ -0,0 +1,988 @@
...
+    :math:`mean(F^2)*num_transmitters`) for homogeneous codes.
```
jackraymond, line 326: Delete this line, it becomes very complicated in the case of a non-defaulted attenuation matrix.

dimod/generators/wireless.py
```diff
@@ -0,0 +1,988 @@
...
+    :math:`N_0 I[Nt] = SNRb `.
+
+    Args:
+        F: Wireless channel as an :math:`i \times j` matrix of complex values,
+            where :math:`i` rows correspond to :math:`y_i` receivers and :math:`j`
+            columns correspond to :math:`v_i` transmitted symbols.
+
+        transmitted_symbols: Transmitted symbols as a column vector.
+
+        channel_noise: Channel noise as a complex value.
```
jackraymond, line 474:
```suggestion
        channel_noise: Channel noise as a column vector.
```

dimod/generators/wireless.py
```diff
@@ -0,0 +1,988 @@
...
+
+        SNRb: Signal-to-noise ratio.
+
+        modulation: Modulation. Supported values are 'BPSK', 'QPSK', '16QAM',
+            '64QAM', and '256QAM'.
+
+        channel_power: Channel power. By default, proportional to the number
```
jackraymond, line 481:
```suggestion
        channel_power: Channel power. By default, equal to the number
```

dimod/generators/wireless.py
```diff
@@ -0,0 +1,988 @@
...
+            generate the received signal, :math:`y`. The number of transmitted 
+            symbols must be consistent with ``F``.
+            
+            For BPSK and QPSK modulations, statistics of the ensemble do not 
+            depend on the choice: all choices are equivalent. By default, 
+            symbols are chosen for all users as :math:`1` or :math:`1 + 1j`, 
+            respectively. Note that for correct analysis by some solvers, applying 
+            spin-reversal transforms may be necessary.
+            
+            For QAM modulations, amplitude randomness affects likelihood in a
```
jackraymond, line 644: Sometimes QPSK is referred to as 4QAM, and this makes sense since they are just conventions. However, statements like this that relate to QAM but not QPSK then risk misleading. not sure of the simplest solution though.

jackraymond (commented):
Looks good to me. Really needs a text book accompaniment

dimod/generators/wireless.py
JoelPasvolsky: I updated to "For QAM modulations such as 16QAM, amplitude randomness affects..."

dimod/generators/wireless.py
```diff
@@ -0,0 +1,988 @@
...
+
+import numpy as np
+
+import dimod
+
+__all__ = ['mimo', 'coordinated_multipoint', ]
+
+mod_params = namedtuple("mod_params", ["bits_per_transmitter", 
+                                       "amplitudes", 
+                                       "transmitters_per_spin",
```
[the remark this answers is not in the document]
JoelPasvolsky: Standardized to bit in [this commit](https://github.com/dwavesystems/dimod/pull/1357/commits/61508e494d86b71c8ce1e8a258b5639b4e1e64fc)

arcondello (approved):
LGTM

dimod/generators/wireless.py
jackraymond: In context of this function, where QPSK is clearly distinguished from QAM there is no issue anyway.

jackraymond (approved):

PR #1363, opened 2024-01-15
Don't include debug info in .so files

arcondello:


setup.py
```diff
@@ -28,7 +28,7 @@
 
 extra_link_args = {
     'msvc': [],
-    'unix': ['-std=c++11'],
+    'unix': ['-std=c++11', '-g0'],
```
randomir, line 31: Have you tried with `-g1`?

randomir (approved):
Nice reduction from 19MB to 9MB.

Issue #1364, opened 2024-02-09
Consider adding a kwarg to skip checking whether a discrete constraint is valid


arcondello:
When one calls `CQM.add_discrete()` we check that the given constraint is a valid discrete constraint - that is that it doesn't overlap with any other discrete constraint. The check happens in one of two places: https://github.com/dwavesystems/dimod/blob/e7bc4d5f570482c51b199dee28db849f24fd27e1/dimod/constrained/constrained.py#L607-L613 or https://github.com/dwavesystems/dimod/blob/e7bc4d5f570482c51b199dee28db849f24fd27e1/dimod/constrained/constrained.py#L671-L683

This is quite a slow check, and many users therefore bypass it by adding the discrete constraint as a constraint and then marking it as discrete.
```python
label = cqm.add_constraint(x + y + z == 1)
cqm.constraints[label].lhs.mark_discrete()
```

This bypasses the check. Luckily, [dwave-preprocessing](https://github.com/dwavesystems/dwave-preprocessing) handles fixing overlapping discrete constraints, so I think it's fairly safe to make the bypass more "legal", albeit with caveats explained clearly in the docs.

Therefore, we should/could consider adding a keyword argument `skip_check` defaulting to `False` that would allow users to bypass the check. If they then create overlapping discrete constraints they will be removed by presolve later.
```
cqm.add_discrete(["a", "b", "c"], skip_check=True)
```
For experience users who promise to construct the model without overlapping discrete constraints by default this is likely to be useful.

PR #1365, opened 2024-02-19
Addressing issue #1364: option for a variable check in add_discrete method

aivarsoo:
As discussed in the issue #1364, the variable check in `ConstrainedQuadraticModel.add_discrete` method  can be skipped in some cases and the user can decide if this is truly necessary. 

This is implemented by adding a `do_var_check` boolean variable (default: `True`), which enables the check if it's true, in the methods `ConstrainedQuadraticModel.add_discrete_from_model`, `ConstrainedQuadraticModel.add_discrete_from_comparison`,  and `ConstrainedQuadraticModel.add_discrete_from_iterable`.

### DIFF
--- a/dimod/constrained/constrained.py
+++ b/dimod/constrained/constrained.py
@@ -494,13 +494,7 @@ class ConstrainedQuadraticModel(cyConstrainedQuadraticModel):
 
             >>> d, e, f = dimod.Binaries(['d', 'e', 'f'])
             >>> cqm.add_discrete(d + e + f == 1, label='discrete-def')
-            'discrete-def'
-            
-            Add a discrete constraint, while skipping the variable overlap
-            check.
-            
-            >>> cqm.add_discrete(d + e + f == 1, label='discrete-def',
-            ...                  check_overlaps=False)
+            'discrete-def'            
 
         """
         # in python 3.8+ we can use singledispatchmethod
--- a/dimod/constrained/constrained.py
+++ b/dimod/constrained/constrained.py
@@ -510,7 +510,8 @@ class ConstrainedQuadraticModel(cyConstrainedQuadraticModel):
     def add_discrete_from_comparison(self,
                                      comp: Comparison,
                                      label: Optional[Hashable] = None,
-                                     copy: bool = True) -> Hashable:
+                                     copy: bool = True,
+                                     do_var_check: bool = True) -> Hashable:
         """Add a one-hot constraint from a comparison.
 
         One-hot constraints can represent discrete variables (for example a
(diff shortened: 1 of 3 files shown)


dimod/constrained/constrained.py
```diff
@@ -510,7 +510,8 @@ def add_discrete(self, data, *args, **kwargs) -> Hashable:
     def add_discrete_from_comparison(self,
                                      comp: Comparison,
                                      label: Optional[Hashable] = None,
-                                     copy: bool = True) -> Hashable:
+                                     copy: bool = True,
+                                     do_var_check: bool = True) -> Hashable:
```
arcondello, line 514: I think `check_valid` or `check_overlaps` would be a better name, if we want to keep the "direction" of having it default to `True`.

arcondello (commented):
This does need a [release note](https://github.com/dwavesystems/dimod?tab=readme-ov-file#release-notes). Probably something like
```
Add <kwarg name> keyword argument to `ConstrainedQuadraticModel.add_discrete()`, see `#1364 <https://github.com/dwavesystems/dimod/issues/1364>`_.
```
Also, ideally we would add tests to [TestAddDiscrete](https://github.com/dwavesystems/dimod/blob/e7bc4d5f570482c51b199dee28db849f24fd27e1/tests/test_constrained.py#L168)
```python
def test_overlap_check(self):
    cqm.add_discrete("abc")

        cqm.add_discrete("bcd", check_overlaps=True)
    cqm.add_discrete("bcd", check_overlap=False)  # overlapping
```

dimod/constrained/constrained.py
```diff
@@ -539,6 +540,9 @@ def add_discrete_from_comparison(self,
                 set to `False` to improve performance, but subsequently mutating
                 the model can cause issues.
 
+            do_var_check: If `True` we perform a variable check. In particular, if 
+                the variables already exist, we make sure they're not used.
```
arcondello, line 544:
```suggestion
                the variables already exist, we make sure they're not used in another discrete constraint.
```
here and elsewhere.

dimod/constrained/constrained.py
```diff
@@ -669,7 +682,7 @@ def add_discrete_from_model(self,
             raise ValueError("discrete constraints must be linear")
 
         for v, bias in qm.iter_linear():
-            if v in self.variables:
+            if do_var_check and v in self.variables:
                 # it already exists, let's make sure it's not already used
                 if any(v in self.constraints[label].lhs.variables for label in self.discrete):
```
arcondello, line 687:
```suggestion
            if v in self.variables:
                # it already exists, let's make sure it's not already used
                if do_var_check and any(v in self.constraints[label].lhs.variables for label in self.discrete):
```

dimod/constrained/constrained.py
aivarsoo: `check_overlaps` is a great name! I'll stick with that

arcondello:
FWIW, I think this keyword arg is sufficiently advanced that we don't need to add a docstring example for it.

arcondello (approved):

PR #1371, opened 2024-06-13
Add QuadraticModelBase::remove_interactions() method

arcondello:
This is designed to be used by `dwave-preprocessing` to speed up the removal of small biases.

### DIFF
--- a/dimod/include/dimod/abc.h
+++ b/dimod/include/dimod/abc.h
@@ -331,6 +331,12 @@ class QuadraticModelBase {
     /// Remove the interaction between variables `u` and `v`.
     bool remove_interaction(index_type u, index_type v);
 
+    /// Remove all interactions for which `filter` returns `true`.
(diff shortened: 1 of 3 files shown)


dimod/include/dimod/abc.h
```diff
@@ -889,6 +895,33 @@ bool QuadraticModelBase<bias_type, index_type>::remove_interaction(index_type u,
...
+                                     return filter(u, v, bias);
+                                 });
+
+        num_removed += n.end() - it;
+
+        n.erase(it, n.end());
+
+        u += 1;
+    }
+
```
wbernoudy, line 921: Maybe add an 
```assert(num_removed % 2 == 0);```
?

wbernoudy (approved):

dimod/include/dimod/abc.h
arcondello: done

PR #1373, opened 2024-06-24
Remove build-time dependency on NumPy

arcondello:
This will make NumPy 2.0 support much easier. Will do that in a followup PR.

Contains the commits from https://github.com/dwavesystems/dimod/pull/1374. I will rebase once that one is merged.

PR #1374, opened 2024-06-24
Use M1 resource class for macos builds on CircleCI

arcondello:
I'll uncap the NumPy version in another PR (i.e., https://github.com/dwavesystems/dimod/pull/1373).

See also https://github.com/dwavesystems/dwave-ocean-sdk/issues/294.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -267,7 +267,7 @@ jobs:
             python -m venv env
             . env/bin/activate
             pip install dimod --no-index -f dist/ --force-reinstall --no-deps
-            pip install << parameters.dependencies >> --upgrade --only-binary=numpy
+            pip install "<< parameters.dependencies >>" --upgrade --only-binary=numpy
       - run: &unix-run-tests
           name: run tests
           command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -259,7 +267,7 @@ jobs:
             python -m venv env
             . env/bin/activate
             pip install dimod --no-index -f dist/ --force-reinstall --no-deps
-            pip install << parameters.dependencies >> --upgrade --only-binary=numpy
+            pip install "<< parameters.dependencies >>" --upgrade --only-binary=numpy
```
randomir, line 270: I recommend quoting individual specifiers -- below in the list of parameter values -- not the parameter value here. That way you can use multiple specifiers per parameter (e.g. `"'numpy<2' scipy~=1.0"`). For example, see cloud-client's config, or [dwave-inspector's](https://github.com/dwavesystems/dwave-inspector/blob/master/.circleci/config.yml) one.

Alternatively, you can use tilde in version specifier, and avoid quoting altogether.

.circleci/config.yml
```diff
@@ -3,6 +3,7 @@ version: 2.1
 orbs:
   codecov: codecov/codecov@3
   win: circleci/windows@5.0.0
+  macos: circleci/macos@2.4.1
```
randomir, line 6: It's probably fine to reduce precision to minor or major version here, that way the latest is used.

randomir (approved):
lgtm

.circleci/config.yml
arcondello: I will likely revert this commit as part of the NumPy 2.0 PR, so I will leave it be for now.

PR #1375, opened 2024-06-25
Add support for Numpy 2.0

arcondello:


arcondello:
Is missing a release note. Will add.

dimod/binary/binary_quadratic_model.py
```diff
@@ -292,7 +292,7 @@ def __len__(self):
         return self.num_variables
 
     def __repr__(self):
-        return "{!s}({!s}, {!s}, {!r}, {!r})".format(type(self).__name__,
+        return "{!s}({!s}, {!s}, {}, {!r})".format(type(self).__name__,
```
randomir, line 295: fix indent of next line

dimod/sampleset.py
```diff
@@ -1534,7 +1539,7 @@ def aggregate(self):
         # fix the number of occurrences
         record.num_occurrences = 0
         for old_idx, new_idx in enumerate(inverse):
-            record[new_idx].num_occurrences += self.record[old_idx].num_occurrences
+            record["num_occurrences"][new_idx] += self.record[old_idx].num_occurrences
```
randomir, line 1542: I'm curious, what's the reason for this change?

randomir (approved):

dimod/sampleset.py
arcondello: The old one doesn't work anymore. I think it may be a bug in NumPy? I couldn't find a release note or bug report though.

PR #1372, opened 2024-06-21
Remove ambiguous xrefs to reduce build warnings

JoelPasvolsky:
Part of [sdk](https://github.com/dwavesystems/dwave-ocean-sdk/pull/300), [dwave-system](https://github.com/dwavesystems/dwave-system/pull/524), [dwave-cloud-client](https://github.com/dwavesystems/dwave-cloud-client/pull/633), and [dwave-hybrid](https://github.com/dwavesystems/dwave-hybrid/pull/294) PRs to reduce build warnings

The updates in this dimod PR reduce the number of warnings by half (<70 left).  

Additional minor update: Pyments does not have an `ipythonconsole` lexer so dimod was producing `Could not lex literal_block as "python"` warnings so switched to the standard Python console for those examples.

PR #1378, opened 2024-06-28
Fix lexar warning on ipython prompt

JoelPasvolsky:
[This commit](https://github.com/dwavesystems/dimod/pull/1372/commits/4f3dcf3c471a5d1871d90b302a576a40fba80a2d) fixed lexar warnings on unsupported `ipythonconsole` but unfortunately [this commit](https://github.com/dwavesystems/dimod/pull/1372/commits/80fbae59f3608ce86e48d169a1577932de8600ba), in response to [this review comment](https://github.com/dwavesystems/dimod/pull/1372#discussion_r1653270133), introduced a new lexar problem, `WARNING: Could not lex literal_block as "python"`. 

This PR fixes these new warnings.

### DIFF
--- a/docs/intro/intro_scaling.rst
+++ b/docs/intro/intro_scaling.rst
@@ -97,7 +97,7 @@ Time the construction:
 
     bin_packing(weights)
 
-.. code-block:: python
+.. code-block:: text
 
     In [1]: %timeit bin_packing(weights)
     385 ms ± 9.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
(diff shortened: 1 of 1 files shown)


docs/intro/intro_scaling.rst
```diff
@@ -97,7 +97,7 @@ Time the construction:
 
     bin_packing(weights)
 
-.. code-block:: python
+.. code-block:: text
```
randomir, line 100: Isn't `text` default? I.e. no need to specify it explicitly.

Also, isn't this removing syntax coloring?

JoelPasvolsky: Yes and yes. But, I think it worth explicitly specifying `text` so anyone in the file in future sees that it's deliberate and for such a small snippet, lack of coloring is no catastrophe:

image

An alternative solution is to create a custom lexar for ipython. Seems overkill for these two-line snippets.

PR #1379, opened 2024-07-15
add power-r generator

pau557:
Weights follow a power-law distribution with respect to their magnitude. Good for high precision instances.

### DIFF
--- a/dimod/generators/random.py
+++ b/dimod/generators/random.py
@@ -15,6 +15,8 @@ import warnings
 
 import collections.abc as abc
 
+import numbers
+
 from typing import Callable, Optional, Sequence, Union
 
 import numpy as np
@@ -505,7 +507,7 @@ def doped(p: float, graph: GraphLike,
 
 @graph_argument('graph')
 def power_r(r: int, graph: GraphLike,
-          seed: Optional[int] = None) -> BinaryQuadraticModel:
+          seed: Union[None, int, np.random.Generator] = None,) -> BinaryQuadraticModel:
     """Generate an Ising model for a POWERr problem.
 
     In POWERr problems all linear biases are zero and quadratic values are integers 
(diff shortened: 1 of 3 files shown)


dimod/generators/random.py
```diff
@@ -500,3 +500,61 @@ def doped(p: float, graph: GraphLike,
...
+            NetworkX graph.
+
+        seed: Random seed.
+
+    Returns:
+        A binary quadratic model.
+
+    """
+
+    if not isinstance(r, int):
```
arcondello, line 533:
```
    if not isinstance(r, numbers.Integral):
```
to support integer-like types (like `np.int8`).

dimod/generators/random.py
```diff
@@ -500,3 +500,61 @@ def doped(p: float, graph: GraphLike,
...
     return bqm
+
+
+
+@graph_argument('graph')
+def power_r(r: int, graph: GraphLike,
+          seed: Optional[int] = None) -> BinaryQuadraticModel:
+    """Generate an Ising model for a POWERr problem.
+
+    In POWERr problems all linear biases are zero and quadratic values are integers
```
arcondello, line 511: Is there a citation for this problem class?

arcondello (commented):
A few minor nits but otherwise looking good!

arcondello:
Oh, also needs a [release note](https://github.com/dwavesystems/dimod?tab=readme-ov-file#release-notes). Probably something like `Add ``power_r()`` generator function to ``dimod.generators``. The weights of generated Ising problems follow a power-law distribution with respect to their magnitude.`

dimod/generators/random.py
pau557: not yet!

pau557:
> Oh, also needs a [release note](https://github.com/dwavesystems/dimod?tab=readme-ov-file#release-notes). Probably something like `` Add ``power_r()`` generator function to ``dimod.generators``. The weights of the generated Ising problems follow a power-law distribution with respect to their magnitude. ``

pau557:
Getting an error with `reno`. Will try again tomorrow

dimod/generators/random.py
```diff
@@ -500,3 +502,63 @@ def doped(p: float, graph: GraphLike,
         bqm.add_interaction(u, v, J)
 
     return bqm
+
+
+
+@graph_argument('graph')
+def power_r(r: int, graph: GraphLike,
+          seed: Union[None, int, np.random.Generator] = None,) -> BinaryQuadraticModel:
```
arcondello, line 510: Careful with the indentation. We should also make `seed` a keyword only argument.

```suggestion
def power_r(r: int, 
            graph: GraphLike,
            *,
            seed: Union[None, int, np.random.Generator] = None,
            ) -> BinaryQuadraticModel:
```

pau557: is it ok to use the formatting that Black suggests instead?

```python
@graph_argument("graph")
def power_r(
    r: int,
    graph: GraphLike,
    *,
    seed: Union[None, int, np.random.Generator] = None,
) -> BinaryQuadraticModel:


```

arcondello: Hah, technically a pep8 violation (https://github.com/psf/black/issues/1178 for the rabbit hole), but sure, that's OK.

arcondello (approved):

PR #1380, opened 2024-07-19
create qap generator

eordog:


dimod/generators/qap.py
arcondello, line 1: I think naming this file `quadratic_assignment.py` would be a bit more explicit.

dimod/generators/qap.py
```diff
@@ -0,0 +1,69 @@
...
+
+        flows: numpy array containing the flow between facility i and facility j
+
+    Returns:
+        The constrained quadratic model encoding the quadratic assignment problem. Variables are
+        denoted as ``x_{i}_{j}`` where ``x_{i}_{j} == 1`` means that facility ``i`` is
+        placed in location ``j``.
+
+    """
+
```
arcondello, line 43: Because we're accepting `ArrayLike`, we should add something like
```python
    distance_matrix = np.atleast_2d(np.asarray(distance_matrix))
    flow_matrix = np.atleast_2d(np.asarray(flow_matrix))
```
so that we also accept `list`s etc.

dimod/generators/qap.py
```diff
@@ -0,0 +1,69 @@
...
+    if distance_matrix.shape != flow_matrix.shape:
```
arcondello, line 44: We should also test that they are square, and 2d.

dimod/generators/qap.py
```diff
@@ -0,0 +1,69 @@
...
+            obj.set_quadratic(x[(i,j)], x[(k,l)], flow_matrix[i][k]*distance_matrix[j][l]
+                              + flow_matrix[k][i]*distance_matrix[j][l])
+
+    model.set_objective(obj)
+
+    for i in range(num_locations):
+        constraint = [(x[(i,j)], 1) for j in range(num_locations)] + [(-1,)]
+        model.add_constraint(constraint, sense='==')
+
+    for j in range(num_locations):
```
arcondello, line 65: We could/should add one of these as a [discrete constraint](https://docs.ocean.dwavesys.com/en/stable/docs_dimod/reference/generated/dimod.ConstrainedQuadraticModel.add_discrete.html).

arcondello (commented):
We also need tests and a release note. I think https://github.com/dwavesystems/dimod/pull/1379 is a pretty good example of what we'd be looking for. Let me know if you'd like any help!

dimod/generators/qap.py
```diff
@@ -0,0 +1,69 @@
...
+
+from dimod.binary_quadratic_model import BinaryQuadraticModel
+from dimod.constrained import ConstrainedQuadraticModel
+
+__all__ = ['quadratic_assignment']
+
+def quadratic_assignment(distance_matrix: np.typing.ArrayLike,
+                         flow_matrix: np.typing.ArrayLike,
+                         ) -> ConstrainedQuadraticModel:
+    """Generates a constrained quadratic model encoding a quadratic assignment problem.
```
JoelPasvolsky, line 27:
```suggestion
    """Generates a constrained quadratic model encoding a quadratic-assignment problem.
```

dimod/generators/qap.py
```diff
@@ -0,0 +1,69 @@
...
+
+    Given distance and flow matrices, generates a :class:`~dimod.ConstrainedQuadraticModel` 
+    for the corresponding quadratic assignment problem.
```
JoelPasvolsky, line 30:
```suggestion
    for the corresponding quadratic-assignment problem. 
```

dimod/generators/qap.py
```diff
@@ -0,0 +1,69 @@
...
+def quadratic_assignment(distance_matrix: np.typing.ArrayLike,
+                         flow_matrix: np.typing.ArrayLike,
+                         ) -> ConstrainedQuadraticModel:
+    """Generates a constrained quadratic model encoding a quadratic assignment problem.
+
+    Given distance and flow matrices, generates a :class:`~dimod.ConstrainedQuadraticModel` 
+    for the corresponding quadratic assignment problem. 
+
+    Args:
+        distances: numpy array containing the distance between location i and location j
```
JoelPasvolsky, line 33:
```suggestion
        distances: Distances between locations :math:`i` and :math:`j` as a NumPy array. 
```

dimod/generators/qap.py
```diff
@@ -0,0 +1,69 @@
...
+
+        flows: numpy array containing the flow between facility i and facility j
```
JoelPasvolsky, line 35:
```suggestion
        flows: flow between facilities :math:`i` and :math:`j` as a NumPy array.
```

dimod/generators/qap.py
```diff
@@ -0,0 +1,69 @@
...
+
+    Returns:
+        The constrained quadratic model encoding the quadratic assignment problem. Variables are
```
JoelPasvolsky, line 38:
```suggestion
        The constrained quadratic model encoding the quadratic-assignment problem. Variables are
```

JoelPasvolsky (commented):
Very nice! 
Please also update the [generators RST file](https://github.com/dwavesystems/dimod/blob/main/docs/reference/generators.rst)

I also made some small phrasing suggestions inline for your consideration.

dimod/generators/qap.py
eordog: Thank you! All suggestions are applied.

dimod/generators/qap.py
eordog: Thank you! All suggestions are applied.

arcondello (approved):
LGTM

Issue #1381, opened 2024-07-29
Consider filtering functionality in `ConstrainedQuadraticModel.iter_constraint_data()`


arcondello:
Currently [`ConstrainedQuadraticModel.iter_constraint_data()`](https://github.com/dwavesystems/dimod/blob/835688330a6979bb6eeab7d21bbc8f845f37acbe/dimod/constrained/constrained.py#L1147) always iterates over every constraint in the model.

We should consider adding some keyword arguments to make that a bit easier when handling constraints in bulk. I am imagining something like
```Python
     def iter_constraint_data(sample_like, *, filter_function=None):
        if callable(filter_function):
            yield from filter(filter_function, self.iter_constraint_data(sample_like))
            return

        ...
```
and/or
```Python
     def iter_constraint_data(sample_like, *, labels=None):
        ...

        if labels is None:
            labels = self.constraint_labels
            
        for label in labels:
            constraint = self.constraint(label)

            lhs = constraint.lhs.energy((sample, labels))
            rhs = constraint.rhs
            sense = constraint.sense
            ...
```

mirko-sof:
Hi @arcondello,

I am interested in this feature. I have some pre-processing and post-processing steps which start from a feasible solution and then make changes to it by changing the values of a few variables. Every time I make such a change I use the `iter_constraint_data` method to check if the change violates any constraints. When doing this it would save computation time if I could check only those constraints which involve the variables that are about to be changed.

The second approach that you outlined above (passing a `labels` argument) looks to me like it would be suitable for saving computation time. The first approach (passing a `filter_function`) looks like it still iterates over all the constraints because the `filter_function` needs to be applied to each constraint. Is my understanding correct?

arcondello:
> The second approach that you outlined above (passing a labels argument) looks to me like it would be suitable for saving computation time. The first approach (passing a filter_function) looks like it still iterates over all the constraints because the filter_function needs to be applied to each constraint. Is my understanding correct?

Yes exactly. If you know what specific constraints you're looking for by label, we can easily fetch those. But if you want to filter by some other property, or by say a regex of the label, we need to iterate over all constraints.

That said, there may be some performance benefit even in the latter case because you can bypass [the energy calculation](https://github.com/dwavesystems/dimod/blob/835688330a6979bb6eeab7d21bbc8f845f37acbe/dimod/constrained/constrained.py#L1195) which is where most of the time is spent.

[...]

PR #1382, opened 2024-08-02
Add labels keyword argument to CQM.iter_constraint_data()

arcondello:
See https://github.com/dwavesystems/dimod/issues/1381. This addresses one of the options laid out.

### DIFF
--- a/dimod/constrained/constrained.py
+++ b/dimod/constrained/constrained.py
@@ -1144,7 +1144,11 @@ class ConstrainedQuadraticModel(cyConstrainedQuadraticModel):
 
         return cqm
 
-    def iter_constraint_data(self, sample_like: SamplesLike) -> Iterator[ConstraintData]:
+    def iter_constraint_data(self,
+                             sample_like: SamplesLike,
+                             *,
+                             labels: typing.Optional[typing.Iterable[typing.Hashable]] = None,
+                             ) -> Iterator[ConstraintData]:
         """Yield information about the constraints for the given sample.
 
         Note that this method iterates over constraints in the same order as
@@ -1153,6 +1157,7 @@ class ConstrainedQuadraticModel(cyConstrainedQuadraticModel):
         Args:
             sample_like: A sample. `sample-like` is an extension of
                 NumPy's array_like structure. See :func:`.as_samples`.
+            labels: A subset of the constraint labels over which to iterate.
 
         Yields:
             A :class:`collections.namedtuple` with the following fields.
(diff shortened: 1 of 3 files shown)


dimod/constrained/constrained.py
```diff
@@ -1144,7 +1144,11 @@ def from_file(cls,
 
         return cqm
 
-    def iter_constraint_data(self, sample_like: SamplesLike) -> Iterator[ConstraintData]:
+    def iter_constraint_data(self,
+                             sample_like: SamplesLike,
+                             *,
+                             labels: typing.Optional[typing.Iterable[typing.Hashable]] = None,
```
thisac, line 1150: The types seem to already be imported.
```suggestion
                             labels: Optional[Iterable[Hashable]] = None,
```

dimod/constrained/constrained.py
```diff
@@ -1219,6 +1233,7 @@ def iter_constraint_data(self, sample_like: SamplesLike) -> Iterator[ConstraintD
     def iter_violations(self, sample_like: SamplesLike, *,
                         skip_satisfied: bool = False,
                         clip: bool = False,
+                        labels: typing.Optional[typing.Iterable[typing.Hashable]] = None,
```
thisac, line 1236:
```suggestion
                        labels: Optional[Iterable[Hashable]] = None,
```

dimod/constrained/constrained.py
```diff
@@ -1153,6 +1157,7 @@ def iter_constraint_data(self, sample_like: SamplesLike) -> Iterator[ConstraintD
         Args:
             sample_like: A sample. `sample-like` is an extension of
                 NumPy's array_like structure. See :func:`.as_samples`.
+            labels: A subset of the constraint labels over which to iterate.
```
thisac, line 1160: Might be worth mentioning that the constraints will be yielded in the same ordering as the labels passed, instead of in order of addition which I presume is the case without a labels subset request.

thisac (approved):
LGTM

dimod/constrained/constrained.py
arcondello: I am being deliberately ambiguous in case we want to change the behavior later.

dimod/constrained/constrained.py
thisac: Fair enough. I guess both is fine, although I generally find using `typing.` everywhere makes the types less legible, although there are benefits to it as well. Python's typing system could definitely be better.

arcondello: I agree it's more cluttered visually. But I prefer cluttered signatures to a cluttered namespace. Determining what's a local symbol vs an imported one is much nicer with the qualified names. With `typing` most folks probably know which is which, but I prefer to be explicit when possible.

PR #1386, opened 2024-08-26
Knapsack Generator

k8culver:
A knapsack generator that allows for custom values, weights, and capacity to be passed in instead of randomly generating these values.

Updated the old random_knapsack generator to call the new generator and pass in the random values.

Added additional test to check if the shapes of values and weights are equal.


arcondello (approved):
Looks great!

Issue #1370, opened 2024-06-03
Invalid escape sequences


DeborahVolpe:
Your code presents some invalid escape sequences that are considered errors from python 11. Please check: https://docs.astral.sh/ruff/rules/invalid-escape-sequence/
A clear and concise description of the bug.

**Steps To Reproduce**
With CI workflow for analyzing the code, errors like: 

E   SyntaxError: invalid escape sequence '\?'

rises in the comment lines

**Expected Behavior**

**Environment**
 - for all the OS
 - for python in range 3.9 to 3.12

PR #1387, opened 2024-08-28
prevent syntax warnings for invalid escape characters

richrines1:
importing dimod currently results in a number of syntax warnings due to invalid escape sequences in docstrings in recent python versions (see #1370). This just prevents the warnings by marking the offending docstrings as `r"raw string literals"`


arcondello:
The CI failures are unrelated to this PR. We'll make a separate PR with a fix.

arcondello (approved):
Looks good to me! Thanks!

PR #1390, opened 2024-09-20
Multiple Knapsack Generator

k8culver:
A multiple knapsack generator that allows for custom values, weights, and capacities to be passed in instead of randomly generating these values.

Updated the old `random_multi_knapsack` generator to call the new generator and pass in the random values.

Added additional test to check if the shapes of values and weights are equal.

### DIFF
--- a/dimod/generators/multi_knapsack.py
+++ b/dimod/generators/multi_knapsack.py
@@ -17,8 +17,62 @@ import numpy as np
 from dimod.binary_quadratic_model import BinaryQuadraticModel
 from dimod.constrained import ConstrainedQuadraticModel
 from typing import Tuple
+from dimod.typing import ArrayLike
 
-__all__ = ['random_multi_knapsack']
+__all__ = ['multi_knapsack', 'random_multi_knapsack']
+
+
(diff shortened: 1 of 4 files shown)


dimod/generators/multi_knapsack.py
```diff
@@ -17,8 +17,62 @@
 from dimod.binary_quadratic_model import BinaryQuadraticModel
 from dimod.constrained import ConstrainedQuadraticModel
 from typing import Tuple
+from dimod.typing import ArrayLike
```
arcondello, line 20: Might as well fix the import order while we're at it.

```python
from typing import Tuple

import numpy as np

from dimod.constrained import ConstrainedQuadraticModel
from dimod.typing import ArrayLike
```

arcondello (approved):

dimod/generators/multi_knapsack.py
k8culver: Updated: https://github.com/dwavesystems/dimod/pull/1390/commits/c75397236d613f7adedf80154df29aaab3a89b89

Issue #1395, opened 2024-12-05
`OrderedDict` repr change causes doctest failures


randomir:
In [Python 3.12, `OrderedDict` has changed](https://docs.python.org/3.12/whatsnew/changelog.html#id139) its `repr` to use regular dictionary formatting instead of pairs of keys and values. See https://github.com/python/cpython/issues/101446 / https://github.com/python/cpython/pull/101661.

As a result, we have a few doctests failing:

```
File "../env/lib/python3.12/site-packages/dimod/reference/composites/tracking.py", line ?, in default
Failed example:
    sampler.input
Expected:
    OrderedDict([('h', {'a': -1}), ('J', {('a', 'b'): 1}), ('num_reads', 5)])
Got:
    OrderedDict({'h': {'a': -1}, 'J': {('a', 'b'): 1}, 'num_reads': 5})
**********************************************************************
File "../env/lib/python3.12/site-packages/dimod/reference/composites/tracking.py", line ?, in default
Failed example:
    sampler.input
Expected:
    OrderedDict([('Q', {('a', 'b'): 1})])
Got:
    OrderedDict({'Q': {('a', 'b'): 1}})
```

PR #1396, opened 2024-12-05
Use dict rather than OrderedDict in TrackingComposite

arcondello:
`TrackingComposite` was added while we still supported Python 2.7 (https://github.com/dwavesystems/dimod/pull/484). Now that `dict` is ordered, we can just use `dict` rather than `OrderedDict`.

Closes https://github.com/dwavesystems/dimod/issues/1395


randomir (approved):
LGTM!

PR #1397, opened 2024-12-18
Add Quadratic Multi Knapsack Generator

k8culver:
A quadratic multiple knapsack generator that allows for custom values, weights, profits and capacities.

@arcondello: @eordog and I chatted about whether it makes more sense for the diagonal of `profits` to be the `values` instead of passing `values` in as a separate array. Some of her benchmarking datasets have `values` as a separate array and some have it as the diagonal of `profits`. Curious if you have any thoughts on this. I went with separate as that's how I did it for `quadratic_knapsack`, whatever we end up going with I think these two functions should be consistent.

### DIFF
--- a/dimod/generators/multi_knapsack.py
+++ b/dimod/generators/multi_knapsack.py
@@ -20,7 +20,7 @@ from dimod.binary_quadratic_model import BinaryQuadraticModel
 from dimod.constrained import ConstrainedQuadraticModel
 from dimod.typing import ArrayLike
 
-__all__ = ['multi_knapsack', 'random_multi_knapsack']
+__all__ = ['multi_knapsack', 'quadratic_multi_knapsack', 'random_multi_knapsack']
 
 
 def multi_knapsack(values: ArrayLike,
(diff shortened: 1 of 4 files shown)


arcondello:
We could just add `values` to the diagonal of `profits` (when creating the model, not mutating the arrays passed in). And allow the user to specify `None` when passing in `values`? Might be overkill but would support both use cases.

k8culver:
> We could just add `values` to the diagonal of `profits` (when creating the model, not mutating the arrays passed in). And allow the user to specify `None` when passing in `values`? Might be overkill but would support both use cases.

If I keep this consistent between `quadratic_knapsack` and `quadratic_multi_knapsack` I'll need to change the order of the parameters for the existing `quadratic_knapsack` function to have `values` be optional, is this a concern since it will impact existing use of this function or how concerned about existing use are we?

arcondello:
Ok, in that case forget that idea! We can always do it later as a feature if it ends up being important.

arcondello (approved):

PR #1411, opened 2025-08-29
Add divisor argument to BQM.scale()

arcondello:
One approach to partially address https://github.com/dwavesystems/dwave-system/issues/591.

This of course doesn't solve all numeric issues, but it should make them rarer. To be tested.


arcondello:
Closing as stale

PR #1414, opened 2026-02-19
Add effective sample size estimators

kevinchern:
Add effective sample size estimators based on [Revisiting the Gelman-Rubin Diagnostic](https://arxiv.org/abs/1812.09384)


dimod/sampleset.py
```diff
@@ -1984,3 +1984,126 @@ def _as_samples_sampleset(samples_like: SampleSet,
...
+    if x.ndim != 2:
+        raise ValueError("The input matrix ``x`` should have shape (m, n) where m indexes independent "
+                         f"Markov chains, and n indexes time. ``x`` has shape {x.shape}")
+    m, n = x.shape
+    ess = estimate_effective_sample_size(x, b)
+    return m * n / ess
+
+print("waaaaaat0")
+print(estimate_effective_sample_size(np.random.uniform(0, 1, (100, 123))))
+print("waaaaaat")
```
kevinchern: -.-

dimod/ess.py
```diff
@@ -0,0 +1,123 @@
...
+def estimate_effective_sample_size(x: np.ndarray, b: int = None) -> float:
+    """Estimates the effective sample size of `x`.
+
+    The effective sample size (ESS) is the number of effectively independent samples drawn from
+    Markov chains' stationary distribution. The univariate estimator implemented here is the
+    (multivariate) estimator defined in the first equation at the top of page 14 in
+    `<Revisiting the Gelman-Rubin Diagnostic https://arxiv.org/abs/1812.09384>`_.
+
+    Args:
+        x (np.array): An (m, n) matrix where rows index independent Markov chains and columns index
```
randomir: Types can (should) be omitted from docstrings when code is annotated.

dimod/ess.py
```diff
@@ -0,0 +1,123 @@
...
+#    limitations under the License.
+
+from __future__ import annotations
+
+import numpy as np
+
+__all__ = ['estimate_effective_sample_size']
+
+
+def estimate_effective_sample_size(x: np.ndarray, b: int = None) -> float:
```
randomir: Optional:
```suggestion
def estimate_effective_sample_size(x: np.ndarray, b: int | None = None) -> float:
```

dimod/ess.py
```diff
@@ -0,0 +1,123 @@
...
+        x (np.array): An (m, n) matrix where rows index independent Markov chains and columns index
+            time steps.
+        b (int): Batch size of the estimator. If ``None``, then ``b`` is set to the floor of the
+            square root of ``n``. Defaults to None.
+
+    Returns:
+        float: Replicated lugsail estimate of the correlated sample's variance.
+    """
+    m, n = x.shape
+    b = b or round(n**0.5)
```
randomir: Shortcircuiting is not casting. This check will happily pass through `b = -2`.

I suggest you move this to the public function, where you validate `b`.

dimod/ess.py
```diff
@@ -0,0 +1,123 @@
...
+    Args:
+        x (np.array): An (m, n) matrix where rows index independent Markov chains and columns index
+            time steps.
+        b (int): Batch size of the estimator. If ``None``, then ``b`` is set to the floor of the
+            square root of ``n``. Defaults to None.
+
+    Returns:
+        float: An estimate of the effective sample size of ``x``.
+    """
+    if isinstance(b, int) and b < 3:
```
randomir: Since you check `b` at the beginning, and only later set its default from `n`, it's possible to end up with `b < 3` if `n < 7` (and `b` was omitted)...

dimod/ess.py
```diff
@@ -0,0 +1,123 @@
...
+    Args:
+        x (np.array): An (m, n) matrix where rows index independent Markov chains and columns index
+            time steps.
+        b (int): Batch size of the estimator.
+
+    Returns:
+        float: Replicated batch means estimate.
+    """
+    m, n = x.shape
+    if b > n:
```
randomir: This could also be moved to the public function above to centralize (and fix) validation.

randomir (commented):
A few minor suggestions.

dimod/ess.py
```diff
@@ -0,0 +1,123 @@
...
+
+from __future__ import annotations
+
+import numpy as np
+
+__all__ = ['estimate_effective_sample_size']
+
+
+def estimate_effective_sample_size(x: np.ndarray, b: int = None) -> float:
+    """Estimates the effective sample size of `x`.
```
thisac:
```suggestion
    """Estimates the effective sample size of ``x``.
```

dimod/ess.py
```diff
@@ -0,0 +1,123 @@
...
+    # = first equation at the top of page 10
+    # = estimate of the distribution's variance
+
+    ess = m * n * sigma_squared / tau_squared
+    # = estimate of the effective sample size
+    # = first equation at the top of page 14
+    return ess.item()
+
+
+def _estimate_replicated_lugsail_batch_means(x: np.ndarray, b: int = None) -> float:
```
thisac:
```suggestion
def _estimate_replicated_lugsail_batch_means(x: np.ndarray, b: int | None = None) -> float:
```

dimod/ess.py
```diff
@@ -0,0 +1,123 @@
...
+        float: Replicated batch means estimate.
+    """
+    m, n = x.shape
+    if b > n:
+        raise ValueError(
+            f"Batch size should be no more than ``n``. Batch size is {b} and ``n`` is {n}."
+        )
+
+    trimmed_length = b * (n // b)
+    n_batches = trimmed_length / b
```
thisac: Avoids one unnecessary operation, and keeps `n_batches` as int.
```suggestion
    n_batches = n // b
    trimmed_length = b * n_batches
```

dimod/ess.py
```diff
@@ -0,0 +1,123 @@
...
+
+    x = x[:, (n - trimmed_length):]
+    ybar = np.mean(np.split(x, n_batches, axis=1), axis=2).mT
+    muhat = x.mean()
+
+    res = ((ybar - muhat)**2).sum()*b/(n_batches*m-1)
```
thisac:
```suggestion
    res = ((ybar - muhat) ** 2).sum() * b / (n_batches * m - 1)
```

tests/test_ess.py
```diff
@@ -0,0 +1,57 @@
...
+            self.assertTrue(isnan(estimate_ess(np.ones((100, 1000)))))
+        with self.subTest("Null batch size should raise an error."):
+            self.assertRaisesRegex(ValueError, "Batch size should be at least three",
+                                   estimate_ess, np.ones((100, 1000)), 0)
+        with self.subTest("Batch size larger than chain length should raise an error."):
+            self.assertRaisesRegex(ValueError, "Batch size should be no more than ``n``",
+                                   estimate_ess, np.ones((100, 1000)), 1001)
+        with self.subTest("Inputs that are not 2D should raise an error."):
+            self.assertRaisesRegex(ValueError, "The input matrix ``x`` should have shape",
+                                   estimate_ess, np.ones((123, 100, 1000)), 234)
```
thisac: Empty lines in between subtests would make this more readable.

dimod/ess.py
```diff
@@ -0,0 +1,123 @@
+# Copyright 2026 D-Wave Quantum Inc.
```
thisac:
```suggestion
# Copyright 2026 D-Wave
```

dimod/ess.py
kevinchern: ah, nice!

dimod/ess.py
kevinchern: Ahhhhhhhhh, thanks!

dimod/ess.py
kevinchern: RE all `b` parameter comments:
I moved `b` to the public function and validate it there :)

tests/test_ess.py
```diff
@@ -0,0 +1,66 @@
+# Copyright 2026 D-Wave Quantum Inc.
```
randomir:
```suggestion
# Copyright 2026 D-Wave
```

randomir (approved):
LGTM!

pau557 (changes requested):
This looks good so far, but I would add more contextual elements

- Time steps are mentioned in the definition of the x argument, but not in the text body. It should be clear what these time steps represent.
- Similarly, what are the elements of the matrix? We know that they were generated in independent Markov chains at different time steps, but we don’t know what they are. Is it an aggregation metric?

I suggest a practical intuitive example of how this method is expected to be used within the context of dimod samplers.

And I think that it would be great to include an API that accepts dimod samplesets.

jackraymond (approved):
Lovely

jackraymond:
> This looks good so far, but I would add more contextual elements
> 
>     * Time steps are mentioned in the definition of the x argument, but not in the text body. It should be clear what these time steps represent.
> 
>     * Similarly, what are the elements of the matrix? We know that they were generated in independent Markov chains at different time steps, but we don’t know what they are. Is it an aggregation metric?
> 
> 
> I suggest a practical intuitive example of how this method is expected to be used within the context of dimod samplers.
> 
> And I think that it would be great to include an API that accepts dimod samplesets.

Using SimulatedAnnealingSampler as a fixed-temperature Markov Chain, for a ferromagnet J = {(i,j):-1/N for i in range(N) for j in range(i)}, would be a nice example. Allow beta as a control parameter (few effective samples beta~1, many effective samples beta<1). 
Can use ss.record.energy [many effective samples, effectively unimodal], or mag= sum(ss.record.sample,axis=0) [fewest effective samples, bimodal] as the statistic.

kevinchern:
> > This looks good so far, but I would add more contextual elements
> > ```
> > * Time steps are mentioned in the definition of the x argument, but not in the text body. It should be clear what these time steps represent.
> > 
> > * Similarly, what are the elements of the matrix? We know that they were generated in independent Markov chains at different time steps, but we don’t know what they are. Is it an aggregation metric?
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I suggest a practical intuitive example of how this method is expected to be used within the context of dimod samplers.
> > And I think that it would be great to include an API that accepts dimod samplesets.
> 
> Using SimulatedAnnealingSampler as a fixed-temperature Markov Chain, for a ferromagnet J = {(i,j):-1/N for i in range(N) for j in range(i)}, would be a nice example. Allow beta as a control parameter (few effective samples beta~1, many effective samples beta<1). Can use ss.record.energy [many effective samples, effectively unimodal], or mag= sum(ss.record.sample,axis=0) [fewest effective samples, bimodal] as the statistic.

@jackraymond `SimulatedAnnealingSampler`'s interface doesn't allow for extracting the Markov chain 😿; I'll dream up another example (open to suggestions)

jackraymond:
> > > This looks good so far, but I would add more contextual elements
> > > ```
> > > * Time steps are mentioned in the definition of the x argument, but not in the text body. It should be clear what these time steps represent.
> > > 
> > > * Similarly, what are the elements of the matrix? We know that they were generated in independent Markov chains at different time steps, but we don’t know what they are. Is it an aggregation metric?
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I suggest a practical intuitive example of how this method is expected to be used within the context of dimod samplers.
> > > And I think that it would be great to include an API that accepts dimod samplesets.
> > 
> > 
> > Using SimulatedAnnealingSampler as a fixed-temperature Markov Chain, for a ferromagnet J = {(i,j):-1/N for i in range(N) for j in range(i)}, would be a nice example. Allow beta as a control parameter (few effective samples beta~1, many effective samples beta<1). Can use ss.record.energy [many effective samples, effectively unimodal], or mag= sum(ss.record.sample,axis=0) [fewest effective samples, bimodal] as the statistic.
> 
> @jackraymond `SimulatedAnnealingSampler`'s interface doesn't allow for extracting the Markov chain 😿; I'll dream up another example (open to suggestions)

Good point!

kevinchern:
>     * Time steps are mentioned in the definition of the x argument, but not in the text body. It should be clear what these time steps represent.
> 
>     * Similarly, what are the elements of the matrix? We know that they were generated in independent Markov chains at different time steps, but we don’t know what they are. Is it an aggregation metric?
> 
> 
> I suggest a practical intuitive example of how this method is expected to be used within the context of dimod samplers.
> 
> And I think that it would be great to include an API that accepts dimod samplesets

@pau557 @jackraymond thanks again for the feedback!!

I improved the documentation by:
1. Describing more carefully what ESS is,
2. Adding two introductory references, and
3. Adding two intended-usage examples and one misuse example (QPU and MH, and SA respectively)

I also added a separate function that consumes as input a dimod sampleset and calls the array-version of the estimator. It defaults to energy function unless a test function is provided.

dimod/ess.py
```diff
@@ -0,0 +1,307 @@
...
+        >>> num_chains = 10
+        >>> bqm = power_r(512, num_vars)
+        >>> bqm.normalize()
+        >>> def test_fn(ss):
+        >>>     return ss.record.sample.mean(1)
+        >>> neal = SimulatedAnnealingSampler()
+        >>> sample_set = neal.sample(bqm, num_reads=num_reads)
+        >>> print("Effective sample size per chain (misuse):",
+        >>>       estimate_effective_sample_size_sampleset(sample_set, test_fn))
+        Effective sample size per chain (misuse): 58.061778108529666
```
pau557: why ESS 58.06 here, and 96.87 in line 227?

kevinchern: combination of high variance (mostly this), different test function, and bug (initial_states vs initial_state)

kevinchern:
@pau557
Updated documentation:
1) Warn about negative estimates
2) Remove the distracting example
3) Make example with MH two-part: 1 sweep vs 10 sweeps

Feature:
- Negative and NaN estimates will come with a warning message

tests/test_ess.py
```diff
@@ -0,0 +1,108 @@
...
+from dimod.ess import _estimate_replicated_batch_means
+from dimod.ess import estimate_effective_sample_size as estimate_ess
+from dimod.ess import estimate_effective_sample_size_sampleset as estimate_ess_ss
+from dimod.generators import power_r
+
+
+class TestEffectiveSampleSize(unittest.TestCase):
+
+    def test_estimate_ess_sample_set(self):
+        bqm = power_r(512, 64)
```
jackraymond: Why use an obscure model when something like
```dimod.BinaryQuadraticModel('SPIN').from_ising({},{(i,j): -1 for i in range(10) j in range(i)}) ```
would suffice.

**Review by jackraymond — COMMENTED** [review]
LGTM

**Review by jackraymond — APPROVED** [review]
LGTM

File: tests/test_ess.py (hunk 15 above)
Comment (jackraymond) [reply tests/test_ess.py]: Even better (but need to be careful with QPU autoscale)
```
dimod.BinaryQuadraticModel('SPIN').from_ising({i: 0 for i in range(10)}, {})  # 10 independent unbiased spins
```

File: dimod/ess.py [hunk 16]
```diff
@@ -0,0 +1,278 @@
...
+def estimate_effective_sample_size(x: np.ndarray, b: int | None = None) -> float:
+    """Estimates the effective sample size of ``x``.
+
+    NOTE: The estimate can be nan or negative for extreme cases (e.g., constants). This can occur
+    by definition of the estimator and is not a bug.
+
+    The effective sample size (ESS; a formally defined statistic) quantifies the number of,
+    effectively, independent samples drawn from an autocorrelated sampler. For example, samples from
+    a Markov chain sampler such as Metropolis-Hastings. For another example, samples from an annealing
+    quantum computer (due to high-frequency noise). The quantification of this notion of independence
```
pau557: it should be "low-frequency noise"

tests/test_ess.py
pau557: I agree, worth making it a simpler model

pau557:
LGTM

dimod/ess.py
kevinchern: Changed both docstring occurrences

tests/test_ess.py
kevinchern: Removed ran_r and used a simpler model (but nonzero h). zero h = 0 Hamiltonian = constant = nans (expected behaviour)

pau557 (approved):

dimod/ess.py
```diff
@@ -0,0 +1,278 @@
...
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Callable, Iterable
+
+if TYPE_CHECKING:
+    import dimod
```
thisac: Minor preference, but I prefer `from dimod import SampleSet` here to narrow down the import a bit and be more specific about the type being imported.

dimod/ess.py
```diff
@@ -0,0 +1,278 @@
...
+
+import numpy as np
+
+__all__ = ['estimate_effective_sample_size', 'estimate_effective_sample_size_sampleset']
+
+
+def estimate_effective_sample_size_sampleset(
+    sample_set: dimod.SampleSet,
+    test_function: Callable[[dimod.SampleSet], Iterable[float]] | None = None,
+    b: int | None = None
```
thisac: Better naming?
```suggestion
    batch_size: int | None = None
```

dimod/ess.py
```diff
@@ -0,0 +1,278 @@
...
+    """
+    if test_function is None:
+        def test_function(ss: dimod.SampleSet):
+            return ss.record.energy
+    try:
+        stat = list(test_function(sample_set))
+    except Exception as e:
+        raise ValueError("Test function is invalid. Does the test function have a correct signature? "
+                         "The test function should consume a ``dimod.SampleSet`` and return an "
+                         f"iterable of floats of length equal to the sample size. The exception is {e}")
```
thisac: I'd suggest catching more specific exceptions here rather than a general `Exception`, if possible. Also, better to raise from caught exception than printing it in message:
```python
try:
    ...
except Exception as e:
    raise ValueError(...) from e
```

dimod/ess.py
```diff
@@ -0,0 +1,278 @@
...
+    if X.shape[0] != len(sample_set):
+        raise ValueError("The test function should consume a sample set and return an iterable of "
+                         "floats of length equal to the sample size of the sample set. The sample "
+                         f"size of the sample set is {len(sample_set)} and the length of the test "
+                         f"function output has shape {X.shape}.")
+
+    return estimate_effective_sample_size(X.reshape(1, -1), b)
+
+
+def estimate_effective_sample_size(x: np.ndarray, b: int | None = None) -> float:
```
thisac: Similarly here, is `batch_size` a better name?
```suggestion
def estimate_effective_sample_size(x: np.ndarray, batch_size: int | None = None) -> float:
```

dimod/ess.py
```diff
@@ -0,0 +1,278 @@
...
+
+    The univariate estimator implemented here is the (multivariate) estimator defined in the first
+    equation at the top of page 14 in
+    `<Revisiting the Gelman-Rubin Diagnostic https://arxiv.org/abs/1812.09384>`_.
+
+    For an introduction to effective sample size, see
+    `<Stan Manual https://mc-stan.org/docs/2_21/reference-manual/effective-sample-size-section.html>`_
+    or this
+    `<blog post https://andrewcharlesjones.github.io/journal/21-effective-sample-size.html>_.
+
```
thisac: This docstring seems identical to the one below. I'd say it's better to leave it in one function and either refer to it from the other docstring (e.g., "see ... for more details") or move it to the docs (i.e., put it under a "ESS" heading in `docs/utilities.rst`).

dimod/ess.py
```diff
@@ -0,0 +1,278 @@
...
+        float: An estimate of the effective sample size of ``x``. The estimate can be NaN or
+        even negative when the input chain is nearly constant. This can occur by definition of the
+        estimator and is not a bug.
+    """
+    if x.ndim != 2:
+        raise ValueError("The input matrix ``x`` should have shape (m, n) where m indexes "
+                         f"independent Markov chains and n indexes time. ``x`` has shape {x.shape}")
+    m, n = x.shape
+    if b is None:
+        b = int(floor(n**0.5))
```
thisac: Doesn't `floor` already return an `int`?
```suggestion
        b = floor(n**0.5)
```

dimod/ess.py
```diff
@@ -0,0 +1,278 @@
...
+    ess = (m * n * sigma_squared / tau_squared).item()
+    # = estimate of the effective sample size
+    # = first equation at the top of page 14
+    if isnan(ess) or ess < 0:
+        warn("The estimated effective sample size is NaN or negative. This can occur by definition "
+             "of the estimator and is not a bug.")
+    return ess
+
+
+def _estimate_replicated_batch_means(x: np.ndarray, b: int) -> float:
```
thisac: If you decide to update it above.
```suggestion
def _estimate_replicated_batch_means(x: np.ndarray, batch_size: int) -> float:
```

dimod/ess.py
```diff
@@ -0,0 +1,278 @@
...
+    """
+    n = x.shape[1]
+
+    n_batches = n // b
+    trimmed_length = b * n_batches
+
+    x = x[:, (n - trimmed_length):]
+    ybar = np.mean(np.split(x, n_batches, axis=1), axis=2).mT
+
+    res = b*np.var(ybar, ddof=1)
```
thisac:
```suggestion
    res = b * np.var(ybar, ddof=1)
```

thisac (commented):
Don't forget to add `import dimod.ess` to `dimod/__init__.py`, for e.g. the following to work:
```python
import dimod
dimod.ess.estimate_effective_sample_size
```

You can also add `from dimod.ess import *` if you want to be able to do e.g., `from dimod import estimate_effective_sample_size`.

dimod/ess.py
```diff
@@ -14,29 +14,196 @@
...
+from warnings import warn
 
 import numpy as np
 
-__all__ = ['estimate_effective_sample_size']
+__all__ = ['estimate_effective_sample_size', 'estimate_effective_sample_size_sampleset']
+
+
+def estimate_effective_sample_size_sampleset(
+    sample_set: dimod.SampleSet,
```
randomir: What's the reason for having the estimator specialized for `SampleSet`s?

If you want for you estimator to accept both NumPy arrays and SampleSets, why not use `dimod.typing.SamplesLike` here, and convert to an array with [`dimod.as_samples`](https://docs.dwavequantum.com/en/latest/ocean/api_ref_dimod/generated/dimod.as_samples.html#dimod.as_samples)? That would make the API nicer and more consistent with dimod.

dimod/ess.py
```diff
@@ -14,29 +14,196 @@
...
+    or this
+    `<blog post https://andrewcharlesjones.github.io/journal/21-effective-sample-size.html>_.
+
+    Examples:
+        These two examples demonstrate typical use cases of the estimator based on an energy statistic.
+        The first example measures the QPU's ESS. The second example measures the ESS of two
+        Metropolis-Hastings samplers (one with one sweep, another with ten sweeps).
+
+        Example (1): QPU
+        >>> import numpy as np
```
randomir: The inline code notation is useful for one-liners, but not for these long examples. Rather, use
```
.. code-block:: python

    import nump as np
    ...
```

randomir (commented):
Unless you have a very strong reason for having both `estimate_effective_sample_size` and  `estimate_effective_sample_size_sampleset`, I'd suggest you combine them and work with `SamplesLike`.

dimod/ess.py
kevinchern: If I understood correctly, I don't think it's possible to be more exact because the test function can be any arbitrary function.

dimod/ess.py
kevinchern: The examples are slightly different, but let me get back to this after addressing @randomir 's comment about using a "samples_like" signature instead of having two public functions.

dimod/ess.py
kevinchern: Indeed! I always mix up the behaviour between floor/ceil and `//`

dimod/ess.py
kevinchern: Updated. Summary of changes: I split `stat = list(test_function(sample_set))` into `output = test_function(sample_set)` and then `list(output)` to catch arbitrary errors from test function and raise a runtime error. once test function runs, I validate output types.

dimod/ess.py
kevinchern: My rebuttal is the two functions are different objects mathematically.
1. `compute_ess` (abbreviated) maps $R^{N,T}\mapsto R$, where $N$ represents the number of independent Markov chains, $T$ is the length of each chain.
2. `compute_ess_sampleset` maps $\{\pm 1\}^{T, d} \mapsto R^{1, T}$, then that's passed in to `compute_ess` with $N=1$. $d$ here is the dimension of the model, i.e., number of spins. There's no notion of "number of variables" in `compute_ess`

randomir: I see. In that case I think these functions need better names. Perhaps something like:
```
estimate_ess_markov
estimate_ess_ising
```

dimod/__init__.py
```diff
@@ -65,5 +65,8 @@
 
 from dimod.vartypes import *
 
+import dimod.ess
+from dimod.ess import estimate_effective_sample_size
```
randomir: You probably want to import both variants here.

dimod/ess.py
```diff
@@ -0,0 +1,303 @@
...
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Callable, Iterable
+
+if TYPE_CHECKING:
+    from dimod import SampleSet
+    from dimod.typing import SamplesLike, SampleLike
```
randomir: Can be dropped if unused.
```suggestion
```

dimod/ess.py
```diff
@@ -0,0 +1,303 @@
...
+import numpy as np
+
+__all__ = ['estimate_effective_sample_size', 'estimate_effective_sample_size_sampleset']
+
+
+def estimate_effective_sample_size_sampleset(
+    sample_set: SampleSet,
+    test_function: Callable[[SampleSet], Iterable[float]] | None = None,
+    batch_size: int | None = None
+):
```
randomir: I'd reorder functions here. First, the general one, and then the one specialized for samplesets.

Also, would avoid duplication of docs here, and would refer to the general function docs as much as possible; describing only the diff here.

dimod/ess.py
```diff
@@ -0,0 +1,303 @@
...
+    `<Stan Manual https://mc-stan.org/docs/2_21/reference-manual/effective-sample-size-section.html>`_
+    or this
+    `<blog post https://andrewcharlesjones.github.io/journal/21-effective-sample-size.html>_.
+
+    Examples:
+        These two examples demonstrate typical use cases of the estimator based on an energy statistic.
+        The first example measures the QPU's ESS. The second example measures the ESS of two
+        Metropolis-Hastings samplers (one with one sweep, another with ten sweeps).
+
+        Example (1): QPU
```
randomir, line 41: This example being only slightly different from the one already in the `_sampleset()` method, perhaps it's worth highlighting the difference (justify its existence).

dimod/ess.py
kevinchern: Addressed following @thisac 's [suggestion to place common docstrings in a dedicated `ess.rst` file](https://github.com/dwavesystems/dimod/pull/1414#discussion_r2956575506)

dimod/ess.py
kevinchern: Added a few sentences to contrast it with the other function

dimod/ess.py
kevinchern: addressed in a .rst file

kevinchern:
Should this module be part of utilities or another instead? @randomir  @thisac

thisac:
> Should this module be part of utilities or another instead? @randomir @thisac

No strong opinions here. I think it's fine keeping it separate.

docs/ess.rst
```diff
@@ -0,0 +1,29 @@
+.. _dimod_ess:
+
+=========
+Effective Sample Size
+=========
```
randomir: Underlining too [short](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections).
```suggestion
=====================
Effective Sample Size
=====================
```

randomir:
> Should this module be part of utilities or another instead? @randomir @thisac

Not really sure `ess` (effective sample size) warrants a top-level module, but it also seems much more than a utils submodule.

Not a strong opinion, though. We can keep it as is.

dimod/ess.py
```diff
@@ -0,0 +1,275 @@
...
+    by definition of the estimator and is not a bug.
+
+    Examples:
+        These two examples demonstrate typical use cases of the estimator based on an energy statistic.
+        The first example measures the QPU's ESS. The second example measures the ESS of two
+        Metropolis-Hastings samplers (one with one sweep, another with ten sweeps).
+
+        Example (1): QPU
+        .. code-block::python
+            import numpy as np
```
thisac: I don't _think_ this is required, but it makes for a cleaner source.
```suggestion
        .. code-block:: python

            import numpy as np
```

dimod/ess.py
```diff
@@ -0,0 +1,275 @@
...
+            from dimod.ess import estimate_effective_sample_size
+            from dimod.generators import power_r
+            from dwave.system import DWaveCliqueSampler
+            markov_chain_length = 100
```
thisac, line 49: You can, and should, now add in relevant white-spaces and empty lines.
```suggestion
            from dwave.system import DWaveCliqueSampler

            markov_chain_length = 100
```

docs/ess.rst
```diff
@@ -0,0 +1,29 @@
...
+Two such examples of test functions are magnetization and energy.
+
+The univariate estimator implemented here is the (multivariate) estimator defined in the first
+equation at the top of page 14 in
+`Revisiting the Gelman-Rubin Diagnostic <https://arxiv.org/abs/1812.09384>`_.
+
+For an introduction to effective sample size, see the
+`Stan Manual <https://mc-stan.org/docs/2_21/reference-manual/effective-sample-size-section.html>`_
+or this
+`blog post <https://andrewcharlesjones.github.io/journal/21-effective-sample-size.html>_.
```
thisac:
```suggestion
`blog post <https://andrewcharlesjones.github.io/journal/21-effective-sample-size.html>`_.
```

thisac (commented):
Just a few minor doc changes, otherwise LGTM!

dimod/ess.py
```diff
@@ -0,0 +1,277 @@
...
+                for _ in range(num_chains)]
+            )
+            print("Effective sample size per chain (QPU):",
+                  estimate_effective_sample_size(qpu_energy)/num_chains)
+            # Effective sample size per chain (QPU): 71.87414368659094
+
+
+        Example (2): Metropolis-Hastings
+        .. code-block::python
+            import numpy as np
```
thisac, line 68:
```suggestion

        Example (2): Metropolis-Hastings

        .. code-block:: python

            import numpy as np
```

dimod/ess.py
```diff
@@ -0,0 +1,277 @@
...
+            from dwave.samplers import SimulatedAnnealingSampler
+            from dimod.ess import estimate_effective_sample_size
+            from dimod.generators import power_r
+            neal = SimulatedAnnealingSampler()
```
thisac, line 73:
```suggestion
            from dimod.generators import power_r

            neal = SimulatedAnnealingSampler()
```

dimod/ess.py
```diff
@@ -0,0 +1,277 @@
...
+                markov_chains.append(chain)
+            mc_energy = np.array(markov_chains)
+            print("Effective sample size per chain (MH):",
+                  estimate_effective_sample_size(mc_energy)/num_chains)
+            # Effective sample size per chain (MH): 12.496482970401358
+
+        Use a larger number of sweeps to achieve larger ESS
+        >>> num_sweeps = 10
+        Effective sample size per chain (MH): 64.44999653861495
+
```
thisac, line 99:
```suggestion
```

dimod/ess.py
```diff
@@ -0,0 +1,277 @@
...
+):
+    """Estimates the effective sample size of ``sample_set``.
+
+    NOTE: The estimate can be nan or negative for extreme cases (e.g., constants). This can occur
+    by definition of the estimator and is not a bug.
+
+    This function differs from ``compute_ess`` in that it assumes its input, ``sample_set``, consists
+    of states from a single Markov chain. By default, when no test
+    function is supplied, the effective sample size is estimated based on the sample set's energies.
+
```
thisac, line 161:
```suggestion
```

dimod/ess.py
```diff
@@ -0,0 +1,277 @@
...
+        ESS based on the magnetization of the system.
+
+
+        Example (1): QPU
+        .. code-block::python
+            import numpy as np
+            from dwave.system import DWaveCliqueSampler
+            from dimod.ess import estimate_effective_sample_size
+            from dimod.generators import power_r
+            markov_chain_length = 100
```
thisac, line 176:
```suggestion
        Example (1): QPU

        .. code-block:: python

            import numpy as np
            from dwave.system import DWaveCliqueSampler
            from dimod.ess import estimate_effective_sample_size
            from dimod.generators import power_r

            markov_chain_length = 100
```

dimod/ess.py
```diff
@@ -0,0 +1,277 @@
...
+            num_vars = 33
+            bqm = power_r(512, num_vars)
+            bqm.normalize()
+            qpu = DWaveCliqueSampler()
+
+            def test_fn(ss):
+                return ss.record.sample.mean(1)
+            sample_set = qpu.sample(bqm, num_reads=markov_chain_length, answer_mode="raw")
```
thisac, line 185:
```suggestion
                return ss.record.sample.mean(1)

            sample_set = qpu.sample(bqm, num_reads=markov_chain_length, answer_mode="raw")
```

dimod/ess.py
```diff
@@ -0,0 +1,277 @@
...
+            print("Effective sample size (QPU):",
+                  estimate_effective_sample_size_sampleset(sample_set, test_fn))
+            # Effective sample size per chain (QPU): 69.69037448554732
+
```
thisac, line 189:
```suggestion
```

thisac (approved):

PR #1415, opened 2026-02-23
Expand dimod.make_quadratic documentation

iamashwin99:
- Add example output for make_quadratic function
- Added notes on variable types in make_quadratic function.

### DIFF
--- a/dimod/higherorder/utils.py
+++ b/dimod/higherorder/utils.py
@@ -298,7 +298,7 @@ def make_quadratic(poly: Union[Polynomial, BinaryPolynomial], strength: float,
             If not provided, a new binary quadratic model is created.
 
     Notes:
-        When reducing higher-order polynomials to degree 2, the resulting model 
+        When reducing higher-order polynomials to degree 2, the resulting model
         will contain three types of variables:
 
         * **Original Variables**: Variables mapped directly from the input 
--- a/dimod/higherorder/utils.py
+++ b/dimod/higherorder/utils.py
@@ -272,7 +272,7 @@ def make_quadratic_cqm(poly: Union[Polynomial, BinaryPolynomial],
 def make_quadratic(poly: Union[Polynomial, BinaryPolynomial], strength: float,
                    vartype: Optional[Vartype] = None,
                    bqm: Optional[BinaryQuadraticModel] = None) -> BinaryQuadraticModel:
-"""Create a binary quadratic model from a higher order polynomial.
+    """Create a binary quadratic model from a higher order polynomial.
 
     Args:
         poly:
--- a/dimod/higherorder/utils.py
+++ b/dimod/higherorder/utils.py
@@ -272,7 +272,7 @@ def make_quadratic_cqm(poly: Union[Polynomial, BinaryPolynomial],
 def make_quadratic(poly: Union[Polynomial, BinaryPolynomial], strength: float,
                    vartype: Optional[Vartype] = None,
                    bqm: Optional[BinaryQuadraticModel] = None) -> BinaryQuadraticModel:
-    """Create a binary quadratic model from a higher order polynomial.
+"""Create a binary quadratic model from a higher order polynomial.
 
     Args:
         poly:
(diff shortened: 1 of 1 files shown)


dimod/higherorder/utils.py
```diff
@@ -297,10 +297,27 @@ def make_quadratic(poly: Union[Polynomial, BinaryPolynomial], strength: float,
             Terms of the reduced polynomial are added to this binary quadratic model.
             If not provided, a new binary quadratic model is created.
 
+    Notes:
+        When reducing higher-order polynomials to degree 2, the resulting model
+        will contain three types of variables:
```
JoelPasvolsky:
```suggestion
    Returns:
        :class:`~dimod.binary.BinaryQuadraticModel`: A binary quadratic model
        with three types of variables:
```

dimod/higherorder/utils.py
```diff
@@ -297,10 +297,27 @@ def make_quadratic(poly: Union[Polynomial, BinaryPolynomial], strength: float,
...
+        * **Original Variables**: Variables mapped directly from the input 
+          polynomial (e.g., `0`, `1`, `12`).
+        * **Product Variables**: Intermediate variables introduced to reduce 
+          higher-order terms, representing the product of two variables. They 
+          are typically named by joining the original variables (e.g., `'0*1'` 
+          represents the product of `0` and `1`).
+        * **Auxiliary Variables**: Variables introduced to enforce the penalty 
+          constraints that ensure the mathematical relationship of the product 
+          variables holds true at the ground state (e.g., `'aux0,1'` enforces 
+          the constraint for `'0*1'`).
```
JoelPasvolsky:
```suggestion
        *   Original variables: Variables mapped directly from the input
            polynomial (e.g., ``0``, ``1``, ``12``).
        *   Product variables: Intermediate variables introduced to reduce
            higher-order terms, representing the product of two variables. These
            are typically named by joining the original variables (e.g., ``0*1``
            represents the product of ``0`` and ``1``).
        *   Auxiliary variables: Variables introduced to enforce the penalty
            constraints that ensure the mathematical relationship of the product
            variables holds true at the ground state (e.g., ``aux0,1`` enforces
            the constraint for ``0*1``).
```

dimod/higherorder/utils.py
```diff
@@ -297,10 +297,27 @@ def make_quadratic(poly: Union[Polynomial, BinaryPolynomial], strength: float,
...
+
     Examples:
 
         >>> poly = {(0,): -1, (1,): 1, (2,): 1.5, (0, 1): -1, (0, 1, 2): -2}
         >>> bqm = dimod.make_quadratic(poly, 5.0, dimod.SPIN)
+        >>> bqm
+        BinaryQuadraticModel({0: -3.5, 1: -1.5, '0*1': -2.5, 'aux0,1': -5.0, 2: 1.5}, {(1, 0): 1.5, ('0*1', 0): 2.5, ('0*1', 1): 2.5, ('aux0,1', 0): 5.0, ('aux0,1', 1): 5.0, ('aux0,1', '0*1'): 5.0, (2, '0*1'): -2.0}, 10.0, 'SPIN')
```
JoelPasvolsky:
```suggestion
        >>> list(bqm.variables)
        [0, 1, '0*1', 'aux0,1', 2]
```

More readable and short.

JoelPasvolsky (commented):
Thank you for contributing to the Ocean SDK, @iamashwin99.

The comments below are to keep your additions aligned with Ocean formatting, remove trailing whitespaces, and return a shorter, more easily parsable result.

dimod/higherorder/utils.py
iamashwin99: I think the full bqm would be a better description. It would also give hint about how the auxiliary qubits are generated. What do you think?

JoelPasvolsky: Unless it provides high value to users (e.g., a list of parameters, typical timing), we try to avoid long outputs because they're not easy to parse, make the example look daunting, and can require maintenance (e.g., the recent backward-incompatible NumPy update that changes formats). My thinking is that such a short example lets users easily copy & paste so those who want to see the exact BQM structure can do so without difficulty. The list output highlights the generated variables you describe and additional, more robust ways to provide an output with more information could be  

```
>>> len(bqm.linear) + len(bqm.quadratic)
12
```

or 

```
>>> print(bqm.to_polystring())
10 - 3.5*v0 - 1.5*v1 - 2.5*0*1 - 5*aux0,1 + 1.5*v2 + 1.5*v0*v1 + 2.5*v0*0*1 + 2.5*v1*0*1 + 5*v0*aux0,1 + 5*v1*aux0,1 + 5*0*1*aux0,1 - 2*0*1*v2
```

and even for the latter, preferably shorten it with ellipses for readability, something perhaps like:

```
>>> print(bqm.to_polystring())
10 - ... 2.5*0*1 - 5*aux0,1 + ... 2.5*v1*0*1 + 5*v0*aux0,1 + 5*v1*aux0,1 + ...
```

Can one of these or another robust & short output do most of the work? If yes, that would be preferable. Thank you!

iamashwin99:
I see your point, in that case I would suggest we go with your first suggestion. It is more clear. Thanks for your suggestions!

JoelPasvolsky (approved):
Thank you, @iamashwin99!

Some CircleCI tests are failing but unrelated to this PR and are addressed by other PRs.

thisac:
@iamashwin99 Could you rebase on main? This should fix the failing checks, after which we can merge.

PR #1416, opened 2026-02-23
Simplify xcode version management in CI by using `current`

randomir:
Also, fix the use of a legacy resource class, `macos.m1.medium.gen1`.

PR #1417, opened 2026-02-24
Update xcode and airspeed velocity in CI

thisac:


.circleci/config.yml
```diff
@@ -72,8 +72,7 @@ jobs:
         type: string
 
     macos:
-      xcode: 16.4.0  # latest as of August 2025
-    resource_class: macos.m1.medium.gen1
+      xcode: 26.3.0  # latest as of Debruary 2026
```
randomir:
```suggestion
      xcode: 26.3.0  # latest as of February 2026
```

.circleci/config.yml
```diff
@@ -269,8 +268,7 @@ jobs:
 
   test-osx-cpp:
     macos:
-      xcode: 15.3.0
-    resource_class: macos.m1.medium.gen1
+      xcode: 26.3.0  # latest as of Debruary 2026
```
randomir:
```suggestion
      xcode: 26.3.0  # latest as of February 2026
```

randomir (commented):
I tried updating to `16.4.0` and dropping resource class in #1416, but some tests are failing on osx in cibuildwheel on arm. Chose 16.4.0 over latest for compatibility.

We should also figure out why airspeed-velocity test is failing now.

.circleci/config.yml
```diff
@@ -141,10 +140,11 @@ jobs:
 
   test-airspeed-velocity:
     docker:
-      - image: cimg/python:3.10
+      - image: cimg/python:3.12  # latest version supported by asv
 
     steps:
-      - checkout
+      - checkout:
+          method: full
```
thisac, line 147: This seemed to be the fix for the `asv` issues we were having. 

For some reason the checkout step for running benchmarks comparisons on main failed due to corruption. Full checkout instead of blobless (default since Nov 4th, 2025) fixes it.

See https://support.circleci.com/hc/en-us/articles/44509591136923-Blobless-Checkout-and-How-to-Opt-In-and-Out

.circleci/config.yml
```diff
@@ -156,7 +156,7 @@ jobs:
           command: |
             . env/bin/activate
             asv machine --yes
-            asv run --quick --strict
+            asv run --quick
```
thisac, line 159: `--strict` no longer required or available in `asv>=0.6.0` (see [here](https://asv.readthedocs.io/en/stable/changelog.html#id15)).

randomir (approved):
LGTM! Thanks!

PR #1420, opened 2026-04-30
Add problem generators from dwave-networkx algorithms

randomir:
Decouple and migrate BQM/CQM model generators from dwave-networkx/dwave-graphs (`algorithms` namespace) to `dimod.generators`. Also, adapt for dimod/Ocean style of code/docs.

New submodules/generators:
- `dimod.generators.coloring`:
  - `vertex_coloring`, `min_vertex_coloring`
- `dimod.generators.markov`:
  - `markov_network`
- `dimod.generators.matching`:
  - `matching`, `maximal_matching`, `min_maximal_matching`
- `dimod.generators.partition`:
  - `graph_partition`
- `dimod.generators.social`:
  - `structural_imbalance`
- `dimod.generators.tsp`:
  - `traveling_salesperson`/`traveling_salesman`

Most of the generators have been generalized to now accept a `GraphLike`, but some still require a NetworkX graph as input.

We kept `networkx` as an optional dependency, though.

**AI Generation Disclosure**

No AI used.


dimod/generators/coloring.py
```diff
@@ -0,0 +1,248 @@
+# Copyright 2018 D-Wave Systems Inc.
```
arcondello: 2026, here and elsewhere. I think the files are sufficiently changed

arcondello (approved):
Looks great!

dimod/generators/coloring.py
```diff
@@ -0,0 +1,248 @@
...
+    return bqm
+
+
+@graph_argument('G', as_networkx=True)
+def _chromatic_number_upper_bound(G: 'nx.Graph') -> int:
+    # tries to determine an upper bound on the chromatic number of G
+    # Assumes G is not complete
+
+    # nx is an optional dependency, so we defer import
+    # it's safe to import here, as `@graph_argument` would fail otherwise
```
thisac, line 90: Does `@graph_argument` expect `networkx` or will it raise a reasonable exception if called without?

dimod/generators/coloring.py
```diff
@@ -0,0 +1,248 @@
...
+    for u, v in edges:
+        # NAND constraint
+        for c in colors:
+            bqm.add_quadratic((u, c), (v, c), 1)
+
+    return bqm
+
+
+@graph_argument('G', as_networkx=True)
+def _chromatic_number_upper_bound(G: 'nx.Graph') -> int:
```
thisac, line 85: What's the reason for using `@graph_argument` here since we're only allowing `G` to be a `nx.Graph`? Is it simply for the nx dependency check?

dimod/generators/coloring.py
```diff
@@ -0,0 +1,248 @@
...
+        weights = [p / (num_penalized + 1) for p in range(1, num_penalized + 1)]
+
+        for p, c in zip(weights, range(chromatic_lb, chromatic_ub)):
+            for v in graph.nodes:
+                bqm.linear[(v, c)] += p
+
+    return bqm
+
+
+def is_cycle(G: 'nx.Graph') -> bool:
```
thisac: Is there value in making this accept `GraphLike`?

```suggestion
@graph_argument('G', as_networkx=True)
def is_cycle(G: GraphLike) -> bool:
```

dimod/generators/markov.py
```diff
@@ -0,0 +1,107 @@
...
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+import dimod
+
+__all__ = ['markov_network',
+           ]
```
thisac, line 18:
```suggestion
__all__ = ['markov_network']
```

dimod/generators/partition.py
```diff
@@ -0,0 +1,94 @@
...
+
+import itertools
+import math
+
+import dimod
+from dimod.typing import GraphLike
+from dimod.decorators import graph_argument
+
+__all__ = ["graph_partition",
+           ]
```
thisac, line 23:
```suggestion
__all__ = ["graph_partition"]
```

thisac: Seems like this was deliberate. Looks a bit funky to me.

dimod/generators/social.py
```diff
@@ -0,0 +1,77 @@
...
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+import dimod
+
+__all__ = ["structural_imbalance",
+           ]
+
+
+def structural_imbalance(graph: 'nx.Graph') -> dimod.BinaryQuadraticModel:
```
thisac, line 21: Not sure why some of these allow `GraphLike` and some don't. Seems like this could also use the `@graph_argument('graph', as_networkx=True)` decorator, as could a few other ones missing it.

dimod/generators/social.py
```diff
@@ -0,0 +1,77 @@
...
+        ...
+        >>> bqm = dimod.generators.structural_imbalance(S)
+        >>> bqm.linear      # doctest: +SKIP
+        {'Alice': 0.0, 'Bob': 0.0, 'Eve': 0.0}
+        >>> bqm.quadratic   # doctest: +SKIP
+        {('Alice', 'Bob'): -1.0, ('Alice', 'Eve'): 1.0, ('Bob', 'Eve'): 1.0}
+
+    """
+    if not hasattr(graph, 'nodes') or not hasattr(graph, 'edges'):
+        raise ValueError("Signed social network graph in NetworkX format required")
```
thisac, line 63: Maybe this is the reason why not to accept `GraphLike`?

dimod/generators/tsp.py
```diff
@@ -0,0 +1,134 @@
...
+                value = graph[v][u][weight]
+            except KeyError:
+                value = missing_edge_weight
+
+            bqm.add_quadratic((v, pos), (u, nextpos), value)
+
+    return bqm
+
+
+traveling_salesman = traveling_salesperson
```
thisac, line 134: I know these are probably the two common ways to refer to this problem, but I think it'd be nice to also add in `travelling_saleswoman` as an alias. :smile:

dimod/generators/coloring.py
randomir: Yes, in case `as_networkx=True`, input nx graph is simply passed through.

dimod/generators/coloring.py
randomir: Fails gracefully. See https://github.com/dwavesystems/dimod/pull/1421.

dimod/generators/coloring.py
randomir: Sure, if for nothing else, then for nx check. We could in theory rewrite this function to build adjacency dict, like I did in some other functions in this PR, but that's likely not worth the effort.

dimod/generators/markov.py
randomir: I agree, it's not my favorite indentation style, but it's (local) dimod's style.

dimod/generators/social.py
randomir: I've used `GraphLike` on functions that operate on unlabelled graphs, and NetworkX on those that require edge labels, like here.

thisac: Makes sense then. :+1:

thisac (approved):
Thanks @randomir!

dimod/generators/tsp.py
randomir: "Traveling salesman problem" is the standard/legacy terminology, so including it as an alias makes sense. Renaming TSP to traveling salesperson is a bit of a stretch, but fair for gender neutrality. Adding "traveling saleswomen" to the mix is a bit too much, IMO. If you insist, I'd rather remove the salesman variant.

thisac: For clarity and posterity, I wasn't really suggesting adding it. We should keep it as is IMO. 😅

PR #1423, opened 2026-05-12
Drop support for Python 3.9

arcondello:
#### AI Generation Disclosure
No AI used.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,7 +9,6 @@ classifiers =
     Programming Language :: Cython
     Programming Language :: Python
     Programming Language :: Python :: 3
-    Programming Language :: Python :: 3.9
     Programming Language :: Python :: 3.10
     Programming Language :: Python :: 3.11
     Programming Language :: Python :: 3.12
@@ -48,7 +47,7 @@ packages =
     dimod.serialization
     dimod.testing
     dimod.views
-python_requires = >=3.9
+python_requires = >=3.10
 zip_safe = False
 
 [pycodestyle]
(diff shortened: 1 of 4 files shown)


arcondello:
One thing, the CI builds end up installing NumPy from source for the tests because they use a newer manylinux image that we do. IMO that's fine, if a bit slow. We can update our manylinux image in another PR if we like.

setup.cfg
```diff
@@ -9,7 +9,6 @@ classifiers =
     Programming Language :: Cython
     Programming Language :: Python
     Programming Language :: Python :: 3
-    Programming Language :: Python :: 3.9
```
randomir, line 12: Not strictly part of dropping 3.9, but would you add 3.14 to the list?

randomir (approved):
LGTM.

Related, you can also drop some deps from `tests/requirements.txt` that require py<3.9.

arcondello:
> Related, you can also drop some deps from tests/requirements.txt that require py<3.9.

Hmm, that leaves some stuff untested. Let me see if I can raise the version or remove that stuff

randomir:
> Hmm, that leaves some stuff untested. Let me see if I can raise the version or remove that stuff

What I meant was, drop the environment marker and update the pinned version. Not actually remove the whole dep. But it's not clear to me what 
```
pandas==1.1.4; python_version < '3.9'
```
actually gives us.. as we never test on a version older than 3.9 :thinking:

arcondello:
Yeah, this all needs a detailed maintenance pass. I pushed some new versions so we get testing. We really should deprecate those functions though, or sweep their range in CI.

randomir:
Was thinking the same while working on #1420. I can do a pass after that PR.

PR #1424, opened 2026-06-05
Fix `@graph_argument` networkx check

randomir:
#### What does this implement/fix?

Fix a bug introduced in #1421, where we were eagerly checking (during decorated function compile time) if NetworkX is installed when `as_networkx` flag is set.

Now we relax that check and verify that NetworkX is available only during function run time.

Additionally, in this PR we also re-enable running tests (partially) if NetworkX is not installed.

One more fix bundled here: we now run `asv` quick check on PR/branch latest, not `main`. This makes sure asv won't fail after the code is merged to `main`.

#### AI Generation Disclosure

No AI tools used.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -156,7 +156,7 @@ jobs:
           command: |
             . env/bin/activate
             asv machine --yes
-            asv run --quick
+            asv run --quick HEAD^!
 
   # we could do this as part of the various test jobs but getting the pathing
   # and configuration to work correctly is a pain. And since there is not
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -156,7 +156,7 @@ jobs:
           command: |
             . env/bin/activate
             asv machine --yes
-            asv run --quick
+            asv run --quick HEAD^!
```
randomir, line 159: This makes sure asv runs on the latest commit in a branch, not latest in main. IMO, it makes more sense to run asv on relevant changes in PR.

thisac (approved):
LGTM

PR #1426, opened 2026-06-08
Drop `manylinux2014` wheels support

randomir:
`manylinux2014`/`manylinux_2_17` wheels are built on CentOS 7. CentOS 7 reached EOL on June 30th, 2024.

Scientific python community has been gradually migrating to `manylinux_2_28` for years. Latest versions of numpy/pandas only release `maylinux_2_28`, thus slowing down our builds to 30min+.

With this change in place, the build time is better (15min), but still not great.

#### AI Generation Disclosure

No AI tools used

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,8 +30,8 @@ test-command = "python -m unittest discover {project}/tests/"
 
 [tool.cibuildwheel.linux]
 archs = "x86_64 aarch64"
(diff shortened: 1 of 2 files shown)


randomir:
Also, see https://github.com/dwavesystems/dimod/pull/1423#issuecomment-4433457072.

arcondello (approved):

